PR #25, opened 2017-03-20
Mutate elements of array in parallel example using Rayon library

alisha17:
The example in the commit demonstrates how difference on each element of array is calculated in parallel (instead of sequential computation) using Rayon library in Rust.

### DIFF
--- a/src/pages/rayon.md
+++ b/src/pages/rayon.md
@@ -1,0 +1,32 @@
+# Rayon
+[![rayon-badge]][rayon]
+
+## Mutate the elements of an array using parallel processing
+
+```rust
+
+/**
+ * An example of mutating array elements in parallel using rayon
+ *
+ */
+
+extern crate rayon;
+use rayon::prelude::*;
+
+fn main() {
+    let arr = &mut [0, 7, 9, 11];
+
+    arr.par_iter_mut().for_each(|p| *p -= 1);
+
+    println!("{:?}", arr);
+}
+
+```
+
+The example uses the Rayon crate, which is the data parallelism library for Rust. It defines the trait, ParallelIterator, which provides the par_iter_mut() method. This method is actually a call to get a parallel iterator. It lets us write iterator-like chains that execute in parallel. It constructs the values we want and for_each() method, i.e.,the 'operations' method actually carries out the execution.
+
+
+<!-- Links -->
+
+[rayon-badge]: https://img.shields.io/crates/v/rayon.svg?label=rayon
+[rayon]: https://docs.rs/rayon
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ toml = "0.3"
 petgraph = "0.4.3"
 crossbeam = "0.2"
 clap = "2.20.5"
+rayon = "0.6.0"
 
 
 [build-dependencies]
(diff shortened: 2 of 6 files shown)


badboy:
Thanks Alisha! I'll take a look at the code asap. The CI test failed due to some unrelated issue (which we should fix as well, but it's not necessary for you to do it here)

src/pages/rayon.md
```diff
@@ -0,0 +1,46 @@
+# Rayon
+[![rayon-badge]][rayon]
+
+## Calculate difference of value 1 from each element of an array
```
brson: Let's state this so that it is closer to a question a Rust user might ask.

What we ultimately want is for users to come to the cookbook with a question, scan the examples for a use-case statement that sounds like theirs, and see an example of basically what they need to do.

In this case the title of the example is missing some important information, like that it uses parallelism, and that's mutating. Can you come up with a more evocative title?

src/pages/rayon.md
```diff
@@ -0,0 +1,46 @@
...
+    for i in input.iter() {
+        println!("{}", *i)
+    };
+}
+
+fn main() {
+    let arr = &mut [0, 7, 9, 11];
+    difference(arr);
+}
+```
```
brson: Let's condense this a bunch! Since we're creating a book, general commentary about rayon can be done in the book text, and much of the remaining comments are redundant with the code, or the text, obscuring the mechanisms on display.

Really, if you think about what's 'real' code here it boils down to:

```
fn main() {
    let arr = &mut [0, 7, 9, 11];

    arr.par_iter_mut().for_each(|p| *p -= 1);

    for i in arr.iter() {
        println!("{}", *i)
    }
}
```

As an example, that looks more crisp to me. But I think we can make this even more succinct. Looping through that array to print stuff takes a lot of code, but it's kinda beside the point of the example. Instead of printing each element let's just dump the entire array at once:

```
fn main() {
    let arr = &mut [0, 7, 9, 11];

    arr.par_iter_mut().for_each(|p| *p -= 1);

   println!("{:?}", arr);
}
```

Real short and sweet. It's obvious that `par_iter_mut` is the thing we're showing off. This is a simple example users can copy into their own code.

src/pages/rayon.md
```diff
@@ -0,0 +1,46 @@
...
+
+The purpose of this is to display the raw power of the rayon library. Rayon is a data-parallelism library for Rust. Ifunction in the example abovet is extremely lightweight and makes it easy to convert a sequential computation into a parallel one. It also guarantees data-race freedom. It also supports an API called "parallel iterators"; we can see the
+par_iter_mut() in the example above . These let you write iterator-like chains that execute in parallel.
+
+The first step involves defining a function 'difference' which calculates a difference of 1 from each element of array.
+We have given a reference to a mutable array of type i32 as an argument to the function. Then, the parallel iterator,
+par_iter_mut(),computes the difference of 1 from each element in parallel rather than sequentially.
+
+Finally we print all elements of the array to check if our code is producing correct results.
```
brson: Let's consider who the audience is, and what they are here for. They know Rust at this point, and know they need a library for some purpose. They want to know how to iterate in parallel. Let's give them what they want as clearly as we can without a lot of ceremony.

So the last sentence about printing the array, our user doesn't need to be told that - they know how to print arrays already. Likewise, there's a lot of description here of what the function is doing. How much does the user care about that? They'll recognize that it's a function that subtracts 1.

Basically, the information to convey here is that this example uses the `rayon` crate, which defines a trait (what trait is it), that provides a `par_iter_mut` method. What does that method do? What about that `for_each` method?

Please consider these questions and recompose the description of this example.

brson:
Thanks a bunch @alisha17! I left some notes. Please see what you think and make revisions.

brson:
This is great @alisha17. Thanks! I rebased your branch and pushed it to master.

PR #27, opened 2017-03-22
Add serde crates for toml example

badboy:
These are used by the serde example, but where not included in the Cargo.toml


brson:
Thanks @badboy !

PR #45, opened 2017-05-05
Parse URL example #33

bostelk:
Testing the waters...


dtolnay:
Nice! Great choice of example URL.

Would it be possible to do this without an unwrap? Either using `?` somehow or maybe a `match` that prints a message in the error case as well.

I think `Url` comes with a `Display` impl so you can print them directly - it doesn't have to be with as_str().

```rust
println!("url: {}", url);
```

dtolnay:
I merged a different PR involving the url crate so GitHub is showing some merge conflicts. I can deal with those if you want, once your example code is ready.

bostelk:
Alright, I have addressed those issues.

Issue #35, opened 2017-05-04
Add "Create new URLs from a base URL" example for url crate


brson:

dtolnay:
Added in #46.

PR #46, opened 2017-05-06
Add url joining example

mikeastock:
This PR addresses #35

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -26,7 +26,26 @@
 <a name="ex-url-new-from-base"></a>
 ## Create new URLs from a base URL
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/35)
+[![url-badge]][url]
+```rust
+extern crate url;
+
+use url::{ParseError, Url};
+
+const BASE_GITHUB_URL: &'static str = "https://github.com";
+
+fn main() {
+    match build_github_url("rust-lang/cargo") {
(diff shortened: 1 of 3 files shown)


dtolnay:
Thanks! Any thoughts around how this could be implemented without unwrap? In a real codebase you probably would not want the program to crash while doing this manipulation.

mikeastock:
@dtolnay let me know what you think of this take

src/net.md
```diff
@@ -26,7 +26,26 @@
...
+[![url-badge]][url]
+```rust
+extern crate url;
+
+use url::{ParseError, Url};
+
+const BASE_GITHUB_URL: &'static str = "https://github.com";
+
+fn main() {
+    match build_github_url("rust-lang/cargo") {
```
dtolnay: Please use 4 spaces for indentation inside the `match` block since this is what the Rust compiler codebase uses.

src/net.md
```diff
@@ -26,7 +26,26 @@
...
+      Ok(url) => assert_eq!(url.as_str(), "https://github.com/rust-lang/cargo"),
+      Err(error) => panic!("Errored created url: {}", error),
```
dtolnay: This should say "building" or "creating" instead of "created".

mikeastock:
Should be fixed

dtolnay:
And it's live! https://brson.github.io/rust-cookbook/net.html#ex-url-new-from-base

PR #47, opened 2017-05-06
Add example for creating base url by clearing path

### LINKED ISSUE #34: Add "Create a base URL by removing path segments" example for url crate
I suspect this is a common operation.

mikeastock:
Resolves #34

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -39,18 +39,36 @@ fn main() {
 ```rust
 extern crate url;
 
-use url::{ParseError, Url};
+use url::{ParseError as UrlParseError, Url};
+
+#[derive(Debug)]
+enum Error {
+    Parse(UrlParseError),
+    PathSegment,
+}
+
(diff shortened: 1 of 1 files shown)


src/net.md
```diff
@@ -35,7 +35,25 @@ fn main() {
...
+fn main() {
+    match parse_base_url("https://github.com/rust-lang/cargo") {
+        Ok(base_url) => assert_eq!(base_url.as_str(), "https://github.com/"),
+        Err(error) => panic!("Errored parsing url: {}", error),
+    }
+}
+
+fn parse_base_url(url: &str) -> Result<Url, ParseError> {
+    let mut url = Url::parse(url)?;
+    url.path_segments_mut().expect("Url to have a base").clear();
```
dtolnay: Can this `expect` fail? What would the input URL look like? Is there a better way to handle that case than crashing the application?

Issue #36, opened 2017-05-04
Add "Extract the URL origin (scheme / host / port)" example for url crate


brson:

dtolnay:
Added in #48.

PR #48, opened 2017-05-06
Added parsing url, create url from base and retrieve origin examples

### LINKED ISSUE #33: Add "Parse a URL from a string to a `Url` type" example for url crate

Piripant:
I know the parsing url and create url from base example have already been implemented, but since I have taken a fairly different approach, I wanted to make this pull request anyway.

If you'd like to have only the example about retrieving the origin of the url let me know.

Resolves #33 #36
Thanks for you time

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -14,7 +14,20 @@
 <a name="ex-url-parse"/>
 ## Parse a URL from a string to a `Url` type
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/33)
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! Would it be possible to write these without `unwrap`? In real usage you would probably not want these operations to crash your application.

Piripant:
Of course! I'll do it just now

Piripant:
I have just updated the examples to not use unwrap. Let me know what you think.
Thanks.

PR #51, opened 2017-05-07
Updated basics.md with rand crate snippets

budziq:
Following examples were added:
- Generate random floating point numbers
- Generate random values of a custom type

resolves https://github.com/brson/rust-cookbook/issues/50

Unfortunately due to https://github.com/rust-lang-nursery/rand/issues/146 i see no way for the `Generate random values of a custom type` to use SampleRange (which would be really nice)


dtolnay:
Nicely done!

PR #52, opened 2017-05-07
Make the table width consistent

budziq:
Added theme/book.css. It is a direct copy of the standard mdbook
book.css file with one addition of `width: 100%;` to the `table` style.
Infortunately mdbook allows only to substitute whole book.css.

Moreover please note that due to https://github.com/azerupi/mdBook/issues/227
`theme` directory has to be placed at the project root instead of `src`

resolves: https://github.com/brson/rust-cookbook/issues/42

### DIFF
--- a/theme/book.css
+++ b/theme/book.css
@@ -1,0 +1,821 @@
+html,
+body {
+  font-family: "Open Sans", sans-serif;
+  color: #333;
+}
+code {
+  font-family: "Source Code Pro", "Menlo", "DejaVu Sans Mono", monospace;
+  font-size: 0.875em;
+}
+.left {
+  float: left;
+}
+.right {
+  float: right;
+}
+.hidden {
+  display: none;
+}
+h2,
+h3 {
+  margin-top: 2.5em;
+}
+h4,
+h5 {
+  margin-top: 2em;
+}
+.header + .header h3,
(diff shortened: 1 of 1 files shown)


dtolnay:
Thank you! https://github.com/azerupi/mdBook/issues/227 is what I was missing.

The tables are still misaligned because the three columns are different widths in each table. Is that similarly simple to fix?

budziq:
Yeah I have noticed the discrepancy between cell widths. I have no Idea if it will be easy to fix as my background is mostly c/c++ and 100% of my css knowledge was gained in previous 10minutes ;) I can look but no guarantees.

budziq:
Unfortunately another 10min of css googling is not upto the task :(. The best I could do with `table-layout: fixed;` is to make all of table cells of identical width which in my opinion is much worse...
scrot

I guess that that each table cell/header would have to be annotated with explicit width value or css class. And I am not sure if it is possible in Markdown at all. And I would prefer to spend this time learning rust (via implementing some of the snippets) :)

@dtolnay To summarize I guess that either this pull is "good enough" for now and someone would be able to provide a better solution based on this commit or we just close it right away.

dtolnay:
Thanks for looking into it. I will keep the original issue open and someone else can follow up.

PR #53, opened 2017-05-07
Implemented "Make a HTTP GET request" example

budziq:
I took the liberty of adding reqwest to cargo.toml and renaming
"Make a HTTP GET request after parsing a URL" to "Make a HTTP GET request"
as reqwest handles basic URL parsing internally and there seamed
no reason to make the example more complex than required.

Also `rust,no_run` skeptic annotation was used to allow for compilation without making actual http requests.

resolves: https://github.com/brson/rust-cookbook/issues/39
blocked by: https://github.com/seanmonstar/reqwest/pull/79

Unfortunately reqwest is still dependent on serde 0.9 which causes our "encoding.md" examples to fail. Until https://github.com/seanmonstar/reqwest/pull/79 is resolved I would be happy to resolve any problems with the PR.

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -8,7 +8,7 @@
 | [Extract the URL origin (scheme / host / port)][ex-url-origin] | [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Remove fragment identifiers and query pairs from a URL][ex-url-rm-frag] | [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Serialize a `Url`][ex-url-serialize] | [![url-badge]][url] [![serde-badge]][serde] | [![cat-net-badge]][cat-net] [![cat-encoding-badge]][cat-encoding]|
-| [Make a HTTP GET request after parsing a URL][ex-url-reqwest] | [![reqwest-badge]][reqwest] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
+| [Make a HTTP GET request][ex-url-reqwest] | [![reqwest-badge]][reqwest] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 
 [ex-url-parse]: #ex-url-parse
 <a name="ex-url-parse"/>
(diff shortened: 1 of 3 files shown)


brson:
This looks just about perfect @budziq. Thanks!

Since this doesn't use the url crate directly, can you make a few changes to the markup:

- Change the name 'ex-url-reqwest' to perhaps 'ex-reqwest-basic', from both intro.md and net.md
- Remove the url badge from both pages

The way the error handling is set up here is different than [outlined in previous examples](https://brson.github.io/rust-cookbook/about.html#a-note-about-error-handling), though it looks good.

I might write it like

```
fn run() -> Result<()> {
    let mut res = reqwest::get("http://httpbin.org/get")?;
    let mut body = String::new();
    res.read_to_string(&mut body)?;

    println!("Status: {}", res.status());
    println!("Headers:\n{}", res.headers());
    println!("Body: {}", body);

    Ok(())
}

fn main() {
    run().unwrap();
}
```

'main' is just used for dealing with the error, by panicking, and 'run' is the real main function. The main reason I like this pattern is because once the reader sees it once they know how error handling is structured throughout, they can ignore 'main' and read 'run' as 'main'.

I think 'error-chain' has a macro to accomplish this now too, which we might adopt.

I'll make a few small suggestions inline as well.

brson:
Needs rebase.

src/net.md
```diff
@@ -211,10 +211,39 @@ fn main() {
 
 [ex-url-reqwest]: #ex-url-reqwest
 <a name="ex-url-reqwest"></a>
-## Make a HTTP GET request after parsing a URL
+## Make a HTTP GET request
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/39)
```
brson: Please add the appropriate badges here per the examples in basic.md

src/net.md
```diff
@@ -211,10 +211,39 @@ fn main() {
...
+The [`reqwest::get`] function parses the supplied url and makes a
+synchronus HTTP GET request. Obtained [`reqwest::Response`]
+status and headders are printed and custom `Result` with allocated `String` containing the HTTP response body is returned.
 
+```rust,no_run
+extern crate reqwest;
+#[macro_use]
+extern crate error_chain;
+use std::io::Read;
```
brson: Please add a blank line between 'extern' and 'use'

budziq:
Rebased and updated the code. I find the `quick_main!` less noisy and would opt for using it where possible.

Thanks for your time!

src/net.md
```diff
@@ -232,12 +232,44 @@ fn main() {
...
+[ex-url-basic]: #ex-url-basic
+<a name="ex-url-basic"></a>
+## Make a HTTP GET request
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/39)
+[![reqwest-badge]][reqwest] [![cat-net-badge]][cat-net]
 
+The [`reqwest::get`] function parses the supplied url and makes a
+synchronus HTTP GET request. Obtained [`reqwest::Response`]
+status and headders are printed. HTTP response body is red into an allocated [`String`] via [`read_to_string`].
```
fungos: red -> read

budziq: Done, Thanks!

brson:
Thanks @budziq ! Looks great.

PR #55, opened 2017-05-08
implemented "Remove fragment identifiers and query pairs from a URL"

budziq:
resolves: https://github.com/brson/rust-cookbook/issues/37


dtolnay:
Thank you!

PR #57, opened 2017-05-09
Made TOC's and badges consistent between chapters

budziq:
resolves: https://github.com/brson/rust-cookbook/issues/56


dtolnay:
Thank you!

PR #60, opened 2017-05-09
Recommend the use of quick_main.

### LINKED ISSUE #59: Switch error handling setup to quick_main! macro from error-chain
Right now the error handling pattern suggested by the cookbook's ['about'](https://brson.github.io/rust-cookbook/about.html#a-note-about-error-handling) page, and used in at least some of the examples, looks similar to `quick_main!`. Seems like we should just use that.

Switch the existing examples to replace 'main' with 'quick_main!', and update the about page.

cc @dtolnay @Yamakaky

Yamakaky:
Fixes #59.

I didn't modify the others examples since they didn't already use `error-chain`. Should I?

### DIFF
--- a/src/about.md
+++ b/src/about.md
@@ -74,15 +74,14 @@ fn run() -> Result<()> {
     Ok(())
 }
 
-fn main() {
-    run().unwrap();
-}
+quick_main!(run);
 ```
 
 This is using the `error_chain!` macro to define a custom `Error` and
 `Result` type, along with an automatic conversion from two standard
 library error types. The automatic conversion makes the `?` operator
-work.
+work. The `quick_main!` macro is also used to generate the boilerplate `main`
+from above.
 
 For more background on error handling in Rust, read [this page of the
 Rust book][error-docs] and [this blog post][error-blog].


dtolnay:
I would strongly prefer to use the other form of quick_main.

```rust
fn run() -> Result<()> {
    /* ... */
}

quick_main!(run);
```

This makes it look a lot more similar to the user's functions that they are pasting these code samples into.

brson:
Thanks @Yamakaky 🥇 

I think I prefer the version without the closure to, because it looks more 'typical' of Rust code.

brson:
I was thinking, by using `quick_main!` for the error chain case, that means that examples that use error-chain will be written `quick_main!(main)` but those that don't will be written `fn main() { run().unwrap(); }`. I wonder if that's inconsistent and confusing.

We could demand every example use error-chain, even when there's only one error type.

Opinions @dtolnay @Yamakaky ?

dtolnay:
I agree. Let's always use `quick_main`.

Yamakaky:
We seem to want to recommend the use of `error-chain`, so let's do that!

dtolnay:
Great! @Yamakaky could you update the PR? The example above this one should also use `quick_main`.

Issue #61, opened 2017-05-10
Cookbook ideas for log crate


brson:
Write examples for these use cases following the existing precedent and guidelines in CONTRIBUTING.md

- [x] [Log a debug message to the console](https://github.com/brson/rust-cookbook/issues/95)
- [x] [Log an error message to the console](https://github.com/brson/rust-cookbook/issues/96)
- [x] Log messages with a custom logger (added in #74)
- [x] Enable log levels per module (added in #78)
- [x] Log to the Unix syslog (added in #70)
- [x] [Log messages to a custom location](https://github.com/brson/rust-cookbook/issues/97)
- [x] [Include a timestamp on log messages](https://github.com/brson/rust-cookbook/issues/98)

Crate evaluation for `log`: See https://internals.rust-lang.org/t/crate-evaluation-for-2017-05-16-log/5185

Suggest ideas for other examples here if you have them!

budziq:
I recall that sameone has posted interest in these examples on the libz-blitz thread but I would like to create PR later today for "Log to the Linux syslog" since I have a working example already

https://play.integer32.com/?gist=a21a9b6735f80d072ebea471840d2534&version=undefined

rap2hpoutre:
Here is an example I just wrote for a simple use case (console log):

```rust
#[macro_use]
extern crate log;

use log::{LogRecord, LogLevel, LogMetadata, LogLevelFilter};

struct ConsoleLogger;

impl log::Log for ConsoleLogger {
    fn enabled(&self, metadata: &LogMetadata) -> bool {
        metadata.level() <= LogLevel::Info
    }

    fn log(&self, record: &LogRecord) {
        if self.enabled(record.metadata()) {
            println!("{} - {}", record.level(), record.args());
        }
    }
}

fn main() {
    log::set_logger(|max_log_level| {
                        max_log_level.set(LogLevelFilter::Info);
                        Box::new(ConsoleLogger)
                    })
            .unwrap();
    info!("hello log");
    warn!("warning");
    error!("oops");
}
```

Link to playground: http://play.integer32.com/?gist=80a6e5c21d710723a13a3817a6bb195c&version=undefined

Is this kind of example usefull? Did I missed something? Should I submit a PR?

budziq:
[...]

PR #69, opened 2017-05-11
Add "Log messages to the console" recipe

rap2hpoutre:
See https://github.com/brson/rust-cookbook/issues/61

I choose to replace "Log a debug message to the console" and "Log an error message to the console" by just "Log messages to the console" (which covers info, warn and error), because recipe is (imo) almost the same for the various message levels. Maybe I'm wrong.

I'm still learning and need mentoring, maybe the quality of my PR is a bit poor. Anyway, thanks to @budziq for his precious help.

I could add a text description around this example later, but today I'm not really inspired (and to be fair I'm not fluent at all in english, so I'm a bit scared).

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -3,8 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Parse command line arguments][ex-clap-basic] | [![clap-badge]][clap] | [![cat-command-line-badge]][cat-command-line] |
-| [Log a debug message to the console][ex-log-debug] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
-| [Log an error message to the console][ex-log-error] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
+| [Log messages to the console][ex-log-messages] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
 | [Enable log levels per module][ex-log-mod] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
 | [Log to the Unix syslog][ex-log-syslog] | [![log-badge]][log] [![syslog-badge]][syslog] | [![cat-debugging-badge]][cat-debugging] |
 | [Log messages to a custom location][ex-log-custom] | [![log-badge]][log] | [![cat-command-line-badge]][cat-command-line] |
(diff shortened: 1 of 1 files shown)


src/app.md
```diff
@@ -3,8 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Parse command line arguments][ex-clap-basic] | [![clap-badge]][clap] | [![cat-command-line-badge]][cat-command-line] |
-| [Log a debug message to the console][ex-log-debug] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
-| [Log an error message to the console][ex-log-error] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
+| [Log messages to the console][ex-log-messages] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
```
budziq: badges do not reflect the used crates please see the env_logger-badge

also [![cat-command-line-badge]][cat-command-line] should be changed to [![cat-debugging-badge]][cat-debugging]

src/app.md
```diff
@@ -104,21 +103,48 @@ The file passed is: myfile.txt
...
 ```
 
-[ex-log-debug]: #ex-log-debug
-<a name="ex-log-debug"></a>
-## Log a debug message to the console
+[ex-log-messages]: #ex-log-messages
+<a name="ex-log-messages"></a>
+## Log messages to the console
 
 [![log-badge]][log] [![env_logger-badge]][env_logger] [![cat-command-line-badge]][cat-command-line]
```
budziq: spurious env_logger-badge

src/app.md
rap2hpoutre: Thanks again for your help. So you mean replacing:

```
| [Log messages to the console][ex-log-messages] | [![log-badge]][log] [![env_logger-badge]][env_logger] | [![cat-command-line-badge]][cat-command-line] |
```

By 
```
| [Log messages to the console][ex-log-messages] | [![log-badge]][log] | [![cat-debugging-badge]][cat-debugging] |
```

Ok I could do that!

rap2hpoutre:
@budziq 

> as mentioned in #61 (comment) i would suggest making it a more advanced separate example. Something along the lines of "Creating a custom logger". As the trivial consol logging would be implemented in 5 line env_logger snippet

Then in #61:

> I would argue that this example is more involved (and rich with important usecase) than minimal (~5 lines) one that would use env_logger. But one of the maintainers would have to decide ;)

I think I totally misunderstood what you wrote in https://github.com/brson/rust-cookbook/issues/61 AND I think I still don't understand! I thought you said "Your example is OK because it has some details". But now, I'm lost.

Thinking about it, I'm not sure anymore I have a sufficient level to do such a PR (english level+rust level+brain level). Honestly, what do you think? Can I give up? (cause maybe it's a waste of time, somebody else would have done better work) If not, will you have the patience to continue to help me

budziq:
@rap2hpoutre You should definitely not give up :1st_place_medal: as you PR is almost perfect, and please note that this repository is an outreach program for people to learn and be a first step in contribution.

Honestly my english not top notch, so we might have misunderstood, so let me rephrase:
- Basic logging can be implemented in a more standard way by utilizing [`env_logger`](https://crates.io/crates/env_logger) with the smallest number of lines.
- Your example a separate usecase that shows users how to create a basic custom logger that might write to file/ring bells or etc. And IMHO your example (after some polishing) could be included in the cookbook.

rap2hpoutre:
@budziq OK so I could:

1. Re-add the links to (void) "Log a debug message to the console" and "Log an error message to the console" and re-add their "placeholders"
2. Add one _new_ link to "Create a basic custom logger"
3. Rename my example to "Create a basic custom logger"
4. Add a small description like: "Set a basic custom logger `ConsoleLogger` and log various messages to it" (not sure about the word "to")
5. Update intro.md after everything is OK

Right?

dtolnay:
I agree with @budziq, this example is really valuable but not as the very first `log` example that we show. I think `env_logger` would be a better fit for "I just want the log messages to show up." Then we can have this example to show how to define a custom logger.

@rap2hpoutre please rename this example to something like "Log messages in a custom format" and tweak something trivial about it to make it custom, like:

```rust
println!("Rust says: {} - {}", record.level(), record.args());
```

Then please add an example above this one showing a minimal very basic use of `env_logger`.

rap2hpoutre:
@dtolnay Ok thanks. I updated my todo-list above your comment. Ok with this?

dtolnay:
Yes that covers everything. Thank you!

rap2hpoutre:
Ok I will try to do it tomorrow!

rap2hpoutre:
I close this PR because: 
 - It's in conflict, and it has no value to resolve it cause I made mistakes in my commits
 - My commits were dirty ("oops")
 - I deleted lines that I did not have to delete
 - The title, description, etc. are not anymore what it's about. 

New clean (I hope) PR here: #74 (with the right title, description, no conflict, no random commit messages, etc.) Thanks to @budziq & @dtolnay for their help!!

PR #70, opened 2017-05-11
Implements "Log to the Unix syslog" example

budziq:
Please see https://github.com/brson/rust-cookbook/issues/61

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -134,7 +134,31 @@ Your favorite number must be 256.
 
 [![log-badge]][log] [![syslog-badge]][syslog] [![cat-debugging-badge]][cat-debugging]
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/61)
+Messages are logged to [UNIX syslog]. Logger backend is initialized
+with [`syslog::init`].
+[`syslog::Facility`] indicates type of program submitting log.
+[`log::LogLevelFilter`] denotes allowed log verbosity
+and `Option<&str>` holds optional application name.
+
+```rust,no_run
+#[macro_use] extern crate log;
+extern crate syslog;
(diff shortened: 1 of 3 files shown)


src/app.md
```diff
@@ -134,7 +134,31 @@ Your favorite number must be 256.
 
 [![log-badge]][log] [![syslog-badge]][syslog] [![cat-debugging-badge]][cat-debugging]
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/61)
+Messages are logged to [UNIX syslog]. Logger backend is initialized
+with [`syslog::init`].
+[`syslog::Facility`] indicates type of program submitting log.
+[`log::LogLevelFilter`] denotes alowed log verbosity
```
dtolnay: "alowed" should be "allowed"

src/app.md
```diff
@@ -134,7 +134,31 @@ Your favorite number must be 256.
...
+and `Option<&str>` holds optional aplication name.
```
dtolnay: "aplication" should be "application"

src/app.md
```diff
@@ -134,7 +134,31 @@ Your favorite number must be 256.
...
+
+```rust,no_run
```
dtolnay: Why is this `no_run`? Is there some reason it would fail if run in CI?

budziq: Thanks for your time :) 
it is marked as `no_run` in order not to clobber the syslog. I guess that most people would prefer not to have noise in their syslog at every testrun.

budziq:
@dtolnay typos fixed (sorry for that :disappointed:). The example itself is still `no_run` unless it is necessary ;)

PR #77, opened 2017-05-14
Added two more rand examples

budziq:
I do not know if rand examples are still open for contribution.
But I have added two more basic examples:
- "Generate random numbers within a range"
- "Generate random numbers with normal distribution"

These seam to be relatively frequently asked on SO.


dtolnay:
Everything is open for contribution! (Well, governed by what Brian decides in https://github.com/brson/rust-cookbook/issues/64#issuecomment-300650595.)

Thanks, these look great.

PR #78, opened 2017-05-14
Implemented "Enable log levels per module" example

budziq:
I'm not sure if the explanation is clear enough.

Also grouped the env_logger based examples together

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ serde_json = "1.0"
 toml = "0.4"
 url = "1.4"
 reqwest = "0.6"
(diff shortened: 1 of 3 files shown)


dtolnay:
Are these "\t" characters the recommended way to use env_logger? I don't see it in their documentation.

dtolnay:
Oh it's for alignment in the output at the bottom, I see. I would prefer to log just plain messages the way the user probably would want to.

budziq:
Fixed. Sorry for wasting your time. I was afraid the description was unintelligible.

PR #80, opened 2017-05-16
Implemented two examples for "Query the GitHub API"

budziq:
As mentioned in tracking issue https://github.com/brson/rust-cookbook/issues/24
"Query the GitHub API"
"Create and delete Gist with GitHub API"


dtolnay:
Great work!

PR #88, opened 2017-05-17
Added Edit button

budziq:
I have added an edit button to the page template allowing for direct edit of given page in GitHub

Unfortunately I see no other way than pulling whole default mdbook index.hbs and editing it inline.
At least it was possible to avoid pulling the book.js also...

This should resolve: https://github.com/brson/rust-cookbook/issues/82 but please note that it is my first time editing javascript so the solution might not be ideal.

### DIFF
--- a/theme/index.hbs
+++ b/theme/index.hbs
@@ -1,0 +1,117 @@
+<!DOCTYPE HTML>
+<html lang="{{ language }}">
+    <head>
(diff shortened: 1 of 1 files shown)


dtolnay:
Nice! I don't know javascript either so we'll just see whether anyone complains.

PR #107, opened 2017-05-17
Hide non essential parts of "Network" snippets by default

budziq:
Updated net.md to hide error_chain boilerplate by default
it can be exposed by selecting the "expand button"
The expand button is no longer hidden

This is connected to https://github.com/brson/rust-cookbook/issues/106
@dtolnay please check if net.md and decide if such approach is acceptable

- Please note no other section was updated as this is only a demo for the moment ;)

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -31,17 +31,17 @@ URL will print a message containing an explanation of what went wrong.
 
 ```rust
 extern crate url;
-
-#[macro_use]
-extern crate error_chain;
+#
+# #[macro_use]
+# extern crate error_chain;
 
 use url::Url;
-
-error_chain! {
-    foreign_links {
-        UrlParse(url::ParseError);
-    }
-}
+#
+# error_chain! {
+#    foreign_links {
+#        UrlParse(url::ParseError);
+#    }
+# }
 
 fn run() -> Result<()> {
     let s = "https://github.com/rust-lang/rust/issues?labels=E-easy&state=open";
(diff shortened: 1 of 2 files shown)


dtolnay:
Thanks! This is great.

Looks like the hidden parts are awkwardly indented by one space when visible. Is that something we can fix or a mdBook bug?

Is `// fn main()` required?

budziq:
both indentation and `// fn main()` are problems with mdbook:

- the hidden code has to have at least one space from `#` char or the doctests will not compile. Possibly this could be solved in mdbook
- `// fn main()` - is a workaround for mdbook inserting its own boilerplate 
``` #![allow(unused_variables)]
fn main() {
...
```
around playpen if it does not find string `fn main()` in the example (due to quick_main!)

dtolnay:
Let's fix both of those in mdBook before merging this. I think the solutions are:

- It should treat `#` the same way rustdoc does. That means no awkward one space indent.
- Do not insert boilerplate if there is a `quick_main!`.

dtolnay:
I filed https://github.com/azerupi/mdBook/issues/281 and https://github.com/azerupi/mdBook/issues/282 to follow up on the mdBook fixes.

budziq:
@dtolnay I made (hopefully nonidiotic) PR's for both of these issues.
For what is worth the indentation problem can be solved by locally patching the book.js
but I would strongly prefer for this problem to be solved in mdbook.

budziq:
Rebased and cleaned up the commit. As both problems are now fixed in upstream mdbook I would wait for the next release (0.0.22)

budziq:
@dtolnay mdbook has released a version (0.0.22) with fixes for both blockers.
I have rebased the PR for your consideration. 

If the idea and this PR is ok then I'll gladly update rest of the examples.

dtolnay:
Thanks!

PR #118, opened 2017-05-17
Travis "cron" builds run tests with "*" dependencies

budziq:
"cron" builds will trigger cargo test with all dependency versions
replaced with "*" to check for incompatible dependencies.
Also cron builds will never publish gh-pages

This is somewhat hacky so I am open to suggestions. I guess that I would prefer to remove Cargo.lock altogether and replace dependency versions in Cargo.toml in the repo instead of modifying it on the fly but I see no other possibility to make dependency check only in Travis nightly.

reolves https://github.com/brson/rust-cookbook/issues/89

related to https://github.com/brson/rust-cookbook/issues/75

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -12,6 +12,19 @@ if [ "$TRAVIS_BRANCH" != "master" ]; then
     exit 0
 fi
 
+# check for outdated dependencies on nightly builds
+if [ "${TRAVIS_EVENT_TYPE:-}" == "cron" ]; then
+    echo "This is cron build. Checking for outdated dependencies!"
+    rm ./Cargo.lock
+    cargo clean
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! We need `rm Cargo.lock` as well right? Otherwise it will use the locked dependencies.

The reason I prefer this approach is that we need Travis to play two roles:

- Tell us whether PRs are okay to merge
- Tell us when code is out of date

If we remove Cargo.lock and put "*" in Cargo.toml, it can no longer fulfill the first one reliably. An outdated example would prevent unrelated PRs from building green.

budziq:
Done. Honestly I was sure `cargo clean` removes Cargo.lock but never bothered to check.

PR #119, opened 2017-05-18
File exists example

### LINKED ISSUE #117: Detect whether executable exists on system path
Basically like Python's [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which).

gsquire:
Closes #117

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -3,6 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Read lines of strings from a file][ex-std-read-lines] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
+| [Determine if a file exists][ex-std-file-exists] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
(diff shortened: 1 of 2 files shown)


dtolnay:
Thanks! `Path::exists` is already documented with an example [here](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists) so I would prefer not to reproduce that in the cookbook. Ideally the cookbook would show common use cases for which an example does not fit well in the standard library's documentation. Would you be able to update this to show how to find a particular executable command in the [`PATH`](https://en.wikipedia.org/wiki/PATH_(variable))? Like Python's [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which). That's what I meant in #117.

gsquire:
Ah so that's what you meant. Yes, I can do that. Sorry for the confusion.

dtolnay:
On second thought, this may not make a good cookbook example. [It is way more complicated than I thought.](https://github.com/fangyuanziti/which-rs/blob/master/src/lib.rs) Let's hold off until the libz blitz gets a chance to look at the `which` crate.

gsquire:
Should I close this PR then and write some notes on the issue?

dtolnay:
Sorry for the confusion too. How about one of the other examples?

PR #120, opened 2017-05-18
HEAD request

### LINKED ISSUE #109: Use HEAD request to test whether endpoint exists
For example, check whether a particular GitHub user exists: https://api.github.com/users/ferris-the-crab

gsquire:
Closes #109


dtolnay:
Nicely done! Only two change I would suggest:

- In real code I would expect `5, 0` to be accompanied by a quick note about what the numbers mean, so that the reader doesn't have to look up the Duration API to figure out what the timeout is set to. Maybe something like this:

    ```
    // Set a 5 second timeout for the request.
    ```

- The wording of the intro makes it sound like the timeout is a way to avoid needing to receive a body, but actually the HEAD request is the way to avoid needing to receive a body. Could you rearrange those or move the timeout sentence to its own paragraph to make this clearer?

gsquire:
Done. I rephrased it in a different way so it doesn't sound tied together.

dtolnay:
[Here it is.](https://brson.github.io/rust-cookbook/net.html#ex-rest-head) Thanks for the great example!

PR #127, opened 2017-05-18
Replace red on category badges

jeanm:
The crate version badges are colour-coded such that post-1.0 (which we can loosely interpret as “good”) is blue, and pre-1.0 (“bad”) is orange. Given that colour scheme, all those red category badges make me think something bad is going on. This PR replaces them with light grey ones.

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -348,13 +348,13 @@ quick_main!(run);
 
 <!-- Categories -->
 
-[cat-command-line-badge]: https://img.shields.io/badge/-command_line-red.svg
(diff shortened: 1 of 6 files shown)


dtolnay:
Thanks! I agree red is not the right color.

For me the white-on-lightgrey text is much harder to read though.

filesystem
encoding

What is your opinion of something like this?

filesystem
encoding

---

selection_047

jeanm:
Absolutely, or even just a simple link, which I imagine would improve accessibility.

dtolnay:
I merged this with the uncolored style in a1426c3bb79a6fdb68021d1ffdea053a6abec21b.

PR #128, opened 2017-05-18
Basic usage of `env_logger` + debug!

### LINKED ISSUE #95: Log a debug message to the console
The bare minimum env_logger example.

jmcomets:
Fixes #95.

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -111,7 +111,35 @@ Your favorite number must be 256.
 
 [![log-badge]][log] [![env_logger-badge]][env_logger] [![cat-debugging-badge]][cat-debugging]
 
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! Could you show a more realistic example of interpolating a variable in the debug message? I would not want to see this line in any Rust project:

```rust
debug!("this is a debug {}", "message");
```

Same thing for the error PR. In that case it would be nice to call some function from the standard library that returns Result and then log the error value if it returned an error.

dtolnay:
I merged this along with a more realistic message in 3a2a9972fff7128db5b084a96b596d572475b673.

```rust
debug!("Executing query: {}", query);
```

PR #129, opened 2017-05-18
Add basic usage of env_logger's error! macro

### LINKED ISSUE #96: Log an error message to the console
Using env_logger. This is different from #95 in that it should show making a function call that returns a Result, matching on it, and logging in the error case with the error value included in the message.

jmcomets:
Fixes #96.

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -119,7 +119,22 @@ Your favorite number must be 256.
 
 [![log-badge]][log] [![env_logger-badge]][env_logger] [![cat-debugging-badge]][cat-debugging]
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/61)
+```rust
+#[macro_use] extern crate log;
+extern crate env_logger;
+
+fn main() {
+    env_logger::init().unwrap();
+
+    error!("this is an error {}", "message");
+}
+```
+
+Run this code with `cargo run` and you should see the following line:
+
+```
+DEBUG:main: this is a debug message
+```
 
 [ex-log-mod]: #ex-log-mod
 <a name="ex-log-mod"></a>


src/app.md
```diff
@@ -119,7 +119,22 @@ Your favorite number must be 256.
 
 [![log-badge]][log] [![env_logger-badge]][env_logger] [![cat-debugging-badge]][cat-debugging]
 
-[Write me!](https://github.com/brson/rust-cookbook/issues/61)
+```rust
+#[macro_use] extern crate log;
+extern crate env_logger;
+
+fn main() {
+    env_logger::init().unwrap();
```
budziq: Please avoid using unwrap https://brson.github.io/rust-cookbook/about.html#a-note-about-error-handling

src/app.md
```diff
@@ -119,7 +119,22 @@ Your favorite number must be 256.
...
+
+    error!("this is an error {}", "message");
```
budziq: Would you be willing to update the example with an actual error message from `Result` obtained via std lib function call? This would make the example more realistic.

src/app.md
```diff
@@ -119,7 +119,22 @@ Your favorite number must be 256.
...
+}
+```
+
+Run this code with `cargo run` and you should see the following line:
+
+```
+DEBUG:main: this is a debug message
```
budziq: This is not exactly right ;)

src/app.md
```diff
@@ -119,7 +119,22 @@ Your favorite number must be 256.
...
+extern crate env_logger;
+
+fn main() {
+    env_logger::init().unwrap();
+
+    error!("this is an error {}", "message");
+}
+```
+
+Run this code with `cargo run` and you should see the following line:
```
budziq: I would suggest explaining in one sentence what is actually done instead of the relatively obvious result.

src/app.md
dtolnay: This is fine and the env_logger docs recommend unwrap. It fails if you register two loggers but that never happens in any reasonable program.

src/app.md
jmcomets: Absolutely, I'll get this done ASAP.

src/app.md
jmcomets: I was wondering if this example should be as detailed as the `debug!` recipe, as there's already a short explanation there. I'll include a short explanation here as well.

src/app.md
jmcomets: Woops, copy-pasta 😄

src/app.md
budziq: Ok then. I guess other logger examples should be updated for consistency sake as well as a note in https://brson.github.io/rust-cookbook/about.html#a-note-about-error-handling might be needed. Something along the lines: `unless explicitly suggested by the crate docs`;)

src/app.md
budziq: Imho the debug example could describe basics while this one would assume that reader is already acquainted with the previous one and would focus on the part that is different between them.

dtolnay:
I merged this along with a more realistic error in 94689a54d3578b5adee8aeb18bec60cea44303d6.

```rust
error!("Failed to execute query: {}", err);
```

PR #132, opened 2017-05-19
Added example showing how to convert data to/from a hex string using …

bjnyfv:
…the data-encoding crate.

I have added an example which shows how to convert a &[u8] to a hex String and then back again by utilizing the data-encoding crate. Please let me know if there are any improvements that can be made and I will be happy to address them.

Also, please note that the example uses version 2.0.0-rc.1 of data-encoding, which is technically a release candidate. I chose to use this version because the documentation link on crates.io for this crate takes you to the documentation for this version.


src/encoding.md
```diff
@@ -225,6 +226,59 @@ to be percent-encoded. The choice of this set depends on context. For example,
...
+extern crate data_encoding;
+
+#[macro_use]
+extern crate error_chain;
+
+use data_encoding::{HEXUPPER, DecodeError};
+
+error_chain! {
+    foreign_links {
+        Utf8(std::string::FromUtf8Error);
```
dtolnay: This error can't happen right? There is only one `?` in the run function.

dtolnay:
Nicely done! LGTM except for the unneeded error case.

src/encoding.md
bjnyfv: Oops, good catch! Will fix that right now.

bjnyfv:
@dtolnay Fixed. Thank you for the quick responses!

PR #135, opened 2017-05-19
encoding/decoding base64

rap2hpoutre:
This recipe shows how to encode and decode base64 (https://github.com/brson/rust-cookbook/issues/116). 

_A first (messy) PR was created [here](https://github.com/brson/rust-cookbook/pull/121). Thanks to @budziq I'm not using the github interface again._


src/encoding.md
```diff
@@ -278,6 +279,50 @@ fn run() -> Result<()> {
 quick_main!(run);
 ```
 
+[ex-base64]: #ex-base64
+<a name="ex-base64"></a>
+## Encode and decode base64
+
+[![base64-badge]][base64] [![cat-encoding-badge]][cat-encoding]
+
+The [`base64`] crate provides [`encode`] and [`decode`] functions.
```
budziq: I would suggest to to remove the first sentence and add links to [`encode`] and [`decode`] in the second sentence here.

rap2hpoutre:
@budziq Thanks!! I fixed it.

PR #136, opened 2017-05-19
Add 'Recursively find duplicate file names' example

### LINKED ISSUE #126: Recursively find duplicate file names
Walk a directory and print out all file names that appear in more than one subdirectory. For example `a/b/cookbook.md` and `tmp/cookbook.md` would be considered duplicates and print "cookbook.md".

The [`walkdir`](https://docs.rs/walkdir) crate will be useful.

xpayn:
fix #126

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ syslog = "3.2.0"
 tar = "0.4.12"
 tempdir = "0.3.5"
 lazy_static = "0.2"
+walkdir = "1.0.7"
 
 [build-dependencies]
 skeptic = "0.9"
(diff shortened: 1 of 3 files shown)


dtolnay:
Could you implement this without using `unwrap`? It looks like those two unwraps can never fail but I think `filter_map` would let you write this in a way that doesn't need them.

Also let's not use error_chain if nothing can fail. You can write a `fn main() { ... }` directly.

xpayn:
`filter_map` needs a closure that returns an Option<T> and i didn't find a compact way to convert a bool (`is_dir`) into an Option
`filter` needs a closure that returns a bool, thus i can't use `?` inside it.
i'm not quite happy with how it looks right now but my rust knowledge is limited.

maybe there's a clever way to deal with the entries' filenames too (less allocation), but i fighted with it too...

dtolnay:
Maybe something like this. This keeps only the ok values and then filters out directories.

```rust
for entry in WalkDir::new(".").into_iter()
                              .filter_map(Result::ok)
                              .filter(|e| !e.file_type().is_dir()) {
```

xpayn:
it looks better indeed
i was trying to have only one filter but i prefer your solution anyway
thx

PR #138, opened 2017-05-19
implemented "Encode a string as application/x-www-form-urlencoded"

budziq:
The example might be a little contrived. I will gladly update it if necessary.
 
resolves: https://github.com/brson/rust-cookbook/issues/85


dtolnay:
I like it. Please add an assert for both the encoded and decoded value so the reader can see what it looks like without having to run the code.

budziq:
done

PR #140, opened 2017-05-19
Reqwest example posting file to Slack

### LINKED ISSUE #90: POST the contents of a file to a REST api
#80 added an example that shows posting a JSON body. Let's add a separate one to post the contents of a file.

AndyGauge:
Fixes #90


AndyGauge:
I think I'd rather post a file to a restful api that actually accepts octet-stream rather than just url encoding the content of a file.  I'll rewrite for a dropbox api.

dtolnay:
I agree, sending a file with octet-stream would be more orthogonal with the other examples that already exist. Thanks!

PR #141, opened 2017-05-20
Hide error_chain boiler plate by default in all examples

budziq:
Modified all examples to hide noisy error_chain related code

resolves: https://github.com/brson/rust-cookbook/issues/106


dtolnay:
Wow this is going to be way more readable. Thanks!

PR #144, opened 2017-05-20
Inline code with hyperlink is now highlighted

budziq:
Inline code with hyperlink has now the same color
and on hover underline as standard hyperlink.

resolves https://github.com/brson/rust-cookbook/issues/142


dtolnay:
Thanks. Let's remember to revert this when https://github.com/azerupi/mdBook/pull/289 is released.

PR #145, opened 2017-05-20
Fix typo in ex-dedup-filenames

luisbg:
s/diretories/directories
in Application Development example Recursively find duplicate file names.


dtolnay:
Great, thank you.

PR #147, opened 2017-05-20
Implemented the "Include timestamp in log messages" example

budziq:
I have given a stab in describing the example without passive voice. I'm afraid that it has become a little to wordy.

resolves https://github.com/brson/rust-cookbook/issues/98


dtolnay:
Thanks! It's much better without passive voice.

PR #148, opened 2017-05-21
Include Paginated API example

### LINKED ISSUE #92: Iterate through paginated requests
The crates.io API would be a good way to demonstrate this. A lot of their endpoints are limited to 100 elements per query, with a `page=2` parameter to see more elements.

```bash
curl -sS \
  -H 'Content-Type: application/json' \
  'https://crates.io/api/v1/crates/serde/reverse_dependencies?per_page=100&page=2'
```

cetra3:
fixes #92


dtolnay:
Thanks! Could you write this in a way that does not panic when an error happens? Please see [a note about error handling](https://brson.github.io/rust-cookbook/about.html#a-note-about-error-handling) for some ideas.

cetra3:
I guess I could return a `Result` in the Iterator, which would propagate the error upwards.  Would make it a tiny bit more inelegant.  Let me have a think.

dtolnay:
I merged this and added real error handling in de913078f3a40bde5a9d6c94e49211089048cd53.

cetra3:
@dtolnay Thanks! I had not had time to get around to it, but hopefully starting you off was good enough.  Your example is much cleaner than my initial one.

PR #150, opened 2017-05-23
Fix hex encoding / decoding example.

bjnyfv:
1. Updated intro.md to include link to the hex encoding / decoding example.
2. Updated text of the hex encoding / decoding badge to display "data-encoding" instead of "url".

Note: The "data-encoding" badge displayed on intro.md is shorter and wider than the other badges. As far as I can tell this is due to the relatively long name of the "data-encoding" crate. If there is some way for me to resolve this please let me know and I will do so.

### DIFF
--- a/src/encoding.md
+++ b/src/encoding.md
@@ -350,22 +350,24 @@ fn run() -> Result<()> {
 
 <!-- Crates -->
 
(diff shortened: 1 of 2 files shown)


dtolnay:
Thanks!

PR #155, opened 2017-05-24
Split logging examples into separate book chapter

budziq:
fixes https://github.com/brson/rust-cookbook/issues/152


dtolnay:
Thanks!

PR #162, opened 2017-05-25
BufReader is constructed from some type that implements Read, not Write

baishuai:
BufReader is constructed from some type that implements `Read`, not `Write`


dtolnay:
Thanks! Could you also fix the link at the bottom of this file on line 404?

PR #166, opened 2017-05-26
Removed passive voice from the last 4 examples

budziq:
- Run an External Command and Process Stdout
- Encode and decode base64
- Log messages with a custom logger
- Log messages to a custom location

fixes https://github.com/brson/rust-cookbook/issues/139


dtolnay:
Nicely done.

PR #173, opened 2017-05-27
Fixed "Consume a paginated RESTful API" example to reuse reqwest::Client

budziq:
fixes https://github.com/brson/rust-cookbook/issues/172


dtolnay:
Thanks for the quick fix.

PR #174, opened 2017-05-27
"Set custom headers and URL parameters for a REST request" example

budziq:
fixes https://github.com/brson/rust-cookbook/issues/91


dtolnay:
Thanks, I merged this with a minor fix in 440a39436ed9402996f12100c20b051bb69dde49.

Issue #154, opened 2017-05-23
Put a cache in front of the shields.io images


dtolnay:
Some shields.io requests are unreasonably slow. It should be okay to layer a 12-hour cache in front of them.

jaemk:
I started throwing [something](https://github.com/jaemk/badge-cache) basic together. I can plop it on my server once I get to cleaning it up.

dtolnay:
Wow! Thank you so much.

dtolnay:
Fixed in #175.

PR #175, opened 2017-05-28
use badge cache in front of shields.io

jaemk:
Issue #154

I've got something basic up and running. There's a noticable difference compared to shields.io, but I'm not sure how well it'll perform under load.

It's running on my personal server at the moment and I didn't feel like buying a pretty domain. If you think it's worth it for the rust community, I (or anyone else) can set up a dedicated server/domain for this. I tried to leave detailed setup directions and sample configs in the [repo](https://github.com/jaemk/badge-cache) for anyone to follow.


dtolnay:
Thanks! I think hosting it on your domain is fine.

I may have missed it in the code but how long do badges remain cached? Please make sure to include that with the [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) and [Expires](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires) headers of every response.

src/about.md
```diff
@@ -207,9 +207,9 @@ as are crates that are pending evaluation.
 [error-blog]: https://brson.github.io/2016/11/30/starting-with-error-chain
 [error-chain]: https://docs.rs/error-chain/
 [Rust Libz Blitz]: https://internals.rust-lang.org/t/rust-libz-blitz/5184
-[cat-science-badge]: https://img.shields.io/badge/science--x.svg?style=social
```
budziq: Wow excellent work :+1: 
Would it be possible to keep the ".svg" part of url?

In this way this would make a perfect drop in replacement that could still be reverted or enrolled with single sed. Currently rollback in case of problems would be quite painful.

jaemk: Yes, good call!

jaemk:
`Cache-Control` headers are now set for an hour. Once @budziq's changes are merged, I'll rebase the url changes here so they match the `shields.io` urls and can be easily changed/reverted!

jaemk:
Urls are now updated to be drop-in compatible with `img.shields.io` so they can be reverted with a search/replace `badge-cache.kominick.com/img.shields.io/g`

dtolnay:
How long are badges cached on your server and where in the code are they refreshed?

jaemk:
I'm caching them for 12 hours. They get cleared out at 11am and 11pm EST by a cron job running `badge-cache admin --clear-cached-badges`

dtolnay:
Could you roll that into the badge-cache code? Two advantages:

- I don't want to end up in a situation where we are serving very old badges and don't notice.
- A badge requested at 10:55 and again at 11:05 should be fine to serve from cache.

During a request, the server should be able to check the age of the cached file and refresh if it is old.

You may even be able to do that asynchronously so that any badge that has been requested in the past x days is automatically kept up to date by refreshing it every 12 hours.

jaemk:
Sure, I'll add it in later tonight/tomorrow.

jaemk:
@dtolnay New code is live and I've re-updated all current links.

PR #178, opened 2017-05-30
RFC - do not merge yet - Clipboard support

budziq:
Hi as discussed in https://github.com/brson/rust-cookbook/issues/171

I have a draft of how clipboard support for snippets might work.
A separate icon http://fontawesome.io/3.2.1/icon/copy/ in each snippet. After clicking contents of the snippet its contents are transferred to system clipboard thanks to clipboard.js.

As always, my skills in regard to css and js are below zero so most likely this is an awful hackjob (especially the handcrafted tooltip) but it gets the job done. 

Also it pulls whole book.js (and rewrites part of it) so most likely the changes would be best left in mdbook itself.

Would there be any interest in such feature?

### DIFF
--- a/theme/book.css
+++ b/theme/book.css
@@ -859,3 +859,21 @@ a:hover > .hljs {
 .light a > .hljs {
   color: #4183c4;
 }
+
+.tooltiptext {
+  visibility: hidden;
+  background-color: black;
+  color: #fff;
+  opacity: 0.8;
+  text-align: center;
+  border-radius: 6px;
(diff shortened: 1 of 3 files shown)


budziq:
Thanks. I've made a backport to mdbook. I'll see if they are interested and what changes need to be made.

cc @azerupi

budziq:
my backport to mdbook has been merged https://github.com/azerupi/mdBook/pull/303#issuecomment

PR #199, opened 2017-06-20
Adds example for "Decompress tarball while removing path prefix"

### LINKED ISSUE #187: "Decompress a tarball while removing a prefix from the paths" recipe
The purpose of this recipe is to show iteration over the entries in the tarball, not just use Archive::unpack.

j-haj:
fixes #187 

Adds example to decompress tarball by iterating over entries while removing a specified path prefix.

All tests passed locally.

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -5,6 +5,7 @@
 | [Parse command line arguments][ex-clap-basic] | [![clap-badge]][clap] | [![cat-command-line-badge]][cat-command-line] |
 | [Decompress a tarball][ex-tar-decompress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
+| [Decompress a tarball while removing a prefix from the paths][ex-tar-strip-prefix] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Traverse directories while skipping dotfiles][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
@@ -114,8 +115,8 @@ Your favorite number must be 256.
 
 [![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
 
-Decompress ([`flate2::read::GzDecoder::new`]) and
-extract ([`tar::Archive::unpack`]) all files from a compressed tarball
+Decompress ([`GzDecoder`]) and
+extract ([`Archive::unpack`]) all files from a compressed tarball
 named `archive.tar.gz` located in the current working directory.
 
 ```rust,no_run
@@ -160,10 +161,10 @@ fn run() -> Result<()> {
 
 Compresses `/var/log` directory into `archive.tar.gz`.
 
-Creates a [`File`] wrapped in [`flate2::write::GzEncoder`]
+Creates a [`File`] wrapped in [`GzEncoder`]
 and [`tar::Builder`]. </br>Adds contents of `/var/log` directory recursively into the archive
 under `backup/logs`path with [`Builder::append_dir_all`].
-[`flate2::write::GzEncoder`] is responsible for transparently compressing the
+[`GzEncoder`] is responsible for transparently compressing the
 data prior to writing it into `archive.tar.gz`.
 
 ```rust,no_run
@@ -187,6 +188,60 @@ fn run() -> Result<()> {
     let enc = GzEncoder::new(tar_gz, Compression::Default);
     let mut tar = tar::Builder::new(enc);
     tar.append_dir_all("backup/logs", "/var/log")?;
+    Ok(())
+}
+#
+# quick_main!(run);
+```
+
+[ex-tar-strip-prefix]: #ex-tar-strip-prefix
+<a name="ex-tar-strip-prefix"></a>
+## Decompress a tarball while removing a prefix from the paths
+
+[![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
+
+Strip a path prefix from the entries of a tarball before unpacking them.
+
+We iterate over the [`Archive::entries`], using [`Path::strip_prefix`] to remove
+the specified path prefix (`bundle/logs`) before extracting the [`tar::Entry`]
+via [`Entry::unpack`].
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+extern crate flate2;
+extern crate tar;
+#
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#         StripPrefixError(::std::path::StripPrefixError);
+#     }
+# }
+
+use std::fs::File;
+use std::path::PathBuf;
+use flate2::read::GzDecoder;
+use tar::Archive;
+
+fn run() -> Result<()> {
+    let file = File::open("archive.tar.gz")?;
+    let mut archive = Archive::new(GzDecoder::new(file)?);
+    let prefix = "bundle/logs";
+    let entries = archive
+        .entries()?
+        .filter_map(|e| e.ok())
+        .map(|mut entry| -> Result<PathBuf> {
+            // Need to get owned data to break the borrow loop
+            let path = entry.path()?.strip_prefix(prefix)?.to_owned();
+            entry.unpack(&path)?;
+            Ok(path)
+        });
+
+    println!("Extracted the following files:");
+    for path in entries.filter_map(|e| e.ok()) {
+        println!("> {}", path.display());
+    }
     Ok(())
 }
 #
@@ -445,19 +500,26 @@ fn run() -> Result<()> {
 
 <!-- Reference -->
 
-[`Default`]: https://doc.rust-lang.org/std/default/trait.Default.html
-[`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
-[`flate2::read::GzDecoder::new`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
-[`flate2::write::GzEncoder`]: https://docs.rs/flate2/*/flate2/write/struct.GzEncoder.html
-[`Iterator::filter`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter
-[`tar::Archive::unpack`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.unpack
-[`tar::Builder`]: https://docs.rs/tar/*/tar/struct.Builder.html
+[`Archive::entries`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.entries
+[`Archive::unpack`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.unpack
 [`Builder::append_dir_all`]: https://docs.rs/tar/*/tar/struct.Builder.html#method.append_dir_all
+[`Default`]: https://doc.rust-lang.org/std/default/trait.Default.html
+[`Entry::path`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.path
+[`Entry::unpack`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack
+[`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
+[`follow_links`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.follow_links
+[`GzDecoder`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
+[`glob_with`]: https://docs.rs/glob/*/glob/fn.glob_with.html
+[`GzEncoder`]: https://docs.rs/flate2/*/flate2/write/struct.GzEncoder.html
+[`Iterator::filter`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter
+[`MatchOptions`]: https://docs.rs/glob/*/glob/struct.MatchOptions.html
+[`Path::strip_prefix`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix
+[`tar::Archive`]: https://docs.rs/tar/*/tar/struct.Archive.html
+[`tar::Builder`]: https://docs.rs/tar/*/tar/struct.Builder.html
+[`tar::Entry`]: https://docs.rs/tar/*/tar/struct.Entry.html
+[`tar::Entries`]: https://docs.rs/tar/*/tar/struct.Entries.html
 [`WalkDir::depth`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html#method.depth
 [`WalkDir::DirEntry`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html
 [`WalkDir::min_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.min_depth
 [`WalkDir::max_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.max_depth
 [`WalkDirIterator::filter_entry`]: https://docs.rs/walkdir/*/walkdir/trait.WalkDirIterator.html#method.filter_entry
-[`follow_links`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.follow_links
-[`MatchOptions`]: https://docs.rs/glob/*/glob/struct.MatchOptions.html
-[`glob_with`]: https://docs.rs/glob/*/glob/fn.glob_with.html
--- a/src/intro.md
+++ b/src/intro.md
@@ -78,6 +78,7 @@ community. It needs and welcomes help. For details see
 | [Parse command line arguments][ex-clap-basic] | [![clap-badge]][clap] | [![cat-command-line-badge]][cat-command-line] |
 | [Decompress a tarball][ex-tar-decompress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
+| [Decompress a tarball while removing a prefix from the paths][ex-tar-strip-prefix] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Traverse directories while skipping dotfiles][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
@@ -253,6 +254,7 @@ Keep lines sorted.
 [ex-std-read-lines]: basics.html#ex-std-read-lines
 [ex-tar-compress]: app.html#ex-tar-compress
 [ex-tar-decompress]: app.html#ex-tar-decompress
+[ex-tar-strip-prefix]: app.html#ex-tar-strip-prefix
 [ex-toml-config]: encoding.html#ex-toml-config
 [ex-url-base]: net.html#ex-url-base
 [ex-url-basic]: net.html#ex-url-basic


src/app.md
```diff
@@ -5,6 +5,7 @@
 | [Parse command line arguments][ex-clap-basic] | [![clap-badge]][clap] | [![cat-command-line-badge]][cat-command-line] |
 | [Decompress a tarball][ex-tar-decompress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
+| [Iterate through entries of tarball, removing path prefix][ex-tar-iterate] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
```
budziq: Please note the original title
"Decompress a tarball while removing a prefix from the paths"

emphasis on decompress 😉 

Please update the title , tag and contents of example

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-tar-iterate]: #ex-tar-iterate
+<a name="ex-tar-iterate"></a>
+## Iterate through the entries of a tarball, removing a path prefix
+
+[![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
+
+Iterates through the elements of a tarball, removing their path prefix.
```
budziq: We aim here to actually decompress the entries

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
...
+
+Creates an archive, `archive.tar.gz`, using the approach shown in the
```
budziq: no need to create the archive here (we already have example for it) lets keep the snippets short and to the point.

Just assume existence of `archive.tar.gz`

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
...
+extern crate flate2;
+#
+# error_chain! {
+#     foreign_links {
+#         Io(::std::io::Error);
+#         StripPrefixError(::std::path::StripPrefixError);
+#     }
+# }
+
+use flate2::Compression;
```
budziq: please sort `use` entries with std first and crate second

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
...
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use std::fs::File;
+use std::path::Path;
+use tar::Archive;
+
+fn run() -> Result<()> {
+    // Create tarball archive
```
budziq: no need to create the tarball here lets assume it exists. the example is `no_run` anyway so our testrunner will only verify that it compiles.

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
...
+    let file = File::open("archive.tar.gz")?;
+
+    // First get a streaming decoder to handle the gzip decompression and pass
+    // that into a tar::Archive
+    let mut archive = Archive::new(GzDecoder::new(file)?);
+
+    // Print out the contents of the archive, removing the `bundle/logs` prefix
+    println!("archive.tar.gz contains the following files:");
+    for entry in archive.entries()? {
+        let entry = &entry?;
```
budziq: please use the functional instead of imperative style if possible

src/app.md
```diff
@@ -189,6 +190,69 @@ fn run() -> Result<()> {
...
+        let path = entry.path()?;
+        println!("\t{}", path.strip_prefix(Path::new("bundle/logs"))?.display());
```
budziq: no need to create new `Path` every iteration

src/intro.md
```diff
@@ -237,6 +238,7 @@ Keep lines sorted.
 [ex-std-read-lines]: basics.html#ex-std-read-lines
 [ex-tar-compress]: app.html#ex-tar-compress
 [ex-tar-decompress]: app.html#ex-tar-decompress
+[ex-tar-iterate]: app.html#ex-tar-iterate
```
budziq: please change the tag to reflect the example that we will have after its finished

j-haj:
@budziq I'm working on making the for loop more functional, but am running into some issues related to ownership. In terms of the general principle though, it seems the original is very clear, where is the best functional approach I have currently is:

```rust
for path in archive.entries()?.filter_map(|r| r.ok())
                                   .map(|e| e.path())
                                   .filter_map(|r| r.ok())
                                   .flat_map(|p| p.strip_prefix(prefix)) {
          println!("\t{}", path.display());
}
```
which seems less clear (and again, this is not currently compiling). On the other hand, maybe you just meant to rewrite as:

```rust
for entry in archive.entries()? {
        let path = (&entry?).path()?.strip_prefix(prefix)?;
        println!("\t{}", path.display());
}
```
A third idea would be to use `foreach` rather than a for-loop.

Can you point me in the right direction? I've been on IRC all morning and it's not clear what the best functional approach is here, and getting the current functional approach to work hasn't gone smoothly 😃

budziq:
I would suggest updating the PR so that the entries are actually decompressed first (with acceptable prose description) using any means neccessary and then we will workout the most idiomatic solution in next step.

budziq:
@j-haj so not to leave you hanging :smile_cat: 

>  it seems the original is very clear, where is the best functional approach I have currently is ... which seems less clear

Hmm, the final solution I have in mind is quite short and has the added benefit that it does not bail out from function on first error (the iterator goes through all the). But I might be a little skewed so please follow my suggestion and provide an iterative solution first and we will look how we can improve it. If the functional result will not be readable to a newcomer, then we might stick to iterative approach.

Here are some further suggestions:
> am running into some issues related to ownership

In the context of the `entries()` iterator each entry is a temporary seen by given function (map, filter_map etc). So doing something like `.map(|e| e.path())` will try to return a `path` borrowed from temporary `e` and `e` stops existing right after the function returns. In general case the trick to satisfy the borrowck is to either:
 - make the returned object owned instead of borrowed
 - do all of the actual work (+side effects) in a single closure (representing one loop iteration) while the temporary variables are still live.
 - pass the borrowed object with the object from which it is borrowed from. - but it does not make sense here

Your first approach for functional implementation is not far from ok. But first try to establish what needs to be done in an iterative solution :+1:  

If you encounter any blockers, please do not hesitate to ask here :smiley:

j-haj:
@budziq Thanks for all the help and suggestions! I'll keep plugging away on the iterative approach first and see where I get this morning!

j-haj:
Just wanted to check in on this -- got a little sidetracked this weekend but this has not fallen off my radar.

budziq:
No rush mate ;)

j-haj:
I have been trying to get the functional approach to work, and got some help in the IRC channel, but still can't get this working... I may just be a little frustrated from how difficult it's been to get this work (with still no success), but if this solution is on the right track I don't think it is a great example for a beginner without copious amounts of explanations on what each part of the solution is doing.

```rust
archive.entries()?.map(|e| { 
        e?.path()?.strip_prefix(prefix)?.to_str().and_then(|p| {
            println!("\t{}", p);
            e?.unpack(p)
        })
    }).filter_map(|e| e.ok());
```

I'm not sure if this is the right track, but maybe you could clean this up by zipping the initial entry with the trimmed path, and then iterate/map over the zipped collection mapping `(entry, path) -> entry.unpack(path)`. But again, I would argue by making the example so compact and functional it becomes less understandable for the new user. On the other hand, maybe beginners aren't the target audience here.

budziq:
Hi @j-haj 
> I may just be a little frustrated from how difficult it's been to get this work

I'm sorry :disappointed:. I don't tend to hang on IRC so the only way to reach me would be via e-mail or github comments.

I see co-maintainership of this repo as an opportunity to sharpen my teaching skills :smile: do in case you have any problems in the future please do not hesitate to poke me here.

> maybe beginners aren't the target audience here.

They are very much so! This is why I've mentioned earlier that iterative approach might eventually be better if the functional one is not readable enough.

So maybe lets work through the iterative approach as I would see it and go from there to the functional that may or may not be better style.

budziq:
We start with a standard iterative approach here
```rust
    let entries = archive.entries()?.filter_map(|e| e.ok());
    for mut entry in entries {
        let pth = entry.path()?.strip_prefix(prefix)?.to_owned();
        entry.unpack(&pth)?;
        println!(">{:?}", pth);
    }
```
and see that most for loops are natural candidates for `Iterator::map`

budziq:
So we might end up with something like
```rust
    let entries = archive
        .entries()?
        .filter_map(|e| e.ok())
        .map(|mut entry| -> Result<PathBuf> {
            let pth = entry.path()?.strip_prefix(prefix)?.to_owned();
            entry.unpack(&pth)?;
            Ok(pth)
        });

    for pth in entries {
        println!(">{:?}", pth);
    }
```

Unfortunately `for_each` is not yet stable so we must either use for loop, `fold` or `collect` to actually consume our iterator and obtain the desired side effects. So I choose a minimal for loop.

j-haj:
@budziq my iterative approach was very close to what you posted, but I was missing the `to_owned` portion. I don't think I would have come up with the functional approach you posted, especially the `|mut entry| -> Result<PathBuf>` part. I guess I need to spend some more time reading up on `to_owned`

budziq:
> my iterative approach was very close to what you posted, but I was missing the `to_owned`

That is great :) Yeah not being able to solve this borrow loop without copying is not really nice.

> I don't think I would have come up with the functional approach you posted

Well I did not came up with it right of the bat. I've started with iterative solution (exactly as I've originally suggested) applied a known transforms (`for` -> `Iterator::map` these are equivalent if you squint a little except for `map` being lazy and `for` loop being eager) and fixed the compiler type inference error with type annotation `-> Result<PathBuf>` (I had an uglier solution relatively quick but this one took some head scratching as the compiler message was not ideal).

FYI (`.filter_map(|e| e.ok())` can be also removed )

> I guess I need to spend some more time reading up on `to_owned`

There is nothing special about the `to_owned` itself. It takes a borrowed value and returns an owned (heap allocated) variant of given type (&str -> String or in this case Path -> PathBuf). 

What is tricky here is that we 
```
let pth = entry.path()?.strip_prefix(prefix)?.to_owned();
           <-----1----> temporarily borrow here
                        <---------2----------> operate on borrowed Path borrowed (return slice reborrowing Path)
                                              <-----3-----> return the owned copy of slice (PathBuf)
<--4--> bind the PathBuf to pth thus releasing the previous borrows made in (1) and (2)
```

Are you still up to finishing this example @j-haj :)?

j-haj:
Definitely still up for finishing this. I'll incorporate changes from our discussion either later tonight or  first thing tomorrow morning. Need to take care of work work now :)

j-haj:
Did we decide on which approach is best to present? On the one hand, the iterative approach is certainly easier to understand (at least to me). On the other, the functional approach is more robust an a better solution (albeit, more difficult to understand). I suppose one could also make the argument that the functional approach requires the user to do more research, which is more educational. Another alternative would be presenting both solutions, but that doesn't necessarily fit with the idea of this being a "cookbook" I suppose. What are your thoughts @budziq ?

budziq:
> Did we decide on which approach is best to present?

nope :) and I hope that you'll help me with the decision. I am largely inclined towards functional one but if it's truly much harder than iterative one I can concede. 
Moreover, the example is primarily aimed to provide a snippet for given crate usage and only secondly a guidance on subjectively best programming style ;) 

@j-haj I would suggest following course of action. Start writing the prose for the example and once it's ready we can discuss the snippet itself and your reservations in regard to one or the other approach.

j-haj:
I have most of the discussion done and am currently debugging the example. For some reason no entries are being unpacked and it appears the program thinks there are zero entries in the archive. I'm sure it's a small bug but just wanted to check in so you know things are slowly moving forward 👍

j-haj:
Alright I think we are pretty close. You can check the updated PR. However, I've diverged from Master. How would you like me to get caught up? I've tried a number of different approaches here but ended up resetting each time.

src/app.md
```diff
@@ -325,8 +405,15 @@ fn main() {
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
 [`flate2::read::GzDecoder::new`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
 [`flate2::write::GzEncoder`]: https://docs.rs/flate2/*/flate2/write/struct.GzEncoder.html
+[`std::path::Path::strip_prefix`]: https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix
```
budziq: please use shorter link `Path::strip_prefix`

src/app.md
```diff
@@ -325,8 +405,15 @@ fn main() {
...
+[`tar::Archive`]: https://docs.rs/tar/*/tar/struct.Archive.html
+[`tar::Archive::entries`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.entries
```
budziq: please use shorter link `Archive::entries`

src/app.md
```diff
@@ -325,8 +405,15 @@ fn main() {
...
 [`tar::Archive::unpack`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.unpack
 [`tar::Builder`]: https://docs.rs/tar/*/tar/struct.Builder.html
+[`tar::Entry`]: https://docs.rs/tar/0.4.13/tar/struct.Entry.html
```
budziq: please use wildcard link to always point to the latest documentation. Please see links below

src/intro.md
```diff
@@ -237,6 +238,7 @@ Keep lines sorted.
 [ex-std-read-lines]: basics.html#ex-std-read-lines
 [ex-tar-compress]: app.html#ex-tar-compress
 [ex-tar-decompress]: app.html#ex-tar-decompress
+[ex-tar-decompress-remove-path-prefix]: app.html#ex-tar-decompress-remove-path-prefix
```
budziq: please use shorter example tag something along the lines of `ex-tar-strip-prefix`

src/app.md
```diff
@@ -325,8 +405,15 @@ fn main() {
...
+[`tar::Entry::path`]: https://docs.rs/tar/0.4.13/tar/struct.Entry.html#method.path
+[`tar::Entry::unpack`]: https://docs.rs/tar/0.4.13/tar/struct.Entry.html#method.unpack
```
budziq: please use shorter links `Entry::path` and `Entry::unpack`

src/app.md
```diff
@@ -325,8 +405,15 @@ fn main() {
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
 [`flate2::read::GzDecoder::new`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
```
budziq: I would suggest using just `GzDecoder` here

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+#     }
+# }
+
+use std::fs::File;
+use std::path::PathBuf;
+use flate2::read::GzDecoder;
+use tar::Archive;
+
+fn run() -> Result<()> {
+    // Get a reference to archive.tar.gz, which is a sample tarball created for
```
budziq: i would remove this comment and most of the other comments as well. 
The prose description and code itself should suffice for the reader to understand the example. Providing more text makes the example quite noisy.

Please note that there are already less advanced examples for these crates.

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+        // The `-> Result<PathBuf>` is here to aid the type inference
+        .map(|mut entry| -> Result<PathBuf> {
+            // Note: `to_owned()` is a more generic version of `clone()`
+            let path = entry.path()?.strip_prefix(prefix)?.to_owned();
+            entry.unpack(&path)?;
+            Ok(path)
+        });
+
+    // Since maps are evaluated lazily, we haven't actually extracted anything
+    // from the tarball yet. As we iterate through the map, we unpack the
```
budziq: `// Consume iterator actually extracting the entries`

would suffice but please note that we have less advanced examples showing this so even this might be superfluous.

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+    let file = File::open("archive.tar.gz")?;
+
+    // First get a streaming decoder to handle the gzip decompression and pass
+    // that into a tar::Archive
+    let mut archive = Archive::new(GzDecoder::new(file)?);
+
+    // Specify prefix we are removing
+    let prefix = "bundle/logs";
+
+    let extracted_entries = archive
```
budziq: `extracted_entries` might not bethe best name as these will not get extracted until you consume this iterator. 
`entries` will suffice

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+[ex-tar-decompress-remove-path-prefix]: #ex-tar-decompress-remove-path-prefix
+<a name="ex-tar-decompress-remove-path-prefix"></a>
+## Decompress a tarball while removing a prefix from the paths
+
+[![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
+
+Decompresses the tarball by iterating over the entries and unpacking them after
+removing a specified path prefix.
+
+Create a [`tar::Archive`] from a [`flate2::read::GzDecoder::new`], iterate over
```
budziq: Please note that there are already less advanced examples for these crates showing the decompression. So we might like to focus on the stripping prefix part.

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+fn run() -> Result<()> {
+    // Get a reference to archive.tar.gz, which is a sample tarball created for
+    // this example
+    let file = File::open("archive.tar.gz")?;
+
+    // First get a streaming decoder to handle the gzip decompression and pass
+    // that into a tar::Archive
+    let mut archive = Archive::new(GzDecoder::new(file)?);
+
+    // Specify prefix we are removing
```
budziq: the variable name is enough also please loose the newlines around it

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+use std::fs::File;
+use std::path::PathBuf;
+use flate2::read::GzDecoder;
+use tar::Archive;
+
+fn run() -> Result<()> {
+    // Get a reference to archive.tar.gz, which is a sample tarball created for
+    // this example
+    let file = File::open("archive.tar.gz")?;
+
```
budziq: please remove the newline

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+    let extracted_entries = archive
+        // Get the entries of the archive
+        .entries()?
+        // Entries is an iterator that yields `Result<Entry>`
+        // this `filter_map` allows us to get the `Entry` from the
+        // `Result<Entry>`
+        .filter_map(|e| e.ok())
+        // This is where we strip the path prefix and extract the file.
+        // The `-> Result<PathBuf>` is here to aid the type inference
```
budziq: return type annotations are exactly meant to aid type inference. so this might be superfluous.

budziq:
also to update please use 
```bash
git remote update
git rebase upstream/master # replace "upstream" with the name of git remote for canonical brson/rust-cookbook repo
git push origin contributing --force # assuming that "origin" is the remote name for your fork
```

src/app.md
j-haj: Whoops! Sorry about this one...

src/app.md
j-haj: In my updated discussion I don't reference `GzDecoder` (since it is mentioned in the `ex-tar-decompress` example). The only example that uses `GzDecoder` now is the `ex-tar-decompress` example, so I'd need to change their link as well (which I'm happy to do). Do you want me to make this change?

budziq: Sure make the examples uniform while you're at it :+1:

src/app.md
```diff
@@ -189,6 +190,85 @@ fn run() -> Result<()> {
...
+            entry.unpack(&path)?;
+            Ok(path)
+        });
+
+    // Since maps are evaluated lazily, we haven't actually extracted anything
+    // from the tarball yet. As we iterate through the map, we unpack the
+    // entries individually.
+    println!("Extracted the following files:");
+    for path in extracted_entries {
+        println!("> {:?}", path);
```
j-haj: When running this locally I had to use the following, since there was an `Err` in my entries

```rust
for path in entries.filter_map(|e| e.ok()) {
    println!("> {}", path.display());
}
```

otherwise it stops on the `Err`, which was my first element encountered from `entries` and nothing gets extracted. On the other hand, maybe it's better to add that at the end of the map... what are your thoughts? Is there a more idiomatic approach here?

budziq: As always its a matter of taste. The solution you are proposing is the one I would have used :+1: (unless I wanted to actually handle the errors which we don't in this case)

j-haj:
@budziq In trying to squash my commits into a single commit, I appear to have picked up the changes of `ex-file-skip-dot`. I've been going in circles with this and am not sure what the best approach is for fixing. Any suggestions? Two nights ago I had successfully rebased my commits into a single commit, but then it added my name to a bunch of other commits, which I didn't think was good, so I reset that. It's been stream of rebasing, resetting, etc. and now it seems it is in a very weird state.

budziq:
@j-haj 
I'm afraid that I cannot help you without actually seeing the state of your git tree. 
Good rule of thumb when working with git is to always save your work  to a second separate branch other the one you are working on so it never gets lost and you can always go back to when rebase or merge goes awry. 

I would not drop the changes yet but try to `git cherry-pick` them onto the current master branch.

j-haj:
I have not seen the `git cherry-pick` command before -- I'll read up on that and keep working on getting this resolved...

budziq:
@j-haj hi how are things? If there are any problems I'll gladly help

j-haj:
Working up the courage to sort out this git stuff. I was working on a Firefox bug this weekend so didn't make any progress from last week. This is on my list for today

budziq:
@j-haj no hurry :) I just don't want you to loose motivation when you're on a finish line

j-haj:
@budziq alright I *think* I got everything sorted out... 😰  -- I ran the tests and test case `basics_10` failed. I also noticed an error that cargo wasn't able to find the `bitflags` crate. I'll investigate

budziq:
@j-haj Then PR away :+1: 

> I ran the tests and test case basics_10 failed. I also noticed an error that cargo wasn't able to find the bitflags crate. 

This is yet another manifestation of our testrunner bug https://github.com/brson/rust-skeptic/issues/18. I hope to have a permanent fix prior to my absence during next two weeks.

running `cargo clean` and then `cargo test` should sort it out for you.

j-haj:
Should I close this PR and open a new one?

budziq:
> Should I close this PR and open a new one?

Nope. Just push to your original branch with `git push origin --force` and the PR will be automatically updated and overwritten.

j-haj:
Ok -- I think it should be all set then. Commit 4eef582 should have all discussed changes.

src/app.md
```diff
@@ -193,6 +194,63 @@ fn run() -> Result<()> {
...
+## Decompress a tarball while removing a prefix from the paths
+
+[![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
+
+Strip a path prefix from the entries of a tarball before unpacking them.
+
+In this case we have a tarball, `archive.tar.gz`, whose contents are inside the
+sub-directory `bundle/logs`. First we create a [`tar::Archive`], similar to the
+[decompress a tarball](#ex-tar-decompress) example, using [`GzDecoder`]. Then we
+iterate over the [`Archive::Entries`], using [`Path::strip_prefix`] to remove
```
budziq: `Archive::Entries` is a struct but points to method producing this struct. please either change the link text or url

src/app.md
```diff
@@ -193,6 +194,63 @@ fn run() -> Result<()> {
...
+[ex-tar-strip-prefix]: #ex-tar-strip-prefix
+<a name="ex-tar-strip-prefix"></a>
+## Decompress a tarball while removing a prefix from the paths
+
+[![flate2-badge]][flate2] [![tar-badge]][tar] [![cat-compression-badge]][cat-compression]
+
+Strip a path prefix from the entries of a tarball before unpacking them.
+
+In this case we have a tarball, `archive.tar.gz`, whose contents are inside the
+sub-directory `bundle/logs`. First we create a [`tar::Archive`], similar to the
```
budziq: No need to describe exact same steps which were covered by another less advanced example (nor refer to it).
I would remove the first two sentences of this paragraph and mention the `bundle/logs` prefix in later sentences

j-haj:
See commit 624e31c for edits.

src/app.md
```diff
@@ -445,19 +500,26 @@ fn run() -> Result<()> {
 
 <!-- Reference -->
 
+[`Archive::entries`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.entries
+[`Archive::unpack`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.unpack
+[`Builder::append_dir_all`]: https://docs.rs/tar/*/tar/struct.Builder.html#method.append_dir_all
 [`Default`]: https://doc.rust-lang.org/std/default/trait.Default.html
+[`Entry::path`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.path
```
budziq: this link is not used anywhere

src/app.md
```diff
@@ -445,19 +500,26 @@ fn run() -> Result<()> {
...
+[`Entry::unpack`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
-[`flate2::read::GzDecoder::new`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
-[`flate2::write::GzEncoder`]: https://docs.rs/flate2/*/flate2/write/struct.GzEncoder.html
+[`follow_links`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.follow_links
+[`GzDecoder`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
```
budziq: this link points to a method instead of struct

Issue #202, opened 2017-06-24
Traverse directories while skipping dotfiles


budziq:
Please use [`walkdir`](https://docs.rs/walkdir) crate to show how we can operate on a directory recursively while efficiently omitting hidden files and directories ("dotfiles").

The idea is to show off usage of [`filter_entry`](https://docs.rs/walkdir/*/walkdir/trait.WalkDirIterator.html#method.filter_entry) 

.

tyoc213:
HI there, I have basically this

```rust

fn should_skip(entry: &DirEntry) -> bool {
    !entry.file_name()
         .to_str()
         .map(|s| s != "." && s.starts_with("."))
         .unwrap()
}

fn run() -> Result<()> {
    for entry in WalkDir::new(".").into_iter().filter_entry(|e| should_skip(e)) {
        println!("{:?}", entry);
    }
    Ok(())
}
```

The only problem Im having is that if I mark the current directory as starting with "." or "hidden" it inmediately stops, because being a directory and returning that it should be filtered out, the whole listing of the directory and subdirs is gone :).

So it is OK like that? or should I skip always the first entry being the current directory ".".

budziq:
Hi, nice :+1:  you can always use [`DirEntry::depth`](https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html#method.depth) to never skip the starting dir.

Also would suggest to always start with a draft PR (even if its almost empty). It is easier to give guidance (and further action items). I have few minor suggestions so far:

- do not `unwrap` or `expect`. `should_skip` can use `unwrap_or` for a safe default on failure
- please print out the path of entry with normal display trait "{}"

budziq:
closed with #203

PR #203, opened 2017-06-25
#202 new example for show how to skip hidden files with walkdir

tyoc213:
This is it right now 👍 

fixes https://github.com/brson/rust-cookbook/issues/202

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -7,6 +7,7 @@
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
+| [Traverse directories while skipping dotfiles][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively calculate file sizes at given depth][ex-file-sizes] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 
 [ex-clap-basic]: #ex-clap-basic
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot]: #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Uses [`WalkDirIterator::filter_entry`] to descend recursively into entries passing the `is_not_hidden` predicate thus skipping hidden files and directories whereas [`Iterator::filter`] would be applied to each [`WalkDir::DirEntry`] even if the parent is a hidden directory. 
+
+Root dir `"."` is yielded due to [`WalkDir::depth`] usage in `is_not_hidden` predicate.
+
+[![walkdir-badge]][walkdir] [![cat-filesystem-badge]][cat-filesystem]
+
+```rust,no_run
+extern crate walkdir;
+
+use walkdir::{DirEntry, WalkDir, WalkDirIterator};
+
+fn is_not_hidden(entry: &DirEntry) -> bool {
+    entry
+         .file_name()
+         .to_str()
+         .map(|s| entry.depth() == 0 || !s.starts_with("."))
+         .unwrap_or(false)
+}
+
+fn main() {
+    for result in WalkDir::new(".")
+            .into_iter()
+            .filter_entry(|e| is_not_hidden(e))
+            .filter_map(|v| v.ok()) {
+        println!("{}", result.path().display());
+    }
+}
+```
+
+
 [ex-file-sizes]: #ex-file-sizes
 <a name="ex-file-sizes"></a>
 ## Recursively calculate file sizes at given depth
@@ -325,8 +360,12 @@ fn main() {
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
 [`flate2::read::GzDecoder::new`]: https://docs.rs/flate2/*/flate2/read/struct.GzDecoder.html#method.new
 [`flate2::write::GzEncoder`]: https://docs.rs/flate2/*/flate2/write/struct.GzEncoder.html
+[`Iterator::filter`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter
 [`tar::Archive::unpack`]: https://docs.rs/tar/*/tar/struct.Archive.html#method.unpack
 [`tar::Builder`]: https://docs.rs/tar/*/tar/struct.Builder.html
 [`Builder::append_dir_all`]: https://docs.rs/tar/*/tar/struct.Builder.html#method.append_dir_all
+[`WalkDir::depth`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html#method.depth
+[`WalkDir::DirEntry`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html
 [`WalkDir::min_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.min_depth
 [`WalkDir::max_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.max_depth
+[`WalkDirIterator::filter_entry`]: https://docs.rs/walkdir/*/walkdir/trait.WalkDirIterator.html#method.filter_entry
--- a/src/intro.md
+++ b/src/intro.md
@@ -79,6 +79,7 @@ community. It needs and welcomes help. For details see
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
+| [Traverse directories while skipping dotfiles][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively calculate file sizes at given depth][ex-file-sizes] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 
 ## [Logging](logging.html)
@@ -209,6 +210,7 @@ Keep lines sorted.
 [ex-extract-links-webpage]: net.html#ex-extract-links-webpage
 [ex-file-post]: net.html#ex-file-post
 [ex-file-predicate]: app.html#ex-file-predicate
+[ex-file-skip-dot]: app.html#ex-file-skip-dot
 [ex-file-sizes]: app.html#ex-file-sizes
 [ex-global-mut-state]: basics.html#ex-global-mut-state
 [ex-hex-encode-decode]: encoding.html#ex-hex-encode-decode


tyoc213:
Oh I see, lots of fixes, will take care of them, thx!

src/app.md
```diff
@@ -7,6 +7,7 @@
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
+| [Recursively find all files skipping the ones starting with dot][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
```
tyoc213: Done

src/app.md
```diff
@@ -271,6 +272,51 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot] #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Recursively find all files skipping the ones starting with dot
+
```
tyoc213: I read a little about the pasive voice, I tried to correct that

src/app.md
```diff
@@ -271,6 +272,51 @@ fn run() -> Result<()> {
...
+# extern crate error_chain;
+extern crate walkdir;
+
+use walkdir::{DirEntry, WalkDir, WalkDirIterator};
+#
+# error_chain! {
+#     foreign_links {
+#         WalkDir(walkdir::Error);
+#         Io(std::io::Error);
+#         SystemTime(std::time::SystemTimeError);
```
tyoc213: Well, I dont know if the compiler say to me that Im having more than the needed ones? anyway I jst deleted SystemTime

src/app.md
```diff
@@ -271,6 +272,51 @@ fn run() -> Result<()> {
...
+
+fn should_skip(entry: &DirEntry) -> bool {    
+    !entry.file_name()
+         .to_str()
+         .map(|s| entry.depth() > 0 && s.starts_with("."))
+         .unwrap_or(false)
+}
+
+fn run() -> Result<()> {
+    for result in WalkDir::new(".").into_iter().filter_entry(|e| should_skip(e)) {
```
tyoc213: done

src/app.md
```diff
@@ -271,6 +272,54 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot] #ex-file-skip-dot
```
budziq: Rendering and link gets broken due to missing colon
```diff
@@ -272,7 +272,7 @@ fn run() -> Result<()> {
 # quick_main!(run);

-[ex-file-skip-dot] #ex-file-skip-dot
+[ex-file-skip-dot]: #ex-file-skip-dot
 <a name="ex-file-skip-dot"></a>
 ##  Traverse directories while skipping dotfiles
```

you can check rendering with mdbook
```bash
cargo install mdbook
cd ./path/to/rust-cookbook
mdbook serve -o
```

src/app.md
```diff
@@ -271,6 +272,54 @@ fn run() -> Result<()> {
...
+         .unwrap_or(false)
+}
+
+fn run() -> Result<()> {
+    for result in WalkDir::new(".")
+            .into_iter()
+            .filter_entry(|e| is_not_hidded(e))
+            .filter_map(|v| v.ok()) {
+        println!("{}", result.path().display());
+        println!("{}", result.file_name().to_str().map(|s| s).unwrap_or(""));
```
budziq: Leave out the second println! the first one with `path()` will suffice

src/app.md
```diff
@@ -271,6 +272,54 @@ fn run() -> Result<()> {
...
+
+fn is_not_hidded(entry: &DirEntry) -> bool {
+    !entry
+         .file_name()
+         .to_str()
+         .map(|s| entry.depth() > 0 && s.starts_with("."))
+         .unwrap_or(false)
+}
+
+fn run() -> Result<()> {
```
budziq: there is no actual error condition returned in any code path. So we might just drop error-chain usage here and turn `fn run` to `fn main`

src/app.md
```diff
@@ -271,6 +272,54 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot] #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Iterate over the entries that are not hidden files or hidden directories.
```
budziq: How about:

Iterates over filesystem entries efficiently skipping hidden files and directories with help of [`WalkDir::filter_entry`].

src/app.md
```diff
@@ -271,6 +272,54 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot] #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Iterate over the entries that are not hidden files or hidden directories. 
+
+[`WalkDir::filter_entry`] descend recursively entries passing the predicate whereas [`std::iter::Iterator::filter`] would by applied to each `DirEntry` even if the parent is a hidden directory. [`WalkDir::depth`] is used to skip the level 0 which is the current directory `"."`.
```
budziq: how about:
 "descends recursively into entries passing the `is_not_hidded` predicate."

also:
> "would by applied" - > "would be applied"

and in case of:
> [`WalkDir::depth`] is used to skip the level 0 which is the current directory `"."`
i would suggest rephrasing (we are actually avoiding the "skip" here)

Uses [`WalkDir::depth`] to avoid skipping the root directory even if it starts with '.'.

src/app.md
budziq: Well at the moment error-chain cannot infer if you do not use one of the error variants and assumes that you are planning use new one in the future. So the easiest way to solve it is to remove/comment out all the error variants and check if compiler starts to complain :wink:.

In this case no error variant is needed and we might as well drop error_chain usage here all together and switch `fn run` to `fn main`

src/app.md
tyoc213: OK, I was wondering how I can get the "current" errors? I mean in java and things like that it warns you that you have missed a throw or a catch... but I cant quite get it here?

budziq: Well in rust it's a little more mundane and yet a little more magical at the same time. In fact you cannot return more than one error type instead your error type has to implement `from` for each error variant (actually there is a little more painful [boilerplate](https://rustbyexample.com/error/define_error_type.html) there that error-chain solves with macros). This is resolved at compiletime instead of runtime like in java. 

In case you are missing a variant you will get error like `from not implemented for <missing_error_type>`.
I hope this helps a little

src/app.md
tyoc213: done, I was wondering what will be the correct way to print only the name in this case?

budziq: there would be few ways depending what you want to achieve. If you would like to obtain the `file_name`s collected into a vec you would use further iterator transforms but if you are just for the sideffects (like printing) `match` or `if let` for the Ok value would suffice.

src/app.md
```diff
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot]: #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Iterates over filesystem entries efficiently skipping hidden files and directories with help of [WalkDir::filter_entry].
```
budziq: this link is broken [WalkDir::filter_entry]

You can check rendering with `mdbook serve`

src/app.md
```diff
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot]: #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Iterates over filesystem entries efficiently skipping hidden files and directories with help of [WalkDir::filter_entry]. 
+
+[`WalkDir::filter_entry`] descends recursively into entries passing the `is_not_hidded` predicate. [`std::iter::Iterator::filter`] would be applied to each `DirEntry` even if the parent is a hidden directory. Uses [`WalkDir::depth`] to avoid skipping root directory `"."` at level 0.
```
budziq: [`std::iter::Iterator::filter`] is a little to long I would souggest [`Iterator::filter`].

Also I would suggest to join the two sentences with 
"into entries passing the `is_not_hidded` predicate whereas [`Iterator::filter`] would be applied"

src/app.md
tyoc213: mm for me it is working because it links to https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.filter_entry with 

```markdown
[`WalkDir::filter_entry`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.filter_entry
```

budziq: > mm for me it works

It does not for github markdown renderer and mdbook's pulldown-cmark (which is used for final rendering on releases). How do you verify?

src/app.md
tyoc213: well, I decided to drop completely the firsrt paragraph because was another way to repeat the second one, hope the redaction is OK.

src/app.md
tyoc213: I see that the first paragraph doesnt have the string inside the brakets... now I fixed in another way (compeltely removing this extra paragraph). I just commit.

src/app.md
```diff
@@ -367,4 +365,4 @@ fn main() {
 [`WalkDir::max_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.max_depth
 [`WalkDir::filter_entry`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.filter_entry
```
budziq: should be https://docs.rs/walkdir/*/walkdir/trait.WalkDirIterator.html#method.filter_entry
instead of https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.filter_entry

sorry to not have caught it earlier

src/app.md
```diff
@@ -276,9 +276,7 @@ fn run() -> Result<()> {
 <a name="ex-file-skip-dot"></a>
 ##  Traverse directories while skipping dotfiles
 
-Iterates over filesystem entries efficiently skipping hidden files and directories with help of [WalkDir::filter_entry]. 
-
-[`WalkDir::filter_entry`] descends recursively into entries passing the `is_not_hidded` predicate. [`std::iter::Iterator::filter`] would be applied to each `DirEntry` even if the parent is a hidden directory. Uses [`WalkDir::depth`] to avoid skipping root directory `"."` at level 0.
+[`WalkDir::filter_entry`] descends recursively into entries passing the `is_not_hidded` predicate skipping hidden files and directories except the root directory `"."` at level 0 checked with [`WalkDir::depth`] whereas [`Iterator::filter`] would be applied to each `DirEntry` even if the parent is a hidden directory.
```
budziq: I would start the paragraph with "Uses [`WalkDir::filter_entry`] to descend" and move the "except the root directory `"."` at level 0 checked with [`WalkDir::depth`]" to its own sentence at the bottom.

src/app.md
tyoc213: :O I see, me too!

tyoc213:
Lets try once more!

src/app.md
```diff
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
...
+In the predicate [`WalkDir::depth`] skips level 0 which is the root directory `"."`.
+
+[![walkdir-badge]][walkdir] [![cat-filesystem-badge]][cat-filesystem]
+
+```rust,no_run
+extern crate walkdir;
+
+use walkdir::{DirEntry, WalkDir, WalkDirIterator};
+
+fn is_not_hidded(entry: &DirEntry) -> bool {
```
budziq: should be `is_not_hidden`

src/app.md
```diff
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot]: #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Uses [`WalkDirIterator::filter_entry`] to descend recursively into entries passing the `is_not_hidded` predicate thus skipping hidden files and directories whereas [`Iterator::filter`] would be applied to each `DirEntry` even if the parent is a hidden directory.
```
budziq: - `is_not_hidded` should be `is_not_hidden`
- `DirEntry` should be linkified

src/app.md
```diff
@@ -271,6 +272,40 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-file-skip-dot]: #ex-file-skip-dot
+<a name="ex-file-skip-dot"></a>
+##  Traverse directories while skipping dotfiles
+
+Uses [`WalkDirIterator::filter_entry`] to descend recursively into entries passing the `is_not_hidded` predicate thus skipping hidden files and directories whereas [`Iterator::filter`] would be applied to each `DirEntry` even if the parent is a hidden directory. 
+
+In the predicate [`WalkDir::depth`] skips level 0 which is the root directory `"."`.
```
budziq: While its true that the predicate itself "skips level 0". I feel that it breaks readers a mental flow (requiring them inspect the code in order to understand why we can actually avoid skipping the root which is the main idea).

Mostly we aim for the example pros to be self explanatory so I
I would prefer this to mention that "Root dir is never skipped due to [`WalkDir::depth`] usage in `is_not_hidded` predicate".

Then readers know what is going on without back and forth to the actual code.

tyoc213:
I changed a little the example to revert that `!` also I did do `git rebase -i HEAD~6` and pick the first and the other I put `s` and then `:wq` and then `git push  origin master --force`... so I dont know if it is right or not.

budziq:
> git push origin master --force... so I dont know if it is right or not.

The exact same commands worked for me
rebase

Most likely you have pushed the wrong branch. 

Otherwise this version is ok :+1: 

If you'll still have problems with rebase. I'll do it this once during the merge :smile:

tyoc213:
I think I have a extrange rebase (work in progress) or something like that, so I reset it with `git rebase --abort` and started again

budziq:
then you'll need `git rebase --continue`

tyoc213:
Oh yeah, the second time it worked without conflicts...

budziq:
Perfect:heart:. Thanks @tyoc213!

I've edited the commit message to remove the noise from the squashed commits.

Keep them coming! It gets much smoother :)

PR #204, opened 2017-06-26
MathJax CDN has closed down.

rozza:
I noticed an js console warning when navigating the cookbook:

```WARNING: cdn.mathjax.org has been retired. Check https://www.mathjax.org/cdn-shutting-down/ for migration tips.```

So heres a PR with the location updated as per their [guidance](https://www.mathjax.org/cdn-shutting-down/).

### DIFF
--- a/src/incomplete/rand.md
+++ b/src/incomplete/rand.md
@@ -15,7 +15,7 @@ MathJax.Hub.Queue(function() {
     }
 });
 </script>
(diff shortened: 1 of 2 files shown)


budziq:
Hi @rozza  thanks! But the issue was solved upstream in mdbook some time ago. I don't know when the next mdbook release will happen but I venture it's within next two months @azerupi. 

If you would be willing, you could with seting of local fallback for mathjax in mdbook

budziq:
@rozza I'm sorry you've lost time on this. Please give me heads up if you see any other problems!

rozza:
@budziq no problems, glad its getting sorted, I'll look for an easy ticket to work on :) 👍

PR #210, opened 2017-07-01
Changed license to CC0 v1.0

budziq:
fixes https://github.com/brson/rust-cookbook/issues/209


brson:
This is the right patch, but we're going to have to do some more legwork before we can merge it.

To relicense we're going to need the permission of all the authors, which is a painful process. I've done this a few times and slightly differently each time.

The way we can probably do this is to ping every single contributor and have them make a canned statement here about their intent to license their contribution under CC0, like "I agree to license my contributions to rust-cookbook under the Creative Commons CC0 license". We'll want to make it clear that CC0 is super-permissive - it waives all rights, and is effectively public domain.

There is a slight wrinkle here in that some examples are stolen from the crate's own documentation. In those cases we'll need to get permission from the crate author.

Contributions to this project are individually slight, and it seems unlikely any past contributors are going to be too concerned about the terms, so we don't need to sweat it too much.

I want to think about this a bit more and get second opinions before we pull the trigger though.

aturon:
@budziq I believe there are some scripts we can use to help automate this process. @cmr, is yours still operable?

Michael-F-Bryan:
Just out of curiosity, how many people need to consent before you can change the license? It feels like waiting for 100% would be impractical (people change careers, don't check github, etc), so what is the normal procedure for something like this?

budziq:
Nominally all of them (sans ones whose contributions can be judged insubstantial i.e. fixing typos). Picking a license is a big commitment and changing it late in the project is not (and not meant to be) a streamlined process.
As these things go, ours re-licensing poll is not that bad (86% consent in one week). I'll give it another  two weeks and start nagging via other channels :)

budziq:
I've tried to contact @Yamakaky several times with no effect. Lets check what rust-lang-nursery organization owners think about merging license change without 100% consent.

cc @aturon @steveklabnik

steveklabnik:
Given that single tiny commit, I think it's fine, but we should get @aturon's signoff too.

aturon:
👍 from me, this is too small to be any issue.

Issue #214, opened 2017-07-04
"Handle errors correctly in main" - using error-chain crate


budziq:
super simple example which defines 1 error variant instead of using Box

somebrel:
I'll be working on this issue

budziq:
For now we will go with "Basics" for now and if we get enough examples that fit similar category we will make a separate section.

somebrel:
#234

Issue #209, opened 2017-06-30
Cookbook should probably be CC0 licensed


brson:
Since the content of the cookbook is intended to be copied freely, it would be better licensed CC0, with basically no restrictions.

Edit:

------------
We will track here the consent of all contributors (the list will be periodically updated):
Please drop a comment **here** if you consent or disagree.  [Link to CC0 license](https://creativecommons.org/choose/zero/). Once we gather consent following will be merged #210

- [x] @AlexEne
- [x] @AndyGauge
- [x] @CJStadler
- [x] @CrockAgile
- [x] @FaultyRAM
- [x] @Fierthraix 
- [x] @Lapz 
- [x] @Michael-F-Bryan
- [x] @Ophirr33
- [x] @Phaiax
- [x] @Piripant
- [x] @V1shvesh
- [x] @Wilfred
- [x] @alexcrichton
- [x] @alisha17
- [x] @badboy
- [x] @bjnyfv
- [x] @bobbo
- [x] @bostelk
- [x] @brson
- [x] @budziq
- [x] @cetra3
- [x] @charlesetc
- [x] @crypto-universe
- [x] @cschneid 
- [x] @derekdreery
- [x] @dhardy
- [x] @dhharris
- [x] @glennsl
- [x] @gsquire
- [x] @hegza 
- [x] @j-haj
- [x] @jaemk
- [x] @jbolila
- [x] @jeanm
- [x] @jmcomets
- [x] @jonashagstedt
- [x] @jrakow 
- [x] @lilianmoraru
- [x] @lucasem
- [x] @ludwigpacifici
- [x] @luisbg
- [x] @luser
- [x] @matklad
- [x] @mikeastock
- [x] @mykalu
- [x] @nocduro
- [x] @pepyakin
- [x] @pitdicker 
- [x] @portal-chan
- [x] @rap2hpoutre
- [x] @rofrol
- [x] @sb89
- [x] @skierpage 
- [x] @sotocodes 
- [x] @tyoc213
- [x] @yandexx

abitrolly:
It is not possible to change existing license without asking all contributors if they agree to that. We can also assume that they didn't care in the first place..

budziq:
We are planning to ask all contributors (at least above typo fix ones) for their consent. At the moment we are blocked with this mostly due to lack of time.

Ophirr33:
Is this the right place to give consent? If so, my consent is given.

dhardy:
[...]

PR #234, opened 2017-07-10
Adds 'handle errors correctly in main' example.

sotocodes:
Issue #214

Note: This example has the category-badge **rust patterns** even though in crates.io error-chain does not belong to any category.

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -12,7 +12,7 @@
 | [Declare lazily evaluated constant][ex-lazy-constant] | [![lazy_static-badge]][lazy_static] | [![cat-caching-badge]][cat-caching] [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
-| [Handle errors correctly in main][error-handling] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
+| [Handle errors correctly in main][ex-simple-error-handling] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 
 
 [ex-std-read-lines]: #ex-std-read-lines
@@ -412,40 +412,38 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
-[error-handling]: #error-handling
-<a name="error-handling"></a>
+[ex-simple-error-handling]: #ex-simple-error-handling
+<a name="ex-simple-error-handling"></a>
 ## Handle errors correctly in main
 
 [![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
 
-Uses [error-chain] to handle error occurred when trying to open a file that does not exist.
+Handles the error that occur when we try to open a file that does not exist. We do this using [error-chain], a library that takes care of a lot of boilerplate code that an user would have to write in order to handle erros in Rust. For more details please check the chapter [Error Handling] of TRPL.
 
-First, we call `error_chain!{}` inside a new module named  _errors_. This will create the default types `Error`, `ErrorKind`, `ResultExt` and ` Result`, and store them in `errors`. To be able to use these types we must call `use errors::*`.
+The [`error_chain!`] macro creates the types [`Error`], [`ErrorKind`], [`Result`] and [`ResultExt`]. Additionaly, we use `Io(sd::io::Error)` for automatic conversion between this error chain and other error types not defined by [`error_chain!`]. In this case, [`std::io::Error`]
 
-To handle the erros that may occur when calling `open_foo()`, it must have return type `Reuslt<()>`. We then use [`if let`] when calling it to detect if an error has occurred, and to handle it in an apropiated way.
+To handle the result of a function, in this case `open_foo`, that function must return a [`Result`]. When calling `open_foo`, we use pattern matching to print an error message if an error occur.
+
 ```rust
 #[macro_use]
 extern crate error_chain;
 
-mod errors {
-    error_chain!{}
-}
-
-use errors::*;
-
-fn main() {
-    if let Err(ref e) = open_foo() {
-        println!("error: {}", e);
-        for e in e.iter().skip(1) {
-            println!("caused by: {}", e);
-        }
+error_chain!{
+    foreign_links {
+        Io(std::io::Error);
     }
 }
 
-fn open_foo() -> Result<()> {
+fn main() {
+    if let Err(ref e) = read_foo() {
+        // handle the error here
+        println!("error: {}", e);
+    }
+}
+
+fn read_foo() -> Result<()> {
     use std::fs::File;
-    File::open("foo.txt")
-        .chain_err(|| "cannot open foo.txt")?;
+    File::open("foo.txt")?;
 
     Ok(())
 }
@@ -465,7 +463,6 @@ fn open_foo() -> Result<()> {
 [cat-os]: https://crates.io/categories/os
 [cat-rust-patterns-badge]: https://badge-cache.kominick.com/badge/rust_patterns--x.svg?style=social
 [cat-rust-patterns]: https://crates.io/categories/rust-patterns
-
 [cat-text-processing-badge]: https://badge-cache.kominick.com/badge/text_processing--x.svg?style=social
 [cat-text-processing]: https://crates.io/categories/text-processing
 
@@ -483,8 +480,8 @@ fn open_foo() -> Result<()> {
 [regex-badge]: https://badge-cache.kominick.com/crates/v/regex.svg?label=regex
 [memmap]: https://docs.rs/memmap/
 [memmap-badge]: https://badge-cache.kominick.com/crates/v/memmap.svg?label=memmap
-[error-chain]: https://docs.rs/error-chain/0.10.0/error_chain/
-[error-chain-badge]: https://img.shields.io/crates/v/error-chain.svg?label=error-chain
+[error-chain]: https://docs.rs/error-chain/*/error_chain/
+[error-chain-badge]: https://badge-cache.kominick.com/crates/v/error-chain.svg?label=error-chain
 
 <!-- API links -->
 
@@ -509,8 +506,14 @@ fn open_foo() -> Result<()> {
 [`MutexGuard`]: https://doc.rust-lang.org/std/sync/struct.MutexGuard.html
 [`Mmap::as_slice`]: https://docs.rs/memmap/*/memmap/struct.Mmap.html#method.as_slice
 [`seek`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.seek
-[`if let`]: https://doc.rust-lang.org/book/first-edition/if-let.html
+[`error_chain!`]: https://docs.rs/error-chain/*/error_chain/#declaring-error-types
+[`Error`]: https://docs.rs/error-chain/*/error_chain/example_generated/struct.Error.html
+[`ErrorKind`]: https://docs.rs/error-chain/*/error_chain/trait.ChainedError.html#associatedtype.ErrorKind
+[`Result`]: https://docs.rs/error-chain/*/error_chain/example_generated/type.Result.html
+[`ResultExt`]: https://docs.rs/error-chain/*/error_chain/example_generated/trait.ResultExt.html
+[`std::io::Error`]: https://doc.rust-lang.org/std/io/struct.Error.html
 
 <!-- Reference -->
 
 [race-condition-file]: https://en.wikipedia.org/wiki/Race_condition#File_systems
+[Error Handling]: https://doc.rust-lang.org/book/second-edition/ch09-00-error-handling.html
--- a/src/basics.md
+++ b/src/basics.md
@@ -12,6 +12,7 @@
 | [Declare lazily evaluated constant][ex-lazy-constant] | [![lazy_static-badge]][lazy_static] | [![cat-caching-badge]][cat-caching] [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
+| [Handle errors correctly in main][error-handling] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 
 
 [ex-std-read-lines]: #ex-std-read-lines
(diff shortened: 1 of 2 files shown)


lilianmoraru:
It is ok, this is the category mentioned in `Cargo.toml`(I guess that's where you got the category from).

budziq:
@somebrel 
> Note: This example has the category-badge rust patterns even though in crates.io error-chain does not belong to any category.

As @lilianmoraru mentioned. You were spot on. `rust patterns` is the category used in latest error-chain master (but not yet pushed to crates.io)

lilianmoraru:
@budziq The category was added 6 months ago: https://github.com/brson/error-chain/blame/master/Cargo.toml#L10
Probably just a `crates.io` bug or something like this.

somebrel:
Perfect 💯

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
 
+[error-handling]: #error-handling
+<a name="error-handling"></a>
+## Handle errors correctly in main
+
+[![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
+
+Uses [error-chain] to handle error occurred when trying to open a file that does not exist.
+
+First, we call `error_chain!{}` inside a new module named  _errors_. This will create the default types `Error`, `ErrorKind`, `ResultExt` and ` Result`, and store them in `errors`. To be able to use these types we must call `use errors::*`.
```
budziq: Please paste identifiers without any suffixes like `()` or `{}` and make sure these are linked to docs like in other examples.

budziq: Please linkify the mentioned identifiers/traits

budziq: > inside a new module named  _errors_. 

Please see my later comment about error module. I would remove it from this basic example and just use `error_chain!` in gloal scope.

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+
+To handle the erros that may occur when calling `open_foo()`, it must have return type `Reuslt<()>`. We then use [`if let`] when calling it to detect if an error has occurred, and to handle it in an apropiated way.
+```rust
+#[macro_use]
+extern crate error_chain;
+
+mod errors {
+    error_chain!{}
```
budziq: I would define at least a single error within the chain not to use the `chain_err` on str which is convenient but might not be the best pattern to show newbies.

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+        for e in e.iter().skip(1) {
+            println!("caused by: {}", e);
+        }
+    }
+}
+
+fn open_foo() -> Result<()> {
+    use std::fs::File;
+    File::open("foo.txt")
+        .chain_err(|| "cannot open foo.txt")?;
```
budziq: `chain_err` is an important method. I would mention it in the textual description.

budziq:
> Probably just a crates.io bug or something like this.
Yep you're most likely right :smile:

somebrel:
@budziq  thanks for the feedback, I'll fix it

src/basics.md
somebrel: What do you think about this one @budziq?

```rust
#[macro_use]
extern crate error_chain;

error_chain!{
    foreign_links {
        Io(::std::io::Error);
    }   
}

fn main() {
    if let Err(ref e) = read_foo() {
        // handle the error here
        println!("error: {}", e); 
    }   
}

fn read_foo() -> Result<()> {
    use std::fs::File;
    File::open("foo.txt")?;

    Ok(())
}

```

budziq: @somebrel looks good to me :+1: we can defer showing `chain_err` with custom error and error unrolling to a more advanced example. But its your story (I'm just here to help) I am game either way.

I would only drop the `::` prefix

somebrel: alright, thanks @budziq

src/basics.md
```diff
@@ -412,40 +412,38 @@ fn run() -> Result<()> {
...
+<a name="ex-simple-error-handling"></a>
 ## Handle errors correctly in main
 
 [![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
 
-Uses [error-chain] to handle error occurred when trying to open a file that does not exist.
+Handles the error that occur when we try to open a file that does not exist. We do this using [error-chain], a library that takes care of a lot of boilerplate code that an user would have to write in order to handle erros in Rust. For more details please check the chapter [Error Handling] of TRPL.
 
-First, we call `error_chain!{}` inside a new module named  _errors_. This will create the default types `Error`, `ErrorKind`, `ResultExt` and ` Result`, and store them in `errors`. To be able to use these types we must call `use errors::*`.
+The [`error_chain!`] macro creates the types [`Error`], [`ErrorKind`], [`Result`] and [`ResultExt`]. Additionaly, we use `Io(sd::io::Error)` for automatic conversion between this error chain and other error types not defined by [`error_chain!`]. In this case, [`std::io::Error`]
```
budziq: I would avoid going into details of what goes behind the scenes. The reader interested in the internals will go into linked docs. We should keep the noise to minimum with this basic example. Please remove the mention  of " [`Error`], [`ErrorKind`], [`Result`] and [`ResultExt`]. and their links".

Also please see if you can make the description a little shorter

src/basics.md
```diff
@@ -509,8 +506,14 @@ fn open_foo() -> Result<()> {
 [`MutexGuard`]: https://doc.rust-lang.org/std/sync/struct.MutexGuard.html
 [`Mmap::as_slice`]: https://docs.rs/memmap/*/memmap/struct.Mmap.html#method.as_slice
 [`seek`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.seek
-[`if let`]: https://doc.rust-lang.org/book/first-edition/if-let.html
+[`error_chain!`]: https://docs.rs/error-chain/*/error_chain/#declaring-error-types
+[`Error`]: https://docs.rs/error-chain/*/error_chain/example_generated/struct.Error.html
```
budziq: The links with identifiers from std type linking to error_chain generated example are a nice idea in theory but will create confusion in newbie readers as well as pollute the link namespace. I would get rid of them

somebrel:
Thanks for the feeedback @budziq, I can see how the example is getting better.

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+#[macro_use]
+extern crate error_chain;
+
+error_chain!{
+    foreign_links {
+        Io(std::io::Error);
+    }
+}
+
+fn main() {
```
budziq: I would move the `fn main` to bottom of the example.

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+
+[![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
+
+Handles error that occur when we try to open a file that does not exist. We do this using [error-chain], a library that takes care of a lot of boilerplate code needed in order to [handle errors in Rust].
+
+We call `Io(std::io::Error)` inside `foreign_links` to convert the type `std::io::Error` into error-chain's custom version of it.
+
+To handle the result of a function that might return an error, in this example `open_foo`, it must return a `Result`. To handle that result, we use pattern matching on it when calling the function. In this example we print an error message if that result was an error.
+
+
```
budziq: I would remove the additional newlines keeping just one

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+    if let Err(ref e) = read_foo() {
+        // handle the error here
+        println!("error: {}", e);
+    }
+}
+
+fn read_foo() -> Result<()> {
+    use std::fs::File;
+    File::open("foo.txt")?;
+
```
budziq: - I would also add `read_to_string` here to show multiple operations.
- We may even want to actually return the Result<String> and print it in main in `Ok` variant (but I leave it to you)

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+<a name="ex-simple-error-handling"></a>
+## Handle errors correctly in main
+
+[![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
+
+Handles error that occur when we try to open a file that does not exist. We do this using [error-chain], a library that takes care of a lot of boilerplate code needed in order to [handle errors in Rust].
+
+We call `Io(std::io::Error)` inside `foreign_links` to convert the type `std::io::Error` into error-chain's custom version of it.
+
+To handle the result of a function that might return an error, in this example `open_foo`, it must return a `Result`. To handle that result, we use pattern matching on it when calling the function. In this example we print an error message if that result was an error.
```
budziq: I would avid repetition `To handle the result,` and `To handle that result,`. This might be rephrased and further simplified

src/basics.md
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
 
+[ex-simple-error-handling]: #ex-simple-error-handling
+<a name="ex-simple-error-handling"></a>
+## Handle errors correctly in main
+
+[![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
+
+Handles error that occur when we try to open a file that does not exist. We do this using [error-chain], a library that takes care of a lot of boilerplate code needed in order to [handle errors in Rust].
+
+We call `Io(std::io::Error)` inside `foreign_links` to convert the type `std::io::Error` into error-chain's custom version of it.
```
budziq: > We call `Io(std::io::Error)` 

I wouldn't use `call` in this context. It is actually a definition.

> `foreign_links` to convert the type `std::io::Error` into error-chain's custom version of it

I would rephrase it "[`foreign_links`](https://brson.github.io/error-chain/error_chain/index.html#foreign-links) to allow automatic conversion from [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html) into [`error_chain!`](https://brson.github.io/error-chain/error_chain/macro.error_chain.html) defined type implementing the [`Error`](https://doc.rust-lang.org/std/error/trait.Error.html) trait."

Please note the that I've added 4x links here.

src/basics.md
somebrel: Hey @budziq, that's a good Idea... I modified example:
```rust
#[macro_use]
extern crate error_chain;

error_chain!{
    foreign_links {
        Io(std::io::Error);
    }
}

fn read_foo() -> Result<(String)> {
    use std::fs::File;
    use std::io::Read;

    let mut foo = File::open("foo.txt")?;
    let mut foo_content = String::new();
    foo.read_to_string(&mut foo_content)?;
    Ok(foo_content)
}

fn main() {
    match read_foo() {
        Ok(foo_content) => println!("{}", foo_content),
        Err(error) => println!("error: {}", error),
    };
}
```

budziq: Very nice :+1: 

You can skip the `foo` variable, and there is no need to have `String` in parens. Also the use statements would be on top level.
```rust
use std::fs::File;
use std::io::{Result, Read};

fn read_foo() -> Result<String> {
    let mut foo_content = String::new();
    File::open("foo.txt")?.read_to_string(&mut foo_content)?;
    Ok(foo_content)
}

fn main() {
    match read_foo() {
        Ok(foo_content) => println!("{}", foo_content),
        Err(error) => println!("error: {}", error),
    };
}
```

Unfortunately this example can be done fully without any need for error_chain (as presented above)

So we might need to think on a relatively non contrived way to introduce another error variant except  `Io(std::io::Error);` to justify error_chain usage :smile_cat:. You can either:
- define your own app error and return it if foo_content does not match some arbitrary predicate like (starts_with("#"))
- get the filename from env var (and we already have another error variant)
- parse the foo_contents into something like an int
- pass the path by value and manipulate it (for instance strip)
- your idea

src/basics.md
somebrel: I changed line 425 to be

 _"open_foo returns a Result, which is the type used for returning and propagating errors. We use pattern matching in main to handle that result, which in our case is either an error or the content of foo.txt."_

What do you thin @budziq?

budziq: sounds ok. please remember to linkify the `Result` identifier.
Also I find the Result definition "the type used for returning and propagating errors" a bit awkward (I know that is taken from official docs but I think that it is lacking and will confuse newbies). What we would like to convey is that:
"Result, which is the type that represents a possibility of failure" (or something along these lines)

budziq:
Hi @Phaiax thanks for all the submissions! 

Unfortunately I'll not be able to follow up with anything resembling review for the next two weeks due to holidays 😞 . I may drop some comments when Im on mobile later and I've asked other maintainers to drop from time to time and look into the PR's.

somebrel:
Hey @budziq, thanks for letting me know. Enjoy your holidays! 😃

budziq:
Hi @somebrel. I'm back from holidays so I should be able to give some more hands on help here. Do you need any assistance with this example?

somebrel:
Welcome back @budziq. I been busy these days but I'll be working on this example Tuesday

somebrel:
Hey @budziq  I went for the application error because I think it is a common one. 

```rust
#[macro_use]
extern crate error_chain;

use std::io::BufReader;
use std::io::BufRead;
use std::fs::File;

error_chain!{
    errors {
        ConfigHeaderMissing {
           description("Invalid header.")
           display("Invalid header. Please check foo.config, line 1")
        }
    }
}

fn check_header() -> Result<bool> {
    let f = File::open("foo.config").unwrap();
    let file = BufReader::new(&f);

    if "#foo.config" == file.lines().next().unwrap().unwrap() {
        Ok(true)
    } else {
        Err(ErrorKind::ConfigHeaderMissing.into())
    }
}

fn main() {
    match check_header() {
        Ok(status) => println!("Valid header: {}", status),
        Err(error) => println!("Error: {}", error),
    };
}

**Comment by budziq** [comment]
Hi @somebrel,

Nice idea but I would use it on one of the more advanced examples for error_chain (there are two more progressing in complexity) also we will need to cleanup `check_header` a little (turn `unwrap`s into `?`s , more natural parsing with `nth` or just compare some [env var](https://doc.rust-lang.org/std/env/fn.var.html) with a set of valid values instead).

But for now I would cleanup the example that we have discussed here previously as it was almost perfect.

**Comment by somebrel** [comment]
alright @budziq  I am back to the previous example.

When I try to parse the content of `foo` into an  `i32` I get `ParseIntError { kind: InvalidDigit }`. The problem is that the content of `foo` is not `12345` but `12345\n`... I am not sure how I use error_chain to customize this error msg :/

**Comment by budziq** [comment]
> I am back to the previous example.

Lets keep the new one ready for next example.

> The problem is that the content of foo is not 12345 but 12345\n... 

You'll have to use `trim` to remove whitespace. 

> I am not sure how I use error_chain to customize this error msg :/

No need to customize the error message in this first example. Let's just handle the standard error correctly.

**Comment by budziq** [comment]
Hi @sotocodes, do you need any help with finishing the work here?

File: src/basics.md [hunk 11]
```diff
@@ -411,6 +412,45 @@ fn run() -> Result<()> {
...
+error_chain!{
+    foreign_links {
+        Io(std::io::Error);
+    }
+}
+
+fn main() {
+    if let Err(ref e) = read_foo() {
+        // handle the error here
+        println!("error: {}", e);
```
lilianmoraru: This should use `eprintln!()`

lilianmoraru: Also, I prefer the `quick_main!` macro(it also handles writing to `stderr` and panicking when it can't), but I'll let @budziq decide here if the explicit error handling is more appropriate here

budziq: > This should use eprintln!()

Good point

> Also, I prefer the quick_main! macro(it also handles writing to stderr and panicking when it can't)

I would opt for explicit error handling this time for educational purposes as this aims to be a super basic example. Reader may refer to `quick_main!` usage in nearly every other example.

I would envision for the main here to mirror (but using `eprintln!`)
https://github.com/dtolnay/cargo-expand/blob/b05fcdabf48fa34c90343b1baa8777b9e65684db/src/main.rs#L18

which pretty much boils down to what `quick_main!` expands to.

budziq:
Hi @sotocodes do you need help in going further with this example?

sotocodes:
Hi @budziq, I have just a few days to finish my thesis. After that I'll try to finish it... sorry for the delay :/

budziq:
>  sorry for the delay :/

No problem man. Just checking if you needed any help. Good luck!

budziq:
Hi @sotocodes are you ready to push this though the finish line?

sotocodes:
Hi @budziq, I haven't been able to code Rust in a while and I'm not sure when I'll be back. So I won't be able to finish these one. Sorry :(

budziq:
@sotocodes No worries mate! I wouldn't want for all of this good work go to waste! 
So if you'll give your consent I would like to add the final touches myself and merge your PR in the next few days (the contribution will stil be yours). Are you ok with that?

Also this would make you a full fledged  contributor requiring your vote on upcoming license change to CC0. So please consider giving also [your consent to re-licensing here](https://github.com/rust-lang-nursery/rust-cookbook/issues/209)

sotocodes:
Sure, it would be great if you could finish it :) 

Thanks

ludwigpacifici:
Hey @budziq, If I want to finish this PR, how should I proceed?

budziq:
It would be super cool If you'd get this through the finish line @ludwigpacifici :+1: 

> If I want to finish this PR, how should I proceed?

1. I'd git pull this PR locally with `git fetch upstream pull/234/head`
2. rebase it to latest master (possibly solving some minor conflicts)
3. then commit and push the work to a separate branch on your fork and make a PR of your own (unfortunately no one except for the author is able to update this PR) marking it as work in progress and linking to the old PR
4. address the remaining [review comments to this PR](https://github.com/rust-lang-nursery/rust-cookbook/pull/234/files) and push the update to your PR with separate commit. 

If you'll have any questions or problems please do not hesitate to ask me :)

ludwigpacifici:
Hey @budziq !

Apologize for the wait and thanks for the detailed reply. I have some free time. I'll have a look!

PR #239, opened 2017-07-14
File Loop app example

### LINKED ISSUE #207: "find loops in filesystem"
Use the [same-file](https://github.com/BurntSushi/same-file/) check if one of the files in given directory does not point to one of the path ancestors.

We would prefer not to use `walkdir` crate which already does al the heavy lifting here (using `same-file`) internally.

This might be a basis for a more involved example such as graphing the filesystem with petgraph but I would suggest to just implement the first trivial part of the example and we might think about creating a separate more involved one with petgraph.

cc https://github.com/brson/rust-cookbook/issues/205

Ophirr33:
This is a rough draft of the file loop example. I'd love some feedback on how to make it more concise. Right now there's a lot of boiler plate on iterating through the file system, and check parent paths. Though, the file system boilerplate could help make a better case for walkdir. @budziq , thanks for the tip on how to make a sym link loop. Fixes #207

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -6,6 +6,7 @@
 | [Decompress a tarball][ex-tar-decompress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Compress a directory into a tarball][ex-tar-compress] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
 | [Decompress a tarball while removing a prefix from the paths][ex-tar-strip-prefix] | [![flate2-badge]][flate2] [![tar-badge]][tar] | [![cat-compression-badge]][cat-compression] |
+| [Find loops for a given path][ex-find-file-loops] | [![same_file-badge]][same_file] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find duplicate file names][ex-dedup-filenames] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Recursively find all files with given predicate][ex-file-predicate] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Traverse directories while skipping dotfiles][ex-file-skip-dot] | [![walkdir-badge]][walkdir] | [![cat-filesystem-badge]][cat-filesystem] |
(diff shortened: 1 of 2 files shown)


src/app.md
```diff
@@ -248,6 +249,88 @@ fn run() -> Result<()> {
...
 ```
 
+[ex-find-file-loops]: #ex-find-file-loops
+<a name="ex-find-file-loops"></a>
+## Find loops within a directory
+
+[![same_file-badge]][same_file] [![cat-filesystem-badge]][cat-filesystem]
+
+Use [`same-file::is_same_file`] to detect loops within a directory.
+This manually iterates through the filesystem via the [`std::path`] module.
```
budziq: no need to explicitly mention the `std::path` module.

src/app.md
```diff
@@ -251,83 +251,38 @@ fn run() -> Result<()> {
...
-            return Ok(None);
-        } else {
-            // Check the next parent
-            parent = &file_tree[parent];
+// Check this path against all of its parents
+fn contains_loop(path: &Path) -> io::Result<bool> {
+    let mut path_buf = path.to_path_buf();
+    while path_buf.pop() {
+        if is_same_file(&path_buf, path)? {
+            return Ok(true);
```
budziq: - the contains_loop tests only if the whole path is pointing to one of its ancestors while we would like to find all loops. calling it repeatedly on poped path segments will do the trick
- How about returning the result with optional looped path segments? (this might not come out that clean so I'll leave it up to you)

src/app.md
```diff
@@ -246,6 +247,42 @@ fn run() -> Result<()> {
...
+```rust,no_run
+extern crate same_file;
+
+use std::io;
+use std::path::Path;
+use same_file::is_same_file;
+
+// Check this path against all of its parents
+fn contains_loop(path: &Path) -> io::Result<bool> {
+    let mut path_buf = path.to_path_buf();
```
budziq: we may use AsRef<Path> here as well to make it usable with &str

Ophirr33:
Hey @budziq, thanks for the feedback, I definitely misread what the example was going for.

src/app.md
Ophirr33: Not 100% sure if I'm following. Basically it should detect a loop even when you don't specify the looped directory, right?

Ophirr33:
@budziq I lost your requested changes since I force pushed. How's it looking now?

Ophirr33:
@budziq It checks all paths and will return what paths it's looping between now. Let me know what else I can do

src/app.md
```diff
@@ -248,6 +249,52 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-find-file-loops]: #ex-find-file-loops
+<a name="ex-find-file-loops"></a>
+## Find loops for a given path
+
+[![same_file-badge]][same_file] [![cat-filesystem-badge]][cat-filesystem]
+
+Use [`same-file::is_same_file`] to detect loops for a given path.
```
budziq: Please change the link `same-file::is_same_file` to `same_file::is_same_file`

Ophirr33:
@budziq Done!

budziq:
@Ophirr33 Well done!

Issue #244, opened 2017-07-14
add "Filter a log file by matching multiple regular expressions" example


budziq:
This example should use [regex](https://crates.io/crates/regex) crate.
It should showoff  [`RegexSet`](https://doc.rust-lang.org/regex/regex/struct.RegexSet.html) usage.

cc https://github.com/brson/rust-cookbook/issues/217

See [CONTRIBUTING.md](https://github.com/brson/rust-cookbook/blob/master/CONTRIBUTING.md) for details.

yandexx:
Hi! I'll try this one.

I've made a working example so I'll document it properly and make a PR.

I presume this goes into the Basics section.

budziq:
> I presume this goes into the Basics section.

@yandexx  Right. For now we are grouping these into basics and when the section grows too much we will think about good way to split.

PR #251, opened 2017-07-19
Filter a log file by matching multiple regular expressions

yandexx:
For Issue #244 

A few concerns I have:
- Is it even a good example to only filter by a couple words?
- I used `RegexSetBuilder` instead of `RegexSet`, I think it's a friendlier way to force case insensitive search and shows the builder pattern.
- Is it potentially confusing to use shadowing there with `let line = line?;`

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -9,6 +9,7 @@
 | [Generate random numbers with normal distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Run an external command and process stdout][ex-parse-subprocess-output] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
+| [Filter a log file by matching multiple regular expressions][ex-regex-filter-log] | [![regex-badge]][regex] | [![cat-text-processing-badge]][cat-text-processing]
 | [Declare lazily evaluated constant][ex-lazy-constant] | [![lazy_static-badge]][lazy_static] | [![cat-caching-badge]][cat-caching] [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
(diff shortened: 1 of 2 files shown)


yandexx:
OK the test is also failing since there's no actual log file. Should I then create a several line bogus log file at the beginning of the example and hide it away?

budziq:
Hi @yandexx thanks for all the submissions!

Unfortunately I'll not be able to follow up with anything resembling review for the next two weeks due to holidays 😞 . I may drop some comments when Im on mobile later and I've asked other maintainers to drop from time to time and look into the PR's

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
+containing “warning” or “error” (case insensitive). A [`regex::RegexSet`] is built
+with [`regex::RegexSetBuilder`] using the builder pattern.
+
+```rust
+# #[macro_use]
+# extern crate error_chain;
+extern crate regex;
+
+# use std::fs::File;
+# use std::io::{BufReader, BufRead};
```
budziq: I would not hide these use statements. We generally hide only the error boilerplate or support code that is far from the point.

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
 
+[ex-regex-filter-log]: #ex-regex-filter-log
+<a name="ex-regex-filter-log"></a>
+## Filter a log file by matching multiple regular expressions
+
+[![regex-badge]][regex] [![cat-text-processing-badge]][cat-text-processing]
+
+Reads a file named `application.log` and only outputs the lines 
+containing “warning” or “error” (case insensitive). A [`regex::RegexSet`] is built
+with [`regex::RegexSetBuilder`] using the builder pattern.
```
budziq: > using the builder pattern

I would remove this last part of sentence. The RegexSetBuilder indicates it already.

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
+
+```rust
```
budziq: make the example `no_run` to avoid test fail but still check for compilation.

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
+    let log_path = "application.log";
+    let log_file = File::open(log_path)?;
+    let buffered = BufReader::new(log_file);
+
+    let set = RegexSetBuilder::new(&[r"warning", r"error"])
+                .case_insensitive(true)
+                .build()?;
+
+    for line in buffered.lines() {
+        let line = line?;
```
budziq: in general shadowing is ok but i would suggest using iterator combinators (eg filter_map) to avoid bailing out on first error. Please see the [following example](https://brson.github.io/rust-cookbook/app.html#ex-tar-strip-prefix) for details.

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
+fn run() -> Result<()> {
+    let log_path = "application.log";
+    let log_file = File::open(log_path)?;
+    let buffered = BufReader::new(log_file);
+
+    let set = RegexSetBuilder::new(&[r"warning", r"error"])
+                .case_insensitive(true)
+                .build()?;
+
+    for line in buffered.lines() {
```
budziq: at least one if not both  log_file / buffered variables can be removed if you instantiate BufferedReader here with File::Open as an argument.

src/basics.md
```diff
@@ -270,6 +271,54 @@ fn run() -> Result<()> {
...
+#         Regex(regex::Error);
+#     }
+# }
+#
+fn run() -> Result<()> {
+    let log_path = "application.log";
+    let log_file = File::open(log_path)?;
+    let buffered = BufReader::new(log_file);
+
+    let set = RegexSetBuilder::new(&[r"warning", r"error"])
```
budziq: how about adding more complicated regexes here to make the example more real life. Just think of the last loggs you have greped and construct something similar ;)

src/basics.md
```diff
@@ -278,17 +278,20 @@ fn run() -> Result<()> {
...
 
 Reads a file named `application.log` and only outputs the lines 
-containing “warning” or “error” (case insensitive). A [`regex::RegexSet`] is built
-with [`regex::RegexSetBuilder`] using the builder pattern.
+containing “version X.X.X” or some IP address followed by port 443 (e.g. “192.168.0.1:443”).
 
-```rust
+A [`regex::RegexSet`] is built with [`regex::RegexSetBuilder`].
+Since backslashes are very common in regular expressions, using
+[raw string literals][raw-string-literals] make them more readable.
```
budziq: no need for link renaming here. you can define the link directly as you would like it displayed at the footer.

`[raw string literals]: https://doc.rust-lang.org/reference/tokens.html#raw-string-literals`

src/basics.md
```diff
@@ -270,6 +271,58 @@ fn run() -> Result<()> {
...
+#         Io(std::io::Error);
+#         Regex(regex::Error);
+#     }
+# }
+# 
+fn run() -> Result<()> {
+    let log_path = "application.log";
+    let buffered = BufReader::new(File::open(log_path)?);
+
+    let set = RegexSetBuilder::new(&[r#"version "\d\.\d\.\d""#, r#"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}:443"#])
```
budziq: please break the line into shorter ones and run rustfmt on the example.
also please add some third regex for good measure (how about Some warning and some specified text later on?)

src/basics.md
```diff
@@ -270,6 +271,58 @@ fn run() -> Result<()> {
...
+[raw string literals][raw-string-literals] make them more readable.
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+extern crate regex;
+
+use std::fs::File;
+use std::io::{BufReader, BufRead};
+
```
budziq: no need for a newline here

yandexx:
Thanks, glad to be of help!

brson:
Thanks @yandexx

PR #252, opened 2017-07-19
Extract phone numbers from text

### LINKED ISSUE #241: add "Extract phone numbers from text" example
This example should use [regex](https://crates.io/crates/regex) crate 
cc https://github.com/brson/rust-cookbook/issues/217

AndyGauge:
Fixes #241

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -13,6 +13,7 @@
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Define and operate on a type represented as a bitfield][ex-bitflags] | [![bitflags-badge]][bitflags] | [![cat-no-std-badge]][cat-no-std] |
+| [Extract phone numbers from text][ex-phone] | [![regex-badge]][regex] | [![cat-text-processing-badge]][cat-text-processing] |
 
 
 [ex-std-read-lines]: #ex-std-read-lines
@@ -470,6 +471,107 @@ fn main() {
 }
 ```
 
+[ex-phone]: #ex-phone
+<a name="ex-phone"></a>
+## Extract phone numbers from text
+
+Processes a string of text using [`Regex::captures_iter`] to capture multiple
+phone numbers.  The example here is for US convention phone numbers.
+
+```rust, no_run
+# #[macro_use]
+# extern crate error_chain;
+extern crate regex;
+
+use regex::Regex;
+use std::fmt;
+
+# error_chain!{
+#     foreign_links {
+#         Regex(regex::Error);
+#         Io(std::io::Error);
+#     }
+# }
+#
+#[derive(PartialEq, PartialOrd, Debug)]
+struct PhoneNumber {
+    area: &'static str,
+    exchange: &'static str,
+    subscriber: &'static str,
+}
+
+// Allows printing phone numbers based on country convention.
+impl fmt::Display for PhoneNumber {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "1 ({}) {}-{}",
+            &self.area,
+            &self.exchange,
+            &self.subscriber
+        )
+    }
+}
+
+fn run() -> Result<()> {
+    let phone_text = "
+    +1 505 881 9292 (v) +1 505 778 2212 (c) +1 505 881 9297 (f)
+    (202) 991 9534
+    Alex 5553920011
+    1 (800) 233-2010
+    1.299.339.1020";
+    let re = Regex::new(
+        r"(?x)
+                        (?:\+?1)?                         # Country Code Optional
+                        [\s\.]?
+                        (([2-9]\d{2})|\(([2-9]\d{2})\))  # Area Code
+                        [\s\.\-]?
+                        ([2-9]\d{2})                     # Exchange Code
+                        [\s\.\-]?
+                        (\d{4})                          #Subscriber Number",
+    )?;
+    let mut phone_numbers = re.captures_iter(phone_text).map(|cap| {
+        // Area code populates either capture group 2 or 3.  Group 1 contains optional paranthesis.
+        PhoneNumber {
+            area: if cap.get(2) == None {
+                cap.get(3).map_or("", |m| m.as_str())
+            } else {
+                cap.get(2).map_or("", |m| m.as_str())
+            },
+            exchange: cap.get(4).map_or("", |m| m.as_str()),
+            subscriber: cap.get(5).map_or("", |m| m.as_str()),
+        }
+    });    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (505) 881-9292".to_owned())
+    );
+    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (505) 778-2212".to_owned())
+    );
+    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (505) 881-9297".to_owned())
+    );
+    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (202) 991-9534".to_owned())
+    );
+    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (555) 392-0011".to_owned())
+    );
+    assert_eq!(
+        phone_numbers.next().map(|m| m.to_string()),
+        Some("1 (800) 233-2010".to_owned())
+    );
+    Ok(())
+}
+#
+# quick_main!(run);
+
+```
+
 <!-- Categories -->
 
 [cat-no-std-badge]: https://badge-cache.kominick.com/badge/no_std--x.svg?style=social
@@ -523,6 +625,7 @@ fn main() {
 [`Rng::gen_range`]: https://doc.rust-lang.org/rand/rand/trait.Rng.html#method.gen_range
 [`rand::Rand`]: https://doc.rust-lang.org/rand/rand/trait.Rand.html
 [`Regex`]: https://doc.rust-lang.org/regex/regex/struct.Regex.html
+[`Regex::captures_iter`]: https://doc.rust-lang.org/regex/regex/struct.Regex.html#method.captures_iter
 [`Output`]: https://doc.rust-lang.org/std/process/struct.Output.html
 [`Command`]: https://doc.rust-lang.org/std/process/struct.Command.html
 [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
--- a/src/intro.md
+++ b/src/intro.md
@@ -31,6 +31,7 @@ community. It needs and welcomes help. For details see
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Define and operate on a type represented as a bitfield][ex-bitflags] | [![bitflags-badge]][bitflags] | [![cat-no-std-badge]][cat-no-std] |
+| [Extract phone numbers from text][ex-phone] | [![regex-badge]][regex] | [![cat-text-processing-badge]][cat-text-processing] |
 
 ## [Encoding](encoding.html)
 
@@ -245,6 +246,7 @@ Keep lines sorted.
 [ex-paginated-api]: net.html#ex-paginated-api
 [ex-parse-subprocess-output]: basics.html#ex-parse-subprocess-output
 [ex-percent-encode]: encoding.html#ex-percent-encode
+[ex-phone]: basics.html#ex-phone
 [ex-rand-custom]: basics.html#ex-rand-custom
 [ex-rand-dist]: basics.html#ex-rand-dist
 [ex-rand-float]: basics.html#ex-rand-float


src/basics.md
```diff
@@ -470,6 +471,48 @@ fn main() {
...
+#         Regex(regex::Error);
+#         Io(std::io::Error);
+#     }
+# }
+
+fn run() -> Result<()> {
+    println!("Input a phone number");
+    let mut phone_input = String::new();
+    // Regular expression is looking for 3 groups of digits, capturing them
+    let re = Regex::new(r".*([2-9]\d{2}).*(\d{3}).*(\d{4})")?;
```
budziq: hi  this regex will catch false positives "fiz 233 baz 345  bar 6789" so this might not be the best one.
on the other hand the allowed NANP spec is ridiculously complex and even a subset of it spawns not very nice:
`^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]?\d{3}[\s.-]?\d{4}$` (copied via mobile so it might vie utterly broken)

So lets meet halfway and agree that phone numbers are optional country region code + 3 groups of digits optionally separated by either single space dot or `-`. Country region code is either preceeded by `+` or `00` If its still too complex we might pair it down further as long as it will not give false positives.

Also it might be a good place to showoff named capture groups (at least for the country  code)

And we would still mention in the example that this will work only for nicely formatted numbers.

src/basics.md
```diff
@@ -470,6 +471,48 @@ fn main() {
...
+    stdin().read_line(&mut phone_input)?;
```
budziq: I would suggest to open some arbitrary text file and iterate over all of its lines and phone numbers.

Also we might want to collect the numbers into a set in order to deduplicate them prior to printing.

src/basics.md
```diff
@@ -470,6 +471,48 @@ fn main() {
...
+    // Make sure the regular expression captured something
+    if let Some(captures) = re.captures(&phone_input){
+        let formatted_phone = format!("({}) {}-{}", &captures[1], &captures[2], &captures[3]);
+        println!("Formatted like a boss: {}", formatted_phone);
+    } else {
+        println!("Phone numbers must include area code.  Use numbers and whatever else you want\nFor example +1 (555) 101-9939");
```
budziq: lets remove the lines that are not really to the point

src/basics.md
```diff
@@ -470,6 +471,48 @@ fn main() {
...
+fn run() -> Result<()> {
+    println!("Input a phone number");
+    let mut phone_input = String::new();
+    // Regular expression is looking for 3 groups of digits, capturing them
+    let re = Regex::new(r".*([2-9]\d{2}).*(\d{3}).*(\d{4})")?;
+    stdin().read_line(&mut phone_input)?;
+    // Make sure the regular expression captured something
+    if let Some(captures) = re.captures(&phone_input){
+        let formatted_phone = format!("({}) {}-{}", &captures[1], &captures[2], &captures[3]);
+        println!("Formatted like a boss: {}", formatted_phone);
```
budziq: how about just printing these out instead of formatting? also no need to add additional noise to the output. lets keep the example short and sweet so it can be copy pasted.

Also I leave it up to you if you would consider using PhoneNumber struct with display (please see the external process example)

src/basics.md
AndyGauge: So I've investigated international phone numbers and their format is way different.  Here's the wikipedia article I'm investigating: https://en.wikipedia.org/wiki/National_conventions_for_writing_telephone_numbers
Because it would be easy to create an entire library out of the individual nation's formatting requirements I want to make this only function for US telephone numbers.  (While including the optional +1).  International conventions includes dots, dashes, spaces, and parenthesis optionally between digits.  I think +15551231234 and 5551231234 are both valid phone numbers.  Just that regex alone is actually a bit more complicated than the one you posted via mobile.  (and won't use the ^and$ because we are not validating we are extracting).

Anyway I'll throw something together that eliminates the false positives while looking good enough to reason (while only guaranteeing support for US).

src/basics.md
AndyGauge: Good ideas, but should we just use a static string so that it works in a cookbook?

budziq: you can either provide a short static string (then assert_eq on the results to show the intended output) or a filename (in that case make the example no_run)

src/basics.md
budziq: Ok fair enough 👍 just leave a comment that it'll match only US convention.

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+                         ([2-9]\d{2})               #Exchange Code
+                         [\x20\.\-]*
+                         (\d{4})                    #Subscriber Number")?;
+    let mut phone_numbers = re.captures_iter(phone_text).map(|cap| {
+        PhoneNumber {
+            country: "US".to_owned(),
+            number: [cap[1].to_string(),cap[2].to_string(),cap[3].to_string()].concat()
+        }
+    });
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (505) 881-9292");
```
AndyGauge: Is there any way other than unwrap for this option?

budziq: yep `assert_eq!(phone_numbers.next(), Some("+1 (505) 881-9292"))`

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+                         \+?1?                      #Country Code Optional
+                         [\x20\(\.]*
+                         ([2-9]\d{2})               #Area Code
+                         [\x20\)\.\-]*
+                         ([2-9]\d{2})               #Exchange Code
+                         [\x20\.\-]*
+                         (\d{4})                    #Subscriber Number")?;
+    let mut phone_numbers = re.captures_iter(phone_text).map(|cap| {
+        PhoneNumber {
+            country: "US".to_owned(),
```
budziq: the country does not make much sense here as its unconditionally "US" maybe lets just drop it (also using String as a match tag would not be a best idea).

how about storing area exchange and subscriber codes separately instead?

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+    country: String,
+    number: String,
+}
+
+// This allows printing phone numbers based on country convention.
+impl fmt::Display for PhoneNumber {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.country.as_ref() {
+            "US" => write!(f, "+1 ({}) {}-{}", &self.number[0..3], &self.number[3..6], &self.number[6..10]),
+            _    => write!(f, "{}", self.number),
```
budziq: lets just drop the country and ommit the ±1 from printing . the regex should still conditionally match it

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+# error_chain!{
+#     foreign_links {
+#         Regex(regex::Error);
+#         Io(std::io::Error);
+#     }
+# }
+#
+struct PhoneNumber {
+    country: String,
+    number: String,
```
budziq: maybe the PhoneNumber can refer to not owned str?

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+fn run() -> Result<()> {
+    let phone_text = "
+    +1 505 881 9292
+    (202) 991 9534
+    Alex 5553920011
+    1 (800) 233-2010
+    1.299.339.1020";
+    let re = Regex::new(r"(?x)
+                         \+?1?                      #Country Code Optional
+                         [\x20\(\.]*
```
budziq: the arbitrary number of parentheses matchers look funky.
we probably shoud match  areacode in two alternatives one with one set of paren and other with none.

also there are better ways to match whitespace than 20 ascii code.

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (505) 881-9292");
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (202) 991-9534");
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (555) 392-0011");
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (800) 233-2010");
+    assert!(format!("{}", phone_numbers.next().unwrap()) == "+1 (299) 339-1020");
+    Ok( () )
+}
+#
+# fn main() {
+#    run();
```
budziq: please use  `quick_main!`

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
 }
 ```
 
+[ex-phone]: #ex-phone
+<a name="ex-phone"></a>
+## Extract phone numbers from text
+
+Attempts to process a static string as a list including US phone numbers.  The
+expression is checking that there are 7 digits that come in groups of 3 3 4,
```
budziq: I would not try to explain the actual regex here. we are interested in the crate and its methods and don't try to teach regular expressions

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+separated by spaces, dots, dashes, and parenthesis.  The phone number is then
+tested based on formatting rules.
```
budziq: the fact that we are testing is self evident from the asserts in the code.

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
 }
 ```
 
+[ex-phone]: #ex-phone
+<a name="ex-phone"></a>
+## Extract phone numbers from text
+
+Attempts to process a static string as a list including US phone numbers.  The
```
budziq: This sentence looks a little off, I wouldn't say that the code "attempts" it actually does the parsing. Also I would keep the whole description as a short one sentence mentioning the key method used (`captures_iter`).

src/basics.md
AndyGauge: So when I do that I get 
```
error[E0308]: mismatched types
  --> src/main.rs:53:5
   |
53 |     assert_eq!(phone_numbers.next().unwrap().fmt(), "+1 (505) 881-9292");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `std::result::Result`, found &str
   |
   = note: expected type `std::result::Result<(), std::fmt::Error>`
              found type `&str`
   = note: this error originates in a macro outside of the current crate
```

I'm using the Some value and then running it through format! to show the output of the struct.  Maybe I need to re-write?  I thought the concern about unwrap is that it may panic, but that's the point of an assert--to panic when the values aren't equal.

AndyGauge: When I do that I get this:
```
error[E0308]: mismatched types
  --> src/main.rs:57:5
   |
57 |     assert_eq!(phone_numbers.next(), Some("+1 (299) 339-1020"));
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `PhoneNumber`, found &str
   |
   = note: expected type `std::option::Option<PhoneNumber>`
              found type `std::option::Option<&str>`
   = help: here are some functions which might fulfill your needs:
           - .take()
           - .unwrap()
           - .unwrap_or_default()
   = note: this error originates in a macro outside of the current crate
```

I could do this
```
    assert_eq!(phone_numbers.next(), Some(PhoneNumber {country:"US".to_string(), number:"2993391020".to_string()}));
```
but that doesn't look any better.  I really think if were are trying to offer something someone can use, implementing Display and showing how to have the phone numbers formatted is what people can use.

src/basics.md
AndyGauge: \x20 is required with the (?x) ignore whitespace.  I could use \s but that would match tabs and newlines.  If you are OK with that I can switch it.

budziq:
hi

I no longer see your comment on github. I'm currently on mobile for the
next 1.5weeks so I'll not be able to help with the finer details.

In general you would map the result on the left with to_string and on the
right you would leave Some("the number str".to_owned())

no need to call fmt explicitly or unwrap. the assert_eq is able to compare
complex structs such as Option<String> or something along those lines.
there might be a nicer way but away from the keyboard I cant give a better
suggestion from the top of my head. ay a little with rustc and you'll get
it right :+1:

src/basics.md
AndyGauge: Is this what you mean?
```
[\s\.]*|\(?
```

AndyGauge: \s?[.\(]?

src/basics.md
budziq: that is why I've suggested something like
assert_eq!(phone_numbers.next().map(String::from), Some("+1 (299) 339-1020".to_owned()));

src/basics.md
budziq: - yeah I would go with `\s`
- `\+?1? `  should be moved int an optional group `(\+1)?\s*` so that plus char has to be followed by 1otherwise these two are independent
- I meant that the area codes expression within the parenthesis has to be matched with both opening and closing parens as a whole or | without both of them. currently the match for opening ,closing parens are independent so we can match `(((202 991 9534` .

 I may try to help but writing regexes on mobile is a real pain and I have no way to give you the full answer or even verify if my suggestions are fully sane. So you might try to solving this yourself or wait for me to get back behind a keyboard within next two weeks 😢 .

src/basics.md
```diff
@@ -470,6 +471,80 @@ fn main() {
...
+        match self.country.as_ref() {
+            "US" => write!(f, "+1 ({}) {}-{}", &self.number[0..3], &self.number[3..6], &self.number[6..10]),
+            _    => write!(f, "{}", self.number),
+        }
+    }
+}
+
+fn run() -> Result<()> {
+    let phone_text = "
+    +1 505 881 9292
```
budziq: also this code should allow for multiple numbers per line so let's show it

AndyGauge:
Not sure how my rebase left 2 commits.  I'll try to figure that out.

src/basics.md
```diff
@@ -470,6 +471,84 @@ fn main() {
...
+#[derive (PartialEq, PartialOrd, Debug)]
+ enum CountryCode {
+     US,
+ }
+
+#[derive (PartialEq, PartialOrd, Debug)]
+struct PhoneNumber {
+    country: CountryCode,
+    number: String,
+}
```
AndyGauge: I have tried to do this, but I cannot figure out the lifetime.  Here's the problem:

```
error[E0597]: borrowed value does not live long enough
  --> src\main.rs:49:5
   |
47 |             number: &[area_code,cap[4].to_string(),cap[5].to_string()].concat()
   |                      ---------------------------------------------------------- temporary value created here
48 |         }
49 |     });
   |     ^ temporary value dropped here while still borrowed
...
57 | }
   | - temporary value needs to live until here
```

Here's the struct def:
```
#[derive (PartialEq, PartialOrd, Debug)]
struct PhoneNumber<'a> {
    number: &'a str,
}

// Allows printing phone numbers based on country convention.
impl<'a> fmt::Display for PhoneNumber<'a> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "1 ({}) {}-{}", &self.number[0..3], &self.number[3..6], &self.number[6..10])
    }
}
```
Can you advise?

budziq: There is no clean way of concatenating the str's  without allocation

The easiest way would be to follow my suggestion of separate &str member for each capture (area, extension, subscriber). Then you don't need to concat and allocate

src/basics.md
```diff
@@ -520,32 +520,52 @@ fn run() -> Result<()> {
...
+                        (\d{4})                          #Subscriber Number",
+    )?;
     let mut phone_numbers = re.captures_iter(phone_text).map(|cap| {
-        // Area code populates either capture group 2 or 3.  Capture group 1 contains optional paranthesis.
-        let area_code = if cap.get(2) == None {cap[3].to_string()} else {cap[2].to_string()};
+        // Area code populates either capture group 2 or 3.  Group 1 contains optional paranthesis.
         PhoneNumber {
-            country: CountryCode::US,
-            number: [area_code,cap[4].to_string(),cap[5].to_string()].concat()
+            area: if cap.get(2) == None { &cap[3] } else { &cap[2] },
```
AndyGauge:
```
error[E0597]: `cap` does not live long enough
  --> src\main.rs:56:44
   |
56 |             area: if cap.get(2) == None { &cap[3] } else { &cap[2] },
   |                                            ^^^ does not live long enough
...
60 |     });
   |     - borrowed value only lives until here
   |
   = note: borrowed value must be valid for the static lifetime...
```

Same problem.  cap only exists inside the `iter()` block, I don't see how a reference to a borrowed value is going to work.  This is why I used a String.

budziq: probably we want to use `as_str` instead of depending on &capture `Deref` into a &str. Pease see a soon to be merged PR for suggestions https://github.com/brson/rust-cookbook/pull/247/files

brson:
I rebased and merged. Thanks @AndyGauge

PR #254, opened 2017-07-22
Extract all unique links from a MediaWiki markup

### LINKED ISSUE #245: add "Extract all unique links from a MediaWiki markup" example
This example should use [regex](https://crates.io/crates/regex) crate.

We may also like to provide an alternative implementation with [select](https://crates.io/crates/select) crate

cc https://github.com/brson/rust-cookbook/issues/217

yandexx:
Fixes #245.

- I went for deduplicated Vectors instead of HashMaps, but let me know if HashMap is a better choice after all.

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -17,6 +17,7 @@
 | [POST a file to paste-rs][ex-file-post] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
 | [Listen on unused port TCP/IP][ex-random-port-tcp] | [![std-badge]][std] | [![cat-net-badge]][cat-net] |
 | [Extract all links from a webpage][ex-extract-links-webpage] | [![reqwest-badge]][reqwest] [![select-badge]][select] | [![cat-net-badge]][cat-net] |
+| [Extract all unique links from a MediaWiki markup][ex-extract-mediawiki-links] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
 
 [ex-url-parse]: #ex-url-parse
 <a name="ex-url-parse"/>
@@ -876,6 +877,76 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-extract-mediawiki-links]: #ex-extract-mediawiki-links
+<a name="ex-extract-mediawiki-links"/>
+## Extract all unique links from a MediaWiki markup
+
+[![reqwest-badge]][reqwest] [![regex-badge]][regex] [![cat-net-badge]][cat-net]
+
+Pull the source of a MediaWiki page using [`reqwest::get`] and then
+look for all entries of internal and external links with
+[`Regex::captures_iter`]. Using [`Cow`] avoids excessive [`String`] allocations.
+
+MediaWiki link syntax is described [here][MediaWiki link syntax].
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+#[macro_use]
+extern crate lazy_static;
+extern crate reqwest;
+extern crate regex;
+
+use std::io::Read;
+use std::collections::HashSet;
+use std::borrow::Cow;
+use regex::Regex;
+
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#         Reqwest(reqwest::Error);
+#         Regex(regex::Error);
+#     }
+# }
+# 
+fn extract_links(content: &str) -> Result<HashSet<Cow<str>>> {
+    lazy_static! {
+        static ref WIKI_REGEX: Regex =
+            Regex::new(r"(?x)
+                \[\[(?P<internal>[^\[\]|]*)[^\[\]]*\]\]    # internal links
+                |
+                (url=|URL\||\[)(?P<external>http.*?)[ \|}] # external links
+            ").unwrap();
+    }
+
+    let links: HashSet<_> = WIKI_REGEX
+        .captures_iter(content)
+        .map(|c| match (c.name("internal"), c.name("external")) {
+            (Some(val), None) => Cow::from(val.as_str().to_lowercase()),
+            (None, Some(val)) => Cow::from(val.as_str()),
+            _ => unreachable!(),
+        })
+        .collect();
+
+    Ok(links)
+}
+
+fn run() -> Result<()> {
+    let mut content = String::new();
+    reqwest::get(
+        "https://en.wikipedia.org/w/index.php?title=Rust_(programming_language)&action=raw",
+    )?
+        .read_to_string(&mut content)?;
+
+    println!("{:#?}", extract_links(&content)?);
+
+    Ok(())
+}
+#
+# quick_main!(run);
+```
+
 <!-- Categories -->
 
 [cat-encoding-badge]: https://badge-cache.kominick.com/badge/encoding--x.svg?style=social
@@ -889,6 +960,8 @@ fn run() -> Result<()> {
 
 [hyper-badge]: https://badge-cache.kominick.com/crates/v/hyper.svg?label=hyper
 [hyper]: https://docs.rs/hyper/
+[regex]: https://docs.rs/regex/
+[regex-badge]: https://badge-cache.kominick.com/crates/v/regex.svg?label=regex
 [reqwest-badge]: https://badge-cache.kominick.com/crates/v/reqwest.svg?label=reqwest
 [reqwest]: https://docs.rs/reqwest/
 [select]: https://docs.rs/select/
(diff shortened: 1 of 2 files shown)


src/net.md
```diff
@@ -876,6 +877,85 @@ fn run() -> Result<()> {
...
+## Extract all unique links from a MediaWiki markup
+
+[![reqwest-badge]][reqwest] [![regex-badge]][regex] [![cat-net-badge]][cat-net]
+
+Pull the source of a Mediawiki page using [`reqwest::get`] and then
+find all entries of internal and external links in that source with
+[`Regex::captures_iter`].
+
+In Mediawiki syntax:
+* Internal links take a form of `[[Page]]` or `[[Page|link text]]`.
```
budziq: I would try to minimize the info regarding the actual mediawiki syntax (ideally removing them all together).
I would just add external link to wikipedia site describing the link syntax.

src/net.md
```diff
@@ -876,6 +877,85 @@ fn run() -> Result<()> {
...
+#     }
+# }
+# 
+fn run() -> Result<()> {
+    let mut resp = reqwest::get(
+        "https://en.wikipedia.org/w/index.php?title=Rust_(programming_language)&action=raw",
+    )?;
+
+    let mut content = String::new();
+    resp.read_to_string(&mut content)?;
```
budziq: most likely you can get rid of resp variable by creating the request and reading from it in one go

src/net.md
```diff
@@ -876,6 +877,85 @@ fn run() -> Result<()> {
...
+
+    let mut links = Vec::new();
+    let mut urls = Vec::new();
+
+    // Capture the part of [[Page|link text]] that is inside the brackets
+    for mat in Regex::new(r#"\[\[(?P<link>.+?\|?)\]\]"#)?
+        .captures_iter(content.as_str())
+    {
+        if let Some(m) = mat["link"].split('|').next() {
+            links.push(m.to_string());
```
budziq: using `map` then `collect` into a set would be preferable

src/net.md
```diff
@@ -876,6 +877,85 @@ fn run() -> Result<()> {
...
+        }
+    }
+
+    for mat in Regex::new(r#"(url=|URL\||\[)(?P<url>http.*?)[ \|}]"#)?
+        .captures_iter(content.as_str())
+    {
+        urls.push(mat["url"].to_string());
+    }
+
+    links.sort_by_key(|s| s.to_lowercase());
```
budziq: using set would be preferable instead of sorting and deduping

src/net.md
```diff
@@ -876,6 +877,85 @@ fn run() -> Result<()> {
...
+    )?;
+
+    let mut content = String::new();
+    resp.read_to_string(&mut content)?;
+
+    let mut links = Vec::new();
+    let mut urls = Vec::new();
+
+    // Capture the part of [[Page|link text]] that is inside the brackets
+    for mat in Regex::new(r#"\[\[(?P<link>.+?\|?)\]\]"#)?
```
budziq: single regex with named capture groups would suffice. collecting results into a single HashSet

yandexx: That was the plan but I've tried <pre>\[\[(.+?)(\|?.*?)\]\]</pre> and the first group is too lazy and captures only the first character. I'll research further I guess.

budziq: most likely you'll have to glue the two regexes into one with |

yandexx: But then I can't use a single named group, right?

yandexx: Never mind I think I found a solution that works :)

src/net.md
```diff
@@ -876,6 +877,61 @@ fn run() -> Result<()> {
...
+# }
+# 
+fn run() -> Result<()> {
+    let mut content = String::new();
+    reqwest::get(
+        "https://en.wikipedia.org/w/index.php?title=Rust_(programming_language)&action=raw",
+    )?
+        .read_to_string(&mut content)?;
+
+    let mut links = Regex::new(r#"\[\[(?P<link>[^\[\]|]*)[^\[\]]*\]\]"#)?
```
budziq: i would suggest to have a single regex concatenated with | 
to make the regex readable please use the multiline (whitespace insensitive) regex definition like in https://brson.github.io/rust-cookbook/basics.html#run-an-external-command-and-process-stdout (you can split the regex into smaller parts each with comment starting with #)

having different capture groups would actually be beneficial as you will be able to match on them separately and implement the case insensitive logic (to_lower) in one arm.

src/net.md
```diff
@@ -876,6 +877,61 @@ fn run() -> Result<()> {
...
+        .captures_iter(content.as_str())
+        .map(|c| c["link"].to_string())
+        .collect::<HashSet<_>>();
+    links.extend(
+        Regex::new(r#"(url=|URL\||\[)(?P<link>http.*?)[ \|}]"#)?
+            .captures_iter(content.as_str())
+            .map(|c| c["link"].to_string()),
+    );
+
+    println!("{:?}", links);
```
budziq: i would iterate over the links or use the pretty print syntax

src/net.md
```diff
@@ -876,6 +877,61 @@ fn run() -> Result<()> {
...
+    let mut content = String::new();
+    reqwest::get(
+        "https://en.wikipedia.org/w/index.php?title=Rust_(programming_language)&action=raw",
+    )?
+        .read_to_string(&mut content)?;
+
+    let mut links = Regex::new(r#"\[\[(?P<link>[^\[\]|]*)[^\[\]]*\]\]"#)?
+        .captures_iter(content.as_str())
+        .map(|c| c["link"].to_string())
+        .collect::<HashSet<_>>();
```
budziq: annotating the variable type or the function return type is slightly cleaner.

src/net.md
```diff
@@ -876,6 +877,61 @@ fn run() -> Result<()> {
...
+#         Io(std::io::Error);
+#         Reqwest(reqwest::Error);
+#         Regex(regex::Error);
+#     }
+# }
+# 
+fn run() -> Result<()> {
+    let mut content = String::new();
+    reqwest::get(
+        "https://en.wikipedia.org/w/index.php?title=Rust_(programming_language)&action=raw",
```
yandexx: `lazy_static!` doesn't work with the ? operator so I'll have to unwrap? Let me know if there's an alternative.

src/net.md
yandexx: Doesn't seem to be possible to use multiline since I use raw string literals.

yandexx: Also this is what I have so far and I really don't like the if/else there. My iterator-fu may be lacking :/

```rust
    lazy_static! {
        static ref WIKI_REGEX: Regex = 
            Regex::new(r#"\[\[(?P<internal>[^\[\]|]*)[^\[\]]*\]\]|(url=|URL\||\[)(?P<external>http.*?)[ \|}]"#)
            .unwrap();
    }

    let links: HashSet<String> = WIKI_REGEX
        .captures_iter(content.as_str())
        .map(|c| {
            if c.name("internal").is_some() {
                c["internal"].to_lowercase()
            } else {
                // c.name("external").is_some()
                c["external"].to_string()
            }
        })
        .collect();
```

budziq: > Doesn't seem to be possible to use multiline since I use raw string literals.

please look at the example from docs
```rust
let re = Regex::new(r"(?x)
  (?P<y>\d{4}) # the year
  -
  (?P<m>\d{2}) # the month
  -
  (?P<d>\d{2}) # the day
").unwrap();
```
it also uses a raw literal. actually the `(?x)` is critical here.

in regard to the map you have posted I would use a match instead returning Cow<str>

yandexx: Thanks, `(?x)` has helped!

> in regard to the map you have posted I would use a match instead returning Cow

Hmm but what would I match on? What to return in case of no match? which shouldn't happen but this case still has to return something?

budziq: sorry I meant `match` returning `Cow<str>` (damn autocorrect on mobile)
you would write something along the lines of:

```rust
match (c["internal"], c["external"]) {
        (Some(val), None) => Cow::from(val.to_lowercase()),
        (None, Some(val)) => Cow::from(val),
        _ => unreachable!(),
    }
```
if tje unreachable would be triggered then there is a fundamental error with the regex definition so we want to fail noisily and fix our code and not try to recocer

src/net.md
budziq: unwrapping the regex compilation is ok this is an invariant of your code (regex is well written) and not something should try to recover from

src/net.md
yandexx: Ooh that's clever! Thanks, I'm learning a lot here!

brson:
I rebased and merged. Thanks @yandexx

PR #255, opened 2017-07-22
Basics: improve random number examples

dhardy:
I noticed the random examples were quite short and didn't document range for floating point cases. Improved.

Caveat: this doesn't explain why the `.ind_sample` method is used. I don't understand myself why the `Sample` trait (taking mutable reference to the range) exists, and if it does have a use-case it should probably have been named `MutSample`. But I digress.

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -4,7 +4,7 @@
 |--------|--------|------------|
 | [Read lines of strings from a file][ex-std-read-lines] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Read and write integers in little-endian byte order][ex-byteorder-le] | [![byteorder-badge]][byteorder] | [![cat-encoding-badge]][cat-encoding] |
-| [Generate random floating point numbers][ex-rand-float] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Generate random numbers for simple types][ex-rand-simple] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with normal distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
@@ -111,9 +111,11 @@ fn decode(mut bytes: &[u8]) -> Result<Payload> {
 # quick_main!(run);
 ```
 
-[ex-rand-float]: #ex-rand-float
-<a name="ex-rand-float"></a>
-## Generate random floating point numbers
+[ex-rand-simple]: #ex-rand-simple
+<a name="ex-rand-simple"></a>
+## Generate random numbers for simple types
+
+The `rand` crate provides a convenient source of psuedo-random numbers.
 
 [![rand-badge]][rand] [![cat-science-badge]][cat-science]
 
@@ -122,9 +124,28 @@ extern crate rand;
 use rand::Rng;
 
 fn main() {
+    // Each thread has an automatically-initialised random number generator:
     let mut rng = rand::thread_rng();
-    println!("Random f64: {}", rng.gen::<f64>());
+    
+    // Integers are uniformly distributed over the type's whole range:
+    let n1: u32 = rng.gen();
+    let n2: u16 = rng.gen();
+    let n3: u8 = rng.gen();
+    println!("Random u32, u16, u8: {}, {}, {}", n1, n2, n3);
+    println!("Random i32: {}", rng.gen::<i32>());
+    
+    // Floating point numbers are uniformly distributed in the half-open range [0, 1)
+    println!("Random float: {}", rng.gen::<f64>());
+    println!("Random floats: {:?}", (0..5).map(|_| rng.gen()).collect::<Vec<f32>>());
 }
+```
+
+Example output:
+```
+Random u32, u16, u8: 1224630198, 31187, 43
+Random i32: -515338046
+Random float: 0.7258180295601702
+Random floats: [0.6633928, 0.4400041, 0.49633145, 0.02997303, 0.41612816]
 ```
 
 [ex-rand-range]: #ex-rand-range
@@ -142,9 +163,34 @@ use rand::Rng;
 
 fn main() {
     let mut rng = rand::thread_rng();
-    println!("{}", rng.gen_range(0, 10));
+    println!("Integer: {}", rng.gen_range(0, 10));
+    println!("Float: {}", rng.gen_range(0.0, 10.0));
 }
 ```
+
+Alternatively, one can use the [`Range`](https://doc.rust-lang.org/rand/rand/distributions/range/struct.Range.html) distribution.
+This has the same effect, but may be faster when repeatedly generating numbers
+in the same range.
+
+```rust
+extern crate rand;
+
+use rand::distributions::{Range, IndependentSample};
+
+fn main() {
+    let mut rng = rand::thread_rng();
+    let die = Range::new(1, 7);
+    
+    loop {
+        let throw = die.ind_sample(&mut rng);
+        println!("Roll the die: {}", throw);
+        if throw == 6 {
+            break;
+        }
+    }
+}
+```
+
 
 [ex-rand-dist]: #ex-rand-dist
 <a name="ex-rand-dist"></a>
--- a/src/basics.md
+++ b/src/basics.md
@@ -4,9 +4,9 @@
 |--------|--------|------------|
 | [Read lines of strings from a file][ex-std-read-lines] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Read and write integers in little-endian byte order][ex-byteorder-le] | [![byteorder-badge]][byteorder] | [![cat-encoding-badge]][cat-encoding] |
-| [Generate random numbers for simple types][ex-rand-simple] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Generate random numbers][ex-rand] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
-| [Generate random numbers with normal distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Generate random distributions][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Run an external command and process stdout][ex-parse-subprocess-output] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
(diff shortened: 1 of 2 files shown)


src/basics.md
```diff
@@ -122,11 +124,30 @@ extern crate rand;
...
 fn main() {
+    // Each thread has an automatically-initialised random number generator:
     let mut rng = rand::thread_rng();
-    println!("Random f64: {}", rng.gen::<f64>());
+    
+    // Integers are uniformly distributed over the type's whole range:
+    let n1: u32 = rng.gen();
+    let n2: u16 = rng.gen();
+    let n3: u8 = rng.gen();
+    println!("Random u32, u16, u8: {}, {}, {}", n1, n2, n3);
```
budziq: The output formatting here is a little off. I would suggest pairing output with types.

src/basics.md
```diff
@@ -111,9 +111,11 @@ fn decode(mut bytes: &[u8]) -> Result<Payload> {
...
 ```
 
-[ex-rand-float]: #ex-rand-float
-<a name="ex-rand-float"></a>
-## Generate random floating point numbers
+[ex-rand-simple]: #ex-rand-simple
+<a name="ex-rand-simple"></a>
+## Generate random numbers for simple types
+
+The `rand` crate provides a convenient source of psuedo-random numbers.
```
dhardy: Actually, they do have a link: the badge below links. Never mind, no real reason not to include another link.

src/basics.md
```diff
@@ -122,11 +124,30 @@ extern crate rand;
...
+    println!("Random u32, u16, u8: {}, {}, {}", n1, n2, n3);
+    println!("Random i32: {}", rng.gen::<i32>());
+    
+    // Floating point numbers are uniformly distributed in the half-open range [0, 1)
+    println!("Random float: {}", rng.gen::<f64>());
+    println!("Random floats: {:?}", (0..5).map(|_| rng.gen()).collect::<Vec<f32>>());
 }
 ```
 
+Example output:
```
dhardy: What would be really useful is a **run** button below each example. Is this planned?

src/basics.md
```diff
@@ -142,10 +163,35 @@ use rand::Rng;
 
 fn main() {
     let mut rng = rand::thread_rng();
-    println!("{}", rng.gen_range(0, 10));
+    println!("Integer: {}", rng.gen_range(0, 10));
+    println!("Float: {}", rng.gen_range(0.0, 10.0));
 }
 ```
 
+Alternatively, one can use the [`Range`](https://doc.rust-lang.org/rand/rand/distributions/range/struct.Range.html) distribution.
```
dhardy: Disagree. If someone wants to generate numbers within a range, they'll follow that link, and it's fair to show them the two ways of doing it there. If they don't understand the second example, no real problem since they already have an alternative.

On the other hand, understanding how to do that given just the normal distribution example may not be so obvious.

src/basics.md
budziq: >  Is this planned?

Sorry I wanted to mention it as the main reason but it slipped my mind. Yep the run button is already implemented but disabled due to rust playpen historically not supporting crates. Currently we are coordinating list of supported crates with playpen guys https://github.com/brson/rust-cookbook/issues/87

src/basics.md
budziq: you are essentially duplicating an existing example. this essentially ads only noise

src/basics.md
budziq: I originally wanted to suggest removing the  whole
>"The `rand` crate provides a convenient source of psuedo-random numbers"

as all of that information is either already available (as you have noted) or obvious

src/basics.md
budziq: on the other hand "simulating a six sided die roll" might be a problem that merits its own example so we might decide to either extract it as a separate one or rewrite one of the previously mentioned ones 👍

dhardy:
@budziq updated, including changing the "normal distributions" section to be a "general distributions" section and reference the previous usage.

src/intro.md
```diff
@@ -22,9 +22,9 @@ community. It needs and welcomes help. For details see
 |--------|--------|------------|
 | [Read lines of strings from a file][ex-std-read-lines] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Read and write integers in little-endian byte order][ex-byteorder-le] | [![byteorder-badge]][byteorder] | [![cat-encoding-badge]][cat-encoding] |
-| [Generate random floating point numbers][ex-rand-float] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Generate random numbers][ex-rand] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
-| [Generate random numbers with normal distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Generate random number distributions][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
```
budziq: "Generate random number distributions"  is slightly inaccurate. We are not generating distributions but rather random number with/of/described by given distribution. I would suggest to rewrite the title as:
"Generate random number with given distribution"

src/basics.md
```diff
@@ -111,11 +111,11 @@ fn decode(mut bytes: &[u8]) -> Result<Payload> {
...
 
-[ex-rand-simple]: #ex-rand-simple
-<a name="ex-rand-simple"></a>
-## Generate random numbers for simple types
+[ex-rand]: #ex-rand
+<a name="ex-rand"></a>
+## Generate random numbers
 
-The `rand` crate provides a convenient source of psuedo-random numbers.
+The [`rand`] crate provides a convenient source of psuedo-random numbers.
```
budziq: please remove the backticks. it is a normal link and not linked code

src/basics.md
```diff
@@ -194,23 +184,30 @@ fn main() {
...
 <a name="ex-rand-dist"></a>
-## Generate random numbers with normal distribution
+## Random number distributions
 
 [![rand-badge]][rand] [![cat-science-badge]][cat-science]
 
-Creates a [`Normal`] distribution with mean `3` and standard deviation `5`
-and generates a random value with [`IndependentSample::ind_sample`].
+You've already seen how to create numbers with uniform distribution [above][ex-rand-range]
+(using [`Range`]). Other distributions are used in the same way: create a distribution, then
```
budziq: most of this description is obvious from the trivial code below no need to be verbose here.

src/basics.md
```diff
@@ -194,23 +184,30 @@ fn main() {
...
 [ex-rand-dist]: #ex-rand-dist
 <a name="ex-rand-dist"></a>
-## Generate random numbers with normal distribution
+## Random number distributions
 
 [![rand-badge]][rand] [![cat-science-badge]][cat-science]
 
-Creates a [`Normal`] distribution with mean `3` and standard deviation `5`
-and generates a random value with [`IndependentSample::ind_sample`].
+You've already seen how to create numbers with uniform distribution [above][ex-rand-range]
```
budziq: we don't aim to make the cookbook into a tutorial (what I've meant by "progression" was if there is any missing information between examples or if these are not sorted by their difficulty). We strive for problem solution (hence cookbook) approach of self-contained examples. 

I would roll back the original description here as it was short and to the point and just link to alternative distributions.

src/basics.md
```diff
@@ -194,23 +184,30 @@ fn main() {
...
 extern crate rand;
 
 use rand::distributions::{Normal, IndependentSample};
 
 fn main() {
-    let normal = Normal::new(3.0, 5.0);
     let mut rng = rand::thread_rng();
+    
+    // mean 2, standard deviation 3:
+    let normal = Normal::new(2.0, 3.0);
```
budziq: any reason to change the order of statements or distribution parameters?

src/basics.md
```diff
@@ -566,8 +563,11 @@ fn main() {
 [`Read`]: https://doc.rust-lang.org/std/io/trait.Read.html
 [`Normal`]: https://doc.rust-lang.org/rand/rand/distributions/normal/struct.Normal.html
 [`IndependentSample::ind_sample`]: https://doc.rust-lang.org/rand/rand/distributions/trait.IndependentSample.html#tymethod.ind_sample
+[`rand`]: https://doc.rust-lang.org/rand
```
budziq: there is already a link to rand crate with proper form.
```
[normal link]
`code`
[`linked code`]
```

src/basics.md
```diff
@@ -168,7 +158,7 @@ fn main() {
 }
 ```
 
-Alternatively, one can use the [`Range`](https://doc.rust-lang.org/rand/rand/distributions/range/struct.Range.html) distribution.
+Alternatively, one can use the [`Range`] distribution (otherwise known as uniform).
```
budziq: 👍  I would linkify the "uniform" into some autharitive resource

src/basics.md
```diff
@@ -128,26 +128,16 @@ fn main() {
     let mut rng = rand::thread_rng();
     
     // Integers are uniformly distributed over the type's whole range:
-    let n1: u32 = rng.gen();
```
budziq: no reason to leave out the `u32` it was actually neat.

src/basics.md
```diff
@@ -128,26 +128,16 @@ fn main() {
...
+    let n1: u8 = rng.gen();
     let n2: u16 = rng.gen();
-    let n3: u8 = rng.gen();
-    println!("Random u32, u16, u8: {}, {}, {}", n1, n2, n3);
+    println!("Random u8: {}; random u16: {}", n1, n2);
     println!("Random i32: {}", rng.gen::<i32>());
```
budziq: I would suggest for the formatting to be uniform for all types. each on separate line

dhardy: It's deliberately not uniform. It's supposed to show readers different ways of doing things, so they can use whichever they like.

src/basics.md
dhardy: I would rename the thing if I could. I may actually try, but it's not high up the priorities of what that crate needs.

dhardy:
You know, if you can put this much effort into constructive criticism, I might as well let you write it yourself. There's only so much time I'm prepared to waste on a cookbook I'm sure most people won't give more than a quick glance anyway.

src/basics.md
budziq: I meant making "uniform" into a link pointing to something like:

https://en.m.wikipedia.org/wiki/Uniform_distribution_(continuous)

src/basics.md
budziq: here I meant _output formatting_ I perfectly understand why you are obtaining the values in various ways. 

Unless you mean to teach the users that they can print two variables on the same line without a newline in between.

budziq:
> You know, if you can put this much effort into constructive criticism, I might as well let you write it yourself. 

Well it certainly takes a lot of effort and motivation considering I am juggling +10PR's while on vacation with only mobile at my disposal.

Each an every one is held under the same scrutiny in order to have consistent format and quality. Please note that while the presentation and form might not look like it the cookbook is aimed to become an official resource once its out of its pre-alpha stage.

I agree that contribution is painful at the moment (I guess ) 

> There's only so much time I'm prepared to waste on a cookbook 

Well I would not call it "wasting" you are helping a lot. For instance I would certainly fix the examples myself if I had even a slightest notion that there is a problem in them. 

We are thinking about the final form and format for cookbook and if anything, your problems and suggestions will have an impact on the final output.

I guess that this PR might go smother If we had discussed your suggestions prior to you making the actual work. But unfortunately I did lack the means to drive the conversation.

Sorry to have discouraged you :(

brson:
This looks great, @dhardy. Thanks. I rebased and merged it. I did change the language in the non-uniform distributions example to not reference previous examples. For the moment at least it is true that the examples are pretty independent, and it's pretty likely that the format of the book will change drastically before any kind of final release.

I do think it would be worth having a better narrative progression between examples so that it reads nicely as a book, leads the reader on a journey of some kind. This is great for now though - we're going to have to circle back for a big editing pass in the future.

Thanks for reviewing @budziq (but do feel free to take some time off while you are traveling).

PR #274, opened 2017-08-31
Use a threadpool to calculate SHA1 in all *.iso files in a folder.

ericho:


src/concurrency.md
```diff
@@ -232,6 +233,49 @@ fn run() -> Result<()> {
...
+use std::fs::File;
+use std::io::BufReader;
+use std::io::prelude::*;
+use threadpool::ThreadPool;
+
+fn main() {
+    let cpus = num_cpus::get();
+    let pool = ThreadPool::new(cpus);
+
+    for entry in WalkDir::new(".")
```
budziq: lets go wild and compute the sha1 of all *.iso files in given directory or all files listed in 
or the ones listed in 
https://cloud-images.ubuntu.com/trusty/current/SHA1SUMS
https://cloud-images.ubuntu.com/trusty/current/

In this way we could have an actual useful example. something along these lines - "Verify in parallel downloaded  files consistency with SHA1 algorithm"

ericho:
Thank you @budziq for the feedback, I'll work on the fixes 😄

src/concurrency.md
```diff
@@ -1,11 +1,12 @@
 # Concurrency
 
-| Recipe | Crates | Categories |
```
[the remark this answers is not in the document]
ericho: It is fixed.. I mess with something but I'm not sure what was :smile:

src/concurrency.md
```diff
@@ -232,6 +233,49 @@ fn run() -> Result<()> {
...
+<a name="threadpool-walk"></a>
+## Use a threadpool to calculate MD5 to all files in a folder
+
+[![threadpool-badge]][threadpool] [![num_cpus-badge]][num_cpus] [![walkdir-badge]][walkdir] [![cat-concurrency-badge]][cat-concurrency][![cat-filesystem-badge]][cat-filesystem]
+
+This example uses `threadpool`, `walkdir` and `md5` crates to calculate the MD5 for every file present in the current directory. A threadpool `pool` is created using the number of cpus present in the system. Then every `entry` returned by `walkdir` is passed into this pool to perform the operations of reading and compute of md5. At the end the program waits for all jobs to finish.
+
+```rust,no_run
+extern crate walkdir;
+extern crate md5;
```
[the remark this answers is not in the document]
ericho: I changed the example to use `ring` and SHA1

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+            let tx = tx.clone();
+            pool.execute(move || {
+                let file = match File::open(entry.path()) {
+                    Ok(file) => file,
+                    Err(err) => {
+                        tx.send(Err(err)).expect("Could not send data!");
+                        return;
+                    },
+                };
+                let mut buf_reader = BufReader::new(file);
```
budziq: Reading the whole file into memory is not really the best way to go. Ring provides a nice incremental api.
for its usage please see the 
[calculate the sha-256 digest of a file](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#calculate-the-sha-256-digest-of-a-file) example

Moreover it would be much nicer to extract the file reading and sha1 calculation into a separate function taking a `AsRef<Path>` and returning a `Result`. Then you would have only one `tx.send()` without the verbose matches all over the place.

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+                        let digest = digest::digest(&digest::SHA1, &content);
+                        let file_path = String::from(entry.path().to_string_lossy());
+                        // Send as a Result
+                        tx.send(Ok((digest, file_path))).expect("Could not send data!");
+                    }
+                    Err(e) => tx.send(Err(e)).expect("Could not send data!"),
+                }
+            });
+        }
+
```
budziq: no need to use `take` as long as you `drop(tx)` prior to this loop.

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#     }
+# }
+
+use walkdir::{WalkDir, DirEntry};
+use std::fs::File;
+use std::io::BufReader;
+use std::io::prelude::*;
```
budziq: no need to bring the whole prelude here. please import only the required traits.

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+    let mut file_counter: usize = 0;
+
+    let (tx, rx) = channel();
+
+    // Look in the current directory.
+    for entry in WalkDir::new(".")
+        .follow_links(true)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| !e.file_type().is_dir() && is_iso(e)) {
```
budziq: - I would operate on `Path`or `PathBuf` instead of `DirEntry` here It is lighter, more idiomatic and has all the required interfaces to boot :).
- I would split this closure to `filter` (only the condition) and `map` (the actual body). It would convey the intent better.

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+use std::fs::File;
+use std::io::BufReader;
+use std::io::prelude::*;
+use threadpool::ThreadPool;
+use std::sync::mpsc::channel;
+use ring::digest;
+
+# // Verify the iso extension
+# fn is_iso(entry: &DirEntry) -> bool {
+#     entry.file_name()
```
budziq: If you use `Path` or `PathBuf`  (or better `AsRef<Path>`) instead of `DirEntry` you'll be able to 
use use `Path::extension()` and compare directly to "iso"

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+
+    // Look in the current directory.
+    for entry in WalkDir::new(".")
+        .follow_links(true)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| !e.file_type().is_dir() && is_iso(e)) {
+            // Need to know how many entry were sent to the pool.
+            file_counter = file_counter + 1;
+            let entry = entry.clone();
```
budziq: this will not be needed if you use `PathBuf` instead as the owned value will just get moved into the closure

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+
+fn run() -> Result<()> {
+    let cpus = num_cpus::get();
+    let pool = ThreadPool::new(cpus);
+    let mut file_counter: usize = 0;
+
+    let (tx, rx) = channel();
+
+    // Look in the current directory.
+    for entry in WalkDir::new(".")
```
budziq: I would parametrize the call with a variable event if its value is hardcoded earlier to something like "/home/user/downloads".

budziq: Hi @ericho do you need any help in finalizing this example?

ericho: Hi, sorry, I was busy these last days. I think I solved all the comments. 

BTW, thanks for the review, I learned a lot doing this example :smile:

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+# }
+
+use walkdir::WalkDir;
+use std::fs::File;
+use std::io::BufReader;
+use std::io::prelude::Read;
+use std::path::Path;
+use threadpool::ThreadPool;
+use std::sync::mpsc::channel;
+use ring::digest::{Context, Digest, SHA256};
```
budziq: we probably wanted `SHA1` instead of `SHA256`

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#     }
+# }
+
+use walkdir::WalkDir;
+use std::fs::File;
+use std::io::BufReader;
+use std::io::prelude::Read;
```
budziq: nearly everything in `std::io::prelude` is reexported to `std::io`
so we can write it as
`use std::io::{BufReader, Read}`

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+extern crate ring;
+extern crate num_cpus;
+extern crate threadpool;
+
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#     }
+# }
+
```
budziq: please add additional `#` here so we avoid 2 blank spaces

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+        }
+
+    drop(tx);
+    for t in rx.iter() {
+        let (sha, path) = t?;
+        println!("{:?} {}", sha, path);
+    }
+    Ok(())
+}
+
```
budziq: please remove the single blank space

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+use std::path::Path;
+use threadpool::ThreadPool;
+use std::sync::mpsc::channel;
+use ring::digest::{Context, Digest, SHA256};
+
+// Verify the iso extension
+# fn is_iso(entry: &Path) -> bool {
+#     match entry.extension() {
+#         Some(e) if e == "iso"=> true,
+#         Some(_) => false,
```
budziq: this match can be shortened even further into just two branches
```rust
    match entry.extension() {
        Some(e) if e.to_string_lossy().to_lowercase() == "iso" => true,
        _ => false,
    }
```

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[threadpool-walk]: #threadpool-walk
+<a name="threadpool-walk"></a>
+
+## Calculate SHA1 sum of *.iso files concurrently
+
+[![threadpool-badge]][threadpool] [![num_cpus-badge]][num_cpus] [![walkdir-badge]][walkdir] [![cat-concurrency-badge]][cat-concurrency][![cat-filesystem-badge]][cat-filesystem]
```
budziq: i would add `ring` crate badge here

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+// Verify the iso extension
+# fn is_iso(entry: &Path) -> bool {
+#     match entry.extension() {
+#         Some(e) if e == "iso"=> true,
+#         Some(_) => false,
+#         None => false,
+#     }
+# }
+
+fn compute_digest<P: AsRef<Path>>(filepath: P) -> Result<(Digest, String)> {
```
budziq: if you return `Result<(Digest, P)>` you can completely drop the `filename` variable and just return `Ok((context.finish(), filepath))`

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+    let filename = String::from(filepath.as_ref().to_string_lossy());
+    let file = File::open(filepath)?;
+    let mut buf_reader = BufReader::new(file);
```
budziq: we don't need a separate variable for `file`. we can write it as `BufReader::new(File::open(filepath)?);` (the file will be moved into the `BufReader`anyway)

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
+
+    let (tx, rx) = channel();
+
+    // Look in the current directory.
+    for entry in WalkDir::new("/home/user/Downloads")
+        .follow_links(true)
+        .into_iter()
+        .filter_map(|e| e.ok())
+        .filter(|e| !e.path().is_dir() && is_iso(e.path())) {
+            let path = entry.path().to_path_buf();
```
budziq: instead of `to_path_buf` we can use `to_owned` here.

src/concurrency.md
```diff
@@ -232,6 +233,96 @@ fn run() -> Result<()> {
...
 ```
 
+[threadpool-walk]: #threadpool-walk
+<a name="threadpool-walk"></a>
+
+## Calculate SHA1 sum of *.iso files concurrently
+
+[![threadpool-badge]][threadpool] [![num_cpus-badge]][num_cpus] [![walkdir-badge]][walkdir] [![cat-concurrency-badge]][cat-concurrency][![cat-filesystem-badge]][cat-filesystem]
+
+This example uses `threadpool`, `walkdir` and `ring` crates to calculate the SHA1 for every file present in the current directory. A threadpool `pool` is created using the number of cpus present in the system. Then every `entry` returned by `walkdir` is passed into this pool to perform the operations of reading and compute SHA1. At the end the program waits for all jobs to finish. To get better results, compile this program in release mode.
```
budziq: 1. It would replace

> "This example uses `threadpool`, `walkdir` and `ring` crates to calculate ..."

with "Calculates ...". The reader already knows which crates are used form the heading.

2. Actually this example recursively calculates SHA256 of every iso file present in "/home/user/Downloads"

So please update the code or the description ;)

3. typo - "reading and compute" => "reading and computing" 

4. There are several identifiers mentioned `pool`, `entry` etc. Please link these to docs

budziq:
Thanks @ericho great work!

PR #275, opened 2017-09-02
Add rayon thumbnail generation example

nocduro:


src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+    let output_path = thumb_dir.join(file);
+
+    let ref mut fout = File::create(output_path.as_path())?;
+    img.save(fout, image::JPEG)?;
+
+    Ok(())
+}
+
+#[derive(Debug)]
+struct ImageProblem {
```
budziq: this whole struct and its uses can be removed

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+extern crate glob;
+extern crate image;
+extern crate rayon;
+
+use std::path::{Path, PathBuf};
+use std::fs;
+use std::fs::File;
+
+use glob::{glob_with, MatchOptions};
+use image::{FilterType, ImageResult};
```
budziq: instead of `ImageResult` we will want [ImageError](https://docs.rs/image/0.15.0/image/enum.ImageError.html)

so we can use [error_chain](https://github.com/rust-lang-nursery/error-chain) here. Please see our [note about error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+use rayon::prelude::*;
+
+fn main() {
```
budziq: we will want to be able to use `?` in our main so 
this function should be renamed to `fn run() -> Result<()>` and it will be called from `quick_main!(run);` as described in [a note about error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+[![rayon-badge]][rayon] [![glob-badge]][glob] [![image-badge]][image] [![cat-concurrency-badge]][cat-concurrency] [![cat-filesystem-badge]][cat-filesystem]
+
+This example uses the `glob`, `image`, and `rayon` crates to generate thumbnails for all .jpg files in the current directory. 
+The thumbnails are stored in a new folder called `thumbnails`.
+
+Files are found with [`glob::glob_with`] so we can match case insensitively on both `.jpg` and `.JPG`.
+Images are then resized in parallel with [`par_iter`] using [`DynamicImage::resize`].
+
+
+```rust
```
budziq: we will want this example to be `rust,no_run` to make it compile but not be run in our CI environment.

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+}
+
+/// Find all files that have a `.jpg` extension in the current directory
+fn find_jpg() -> Vec<PathBuf> {
+    let mut paths = Vec::new();
+
+    // use `glob_with()` and MatchOptions so we can match case insensitively to .jpg
+    let options = MatchOptions {
+        case_sensitive: false,
+        require_literal_separator: false,
```
budziq: you can use [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html) to fill in the options other than `case_sensitive`

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+    let files = find_jpg();
+    println!("Found {} files to convert", files.len());
+    
+    for problem in parallel_resize(&files, Path::new("thumbnails"), 300) {
+        println!("{:?} failed with error: {}", problem.file_path, problem.err);
+    }
+}
+
+/// Find all files that have a `.jpg` extension in the current directory
+fn find_jpg() -> Vec<PathBuf> {
```
budziq: I would remove this function and move it to the `run` body. When we finish with it it will be 3-4 liner :)

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+    let mut paths = Vec::new();
+
+    // use `glob_with()` and MatchOptions so we can match case insensitively to .jpg
+    let options = MatchOptions {
+        case_sensitive: false,
+        require_literal_separator: false,
+        require_literal_leading_dot: false,
+    };
+
+    for entry in glob_with("*.jpg", &options).expect("Failed to read glob pattern") {
```
budziq: we will want to use the `?` operator instead of `expect`
also instead of this whole for loop with match we can use a simple [`collect`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect) call and collect the iterator into a `Result<Vec<_>>` and then `?` away the result.

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+    }
+
+    paths
+}
+
+/// Resize `file` to have a maximum dimension of `longest_edge` and save the resized
+/// image to the `thumb_dir` folder
+fn resize_jpg(file: &Path, thumb_dir: &Path, longest_edge: u32) -> ImageResult<()> {
+    // create the folder if it doesn't already exist
+    fs::create_dir_all(thumb_dir)?;
```
budziq: please call the `create_dir_all` only once prior to multithreaded code. These have few nasty pitfals when used concurrently https://doc.rust-lang.org/std/fs/fn.create_dir_all.html#errors

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+            Err(e) => println!("Error with: {:?}", e),
+        }
+    }
+
+    paths
+}
+
+/// Resize `file` to have a maximum dimension of `longest_edge` and save the resized
+/// image to the `thumb_dir` folder
+fn resize_jpg(file: &Path, thumb_dir: &Path, longest_edge: u32) -> ImageResult<()> {
```
budziq: - how about calling the function `make_thumbnail` (now it reads a little like the original is lost)?
- Also , I would suggest for the parameter to be `AsRef<Path>` instead of `&Path` this will be   
- and once we switch to `error_chain` we will return `Result<()>` or even `Result<PathBuf>` (if we decide to return the final thumb path for printing in `run`) here
- I would probably rename the `file` variable to `original` or `orig`

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+    // create the folder if it doesn't already exist
+    fs::create_dir_all(thumb_dir)?;
+
+    let img = image::open(file)?;
+    let img = img.resize(longest_edge, longest_edge, FilterType::Nearest);
```
budziq: We can remove the additional let binding here and separate call to `img.save`.  This can be transformed to chained `image::open(file)?.resize(.....).save(....)?`. Rustfmt will most likely split these to separate lines but we will reduce number of variables.

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+
+    let output_path = thumb_dir.join(file);
+
+    let ref mut fout = File::create(output_path.as_path())?;
```
budziq: i would remove the `ref` and just use `&mut fout` in the `save` call. 

Also `output_path.as_path()` can be converted to `&output_path`

src/concurrency.md
```diff
@@ -70,7 +71,102 @@ fn main() {
...
+
+#[derive(Debug)]
+struct ImageProblem {
+    file_path: PathBuf,
+    err: image::ImageError,
+}
+
+/// Wrapper for `resize_jpg` to resize multiple images at the same time with rayon
+/// Returns a vector of the files that had an error during conversion, along with the actual error
+fn parallel_resize(files: &[PathBuf], thumb_dir: &Path, longest_edge: u32) -> Vec<ImageProblem> {
```
budziq: I would remove this function and substitute it with a call to
```rust
        files.par_iter()
        .map(|path| {
            resize_jpg(path, thumb_dir, longest_edge)
        })
        .collect();
```
collecting all into a `Result<Vec<_>>` or `Vec<Result<_>>` depending if we want to bail on single error or handle one by one.

src/concurrency.md
nocduro: > we will want to use the `?` operator instead of `expect`

Woops, just copied the code from [`glob`'s documentation](https://doc.rust-lang.org/glob/glob/index.html#examples) to get it running. I'll change it over to use `?` and `collect`

src/concurrency.md
nocduro: > Also , I would suggest for the parameter to be AsRef<Path> instead of &Path this will be

I've never used `AsRef<>` before, from a little reading it seems like it does some type coercion, so we can pass other types to be coerced into a `&Path`? So, stuff like `OsStr`, `PathBuf`, `Path` will all work?

src/concurrency.md
nocduro: Yeah, that was left over from some other experiments I was doing. I think I originally did it in case of going into to multiple folders, but never got that far, so it was never tested... 

I'll change it over so it only does it once

src/concurrency.md
nocduro: Sounds good, I mostly copied the image saving code from one of the examples in the [`image` documentation](https://github.com/PistonDevelopers/image#61-opening-and-saving-images)

src/concurrency.md
nocduro: This is one of the parts I had the most trouble with. The way I did it here was so that I could have some files fail, but still finish most of the files. This way the program would do all it could, and then print out the filenames with their specific errors before exiting.

I guess this is more of a design pattern thing than a `rayon` thing, but I think it's the most important part of the example (handling errors in parallel code), since the other examples don't really deal with things that can error.

I'm open to what people with more experience think is the right approach in this case, as this is some of the first parallel code I've written.

src/concurrency.md
budziq: Its not really type coercion but rather allows for the function to be generic over inputs that satisfy AsRef<Path> traitbound. yep these types and few others will be able to work here

budziq: And this particular trait exposes cheap conversion to &Path

src/concurrency.md
budziq: Actually there would be no parallel specific pattern here as rayon hides nearly all parallelism details. 

You would handle the errors exactly as if you would in normal iterator based interface. For instance by returning a `Result` with file path from each `map` invocation and then handle each `Result` one by one. You can look at some of the application dev examples like [decompress a tarball while removing a prefix from path example](https://rust-lang-nursery.github.io/rust-cookbook/app.html#decompress-a-tarball-while-removing-a-prefix-from-the-paths).

If you'll have any problems or doubts, do not hesitate to ping me :)

budziq: @nocduro Do you need any help with finishing this example?

nocduro:
Sorry for taking a while on this, my semester recently started up again and I've been a bit busy with course work.

I did most of the changes, but the error handling part in `par_iter()` is really ugly in my version, so any help with that would be greatly appreciated.

Right now when I run the program with one corrupt .jpg in the directory I get this output:

```
Saving 197 thumbnails into "thumbnails"
"broken.jpg" failed: Format error: first two bytes is not a SOI marker
Done
```

Also, when I updated my fork with `git pull upstream master` and now `cargo test` fails on some of the examples, so I'm not sure what I did there...

```
failures:
    encoding_sect_deserialize_a_toml_configuration_file_line_132
    encoding_sect_handle_invalid_csv_data_with_serde_line_456
    encoding_sect_read_csv_records_with_different_delimeter_line_507
    encoding_sect_serialize_records_to_csv_using_serde_line_403
    net_sect_consume_a_paginated_restful_api_line_632
    net_sect_create_and_delete_gist_with_github_api_line_552
    net_sect_query_the_github_api_line_377
    net_sect_set_custom_headers_and_url_parameters_for_a_rest_request_line_477

test result: FAILED. 68 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test skeptic'
```

nocduro:
Also, I couldn't get the `AsRef<Path>` thing to work because when I tried to call `.join()` for the output_path I would get an error that I didn't know how to fix. Something to do with AsRef<Path> not implementing the `.join()` method or something.

budziq:
> Sorry for taking a while on this, my semester recently started up again and I've been a bit busy with course work.

no reason to be sorry :)

> Also, when I updated my fork with git pull upstream master and now cargo test fails on some of the examples, so I'm not sure what I did there...

Most likely its a skeptic but. `cargo clean` will fix the problem.

>  error handling part in par_iter() is really ugly in my version,

we will fix it in the course of the review :)

> Also, I couldn't get the AsRef<Path> thing to work because when I tried to call .join()

To use the AsRef<Path> you would have to

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+    let files: Vec<_> = glob_with("*.jpg", &options)?
+        .filter_map(|x| x.ok())
+        .collect();
+
+    let thumb_dir = Path::new("thumbnails");
+    create_dir_all(thumb_dir)?;
+
+    println!("Saving {} thumbnails into {:?}", files.len(), thumb_dir);
+
+    let image_failures: Vec<String> = files
```
budziq: no need to fully annotate the type `Vec<_>` will do

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+    // use the default MatchOptions so the search is case insensitive
+    let options: MatchOptions = Default::default();
+    let files: Vec<_> = glob_with("*.jpg", &options)?
+        .filter_map(|x| x.ok())
+        .collect();
+
+    let thumb_dir = Path::new("thumbnails");
+    create_dir_all(thumb_dir)?;
+
+    println!("Saving {} thumbnails into {:?}", files.len(), thumb_dir);
```
budziq: instead of `{:?}` you can just call `thumb_dir.display()` and use standard formatting braces

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+
+fn run() -> Result<()> {
+    // find all files in current directory that have a .jpg extension
+    // use the default MatchOptions so the search is case insensitive
+    let options: MatchOptions = Default::default();
+    let files: Vec<_> = glob_with("*.jpg", &options)?
+        .filter_map(|x| x.ok())
+        .collect();
+
+    let thumb_dir = Path::new("thumbnails");
```
budziq: if you use `AsRef<Path>` in `make_thumbnail` you will be able to use just  `let thumb_dir = "thumbnails"`
as `create_dir_all` also thakes generic `AsRef<Path>` param

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+    create_dir_all(thumb_dir)?;
+
+    println!("Saving {} thumbnails into {:?}", files.len(), thumb_dir);
+
+    let image_failures: Vec<String> = files
+        .par_iter()
+        .map(|path| -> std::result::Result<(), String> {
```
budziq: the return type annotation will not be needed once we fix the error handling blow

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+            match make_thumbnail(path, thumb_dir, 300) {
+                Ok(_) => Ok(()),
```
budziq: such match is realized by `Result::map_err` and the `format!` can be replaced with `error_chain(| | path.display().to_string())`

so we would get something like
```rust
make_thumbnail(path, thumb_dir, 300).map_err(|e| {
                e.chain_err(|| path.display().to_string())
            })
```

of course this is strictly due to the fact that you would like to push another dedicated error with path of the failed element (there are also other ways using error_chain requiring more setup in the `error_chain` section but less code in the actual `run`).

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+    Ok(())
+}
+
+/// Resize `original` to have a maximum dimension of `longest_edge` and save the
+/// resized image to the `thumb_dir` folder
+fn make_thumbnail(original: &Path, thumb_dir: &Path, longest_edge: u32) -> Result<()> {
+    let img = image::open(original)?;
+    let fout = &mut File::create(thumb_dir.join(original))?;
+
+    img.resize(longest_edge, longest_edge, FilterType::Nearest)
```
budziq: you can replace the two last lines with

```rust
Ok(img.resize(longest_edge, longest_edge, FilterType::Nearest)
        .save(fout, image::JPEG)?)
```

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
 
+[![rayon-badge]][rayon] [![glob-badge]][glob] [![image-badge]][image] [![cat-concurrency-badge]][cat-concurrency] [![cat-filesystem-badge]][cat-filesystem]
+
+This example uses the `glob`, `image`, and `rayon` crates to generate thumbnails for all .jpg files in the current directory. 
+The thumbnails are stored in a new folder called `thumbnails`.
+
+Files are found with [`glob::glob_with`] so we can match case insensitively on both `.jpg` and `.JPG`.
+
+```rust,no_run
+#[macro_use]
```
budziq: please hide the `error chain` import and setup

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+extern crate rayon;
+
+use std::path::Path;
+use std::fs::{create_dir_all, File};
+
+use glob::{glob_with, MatchOptions};
+use image::{FilterType, ImageError};
+use rayon::prelude::*;
+
+error_chain! {
```
budziq: please hide the `error chain` import and setup

src/concurrency.md
```diff
@@ -260,6 +342,8 @@ fn run() -> Result<()> {
 
 <!-- Reference -->
 
+[`DynamicImage::resize`]: https://docs.rs/image/*/image/enum.DynamicImage.html#method.resize
```
budziq: Please use this link in description.

src/concurrency.md
```diff
@@ -274,5 +358,6 @@ fn run() -> Result<()> {
 [`mpsc::channel`]: https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html
 [`multiple options`]: https://docs.rs/rayon/*/rayon/slice/trait.ParallelSliceMut.html
 [`num_cpus::get`]: https://docs.rs/num_cpus/*/num_cpus/fn.get.html
+[`par_iter`]: https://docs.rs/rayon/*/rayon/iter/trait.IntoParallelRefIterator.html#tymethod.par_iter
```
budziq: Please use this link in description.

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+fn make_thumbnail(original: &Path, thumb_dir: &Path, longest_edge: u32) -> Result<()> {
+    let img = image::open(original)?;
+    let fout = &mut File::create(thumb_dir.join(original))?;
+
+    img.resize(longest_edge, longest_edge, FilterType::Nearest)
+        .save(fout, image::JPEG)?;
+    Ok(())
+}
+
+quick_main!(run);
```
budziq: please hide the `error_chain` related line

src/concurrency.md
```diff
@@ -70,7 +71,84 @@ fn main() {
...
+        println!("{}", failure);
+    }
+
+    println!("Done");
+    Ok(())
+}
+
+/// Resize `original` to have a maximum dimension of `longest_edge` and save the
+/// resized image to the `thumb_dir` folder
+fn make_thumbnail(original: &Path, thumb_dir: &Path, longest_edge: u32) -> Result<()> {
```
budziq: I would prefer to use AsRef<Path> here. Please note that for this function to work correctly with two different types implementing AsRef<Path> (`original` and `thumb_dir`) you will most likely need to specify two separate generic arguments:
```rust
fn make_thumbnail<PA, PB>(original: PA, thumb_dir: PB, longest_edge: u32) -> Result<()>
where
    PA: AsRef<Path>,
    PB: AsRef<Path>,
{
```

budziq:
Merged! I've made some minor cleanup and squashed all the commits into a single one.

Thanks for sticking with it @nocduro!

PR #279, opened 2017-09-22
centralize links for badges, categories, and crates

### LINKED ISSUE #111: Centralize all the badge and category links
Is there any way we could avoid repeating the same badge and category links on every page that uses them? Perhaps a central file of links that gets included via some mdBook magic into every page?

For example these are repeated on 4 pages so far:

```
[cat-encoding-badge]: https://img.shields.io/badge/-encoding-red.svg
[cat-encoding]: https://crates.io/categories/encoding
```

This is tedious and error-prone on every new example we add.

lucasem:
Fixes #111.


budziq:
Thanks a lot @lucasem !

PR #297, opened 2017-09-26
Add "Run piped external commands" example

ludwigpacifici:
For a given <directory>, shows the `<number>` biggest
files/subdirectories and its human readable size.

Equivalant Unix command:
du -ah <directory> | sort -hr | head -n <number>

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -398,7 +398,7 @@ fn run() -> Result<()> {
 
 [![regex-badge]][regex] [![cat-text-processing-badge]][cat-text-processing]
 
-Reads a file named `application.log` and only outputs the lines 
+Reads a file named `application.log` and only outputs the lines
 containing “version X.X.X”, some IP address followed by port 443
 (e.g. “192.168.0.1:443”), or a specific warning.
 
@@ -622,7 +622,7 @@ fn main() {
 
 [![regex-badge]][regex] [![lazy_static-badge]][lazy_static] [![cat-text-processing-badge]][cat-text-processing]
 
-Replaces all occurrences of the hyphenated British English date pattern `2013-01-15` 
+Replaces all occurrences of the hyphenated British English date pattern `2013-01-15`
 with its equivalent slashed American English date pattern `01/15/2013`.
 
 The method [`Regex::replace_all`] replaces all occurrences of the whole regex. The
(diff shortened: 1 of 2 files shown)


src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+
+```rust,no_run
+# extern crate clap;
+
+# use clap::{App, Arg, ArgMatches};
+use std::error::Error;
+use std::fs::metadata;
+use std::process::{exit, Command, Stdio};
+
+static DU: &'static str = "du";
```
budziq: I'd remove these constants. The example will be shorter and more readable if we write the tool name and argument list directly in the `Command::new` builder chain

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+
+static SORT: &'static str = "sort";
+static SORT_ARGUMENTS: &'static str = "-hr";
+
+static HEAD: &'static str = "head";
+static HEAD_ARGUMENTS: &'static str = "-n";
+
+fn main() {
+    let arguments = declare_command_line_arguments();
+    let results = parse_number_of_results(&arguments);
```
budziq: i would rename the `results` variable to more descriptive name such as `max_results`

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+    let directory = parse_directory(&arguments);
+
+    let du_process = match Command::new(DU)
+        .arg(DU_ARGUMENTS)
+        .arg(directory)
+        .stdin(Stdio::null())
```
budziq: `stdin(Stdio::null())` this line is not neeaded

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        .stdout(Stdio::piped())
+        .spawn()
+    {
```
budziq: The whole error boilerplate for the `spawn` invocations can be replaced with a single `?`'s lets do it!

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+static DU: &'static str = "du";
+static DU_ARGUMENTS: &'static str = "-ah";
+
+static SORT: &'static str = "sort";
+static SORT_ARGUMENTS: &'static str = "-hr";
+
+static HEAD: &'static str = "head";
+static HEAD_ARGUMENTS: &'static str = "-n";
+
+fn main() {
```
budziq: lets change this `fn main()` to `fn run() -> Result<()>` then we will be able to use `?` in the function body to cut its length 2x times

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        .spawn()
+    {
+        Err(why) => {
+            println!("couldn't spawn {}: {}", DU, why.description());
+            exit(1)
+        }
+        Ok(process) => process,
+    };
+
+    let du_stdout = du_process.stdout.unwrap_or_else(|| {
```
budziq: I would suggest using 
```rust
stdout.ok_or_else(|| "Could not capture `du` output!")?
```
and we do not need a separate variable binding. please see my comment on `head_output` below

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        Ok(process) => process,
+    };
+
+    let head_output = head_process.wait_with_output().unwrap_or_else(|err| {
+        println!("Could not capture {} standard output: {}", HEAD, err);
+        exit(6);
+    });
+
+
+    match String::from_utf8(head_output.stdout) {
```
budziq: lets drop this whole boilerplate. and just perform the conversion from inside the `println!`

```rust
    println!(
        "Top {} biggest files and directories in {}\n{}",
        max_results,
        directory,
        String::from_utf8(head_output.stdout)?
    );
```

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+
+Shows the `<N>`th biggest files/subdirectories in the specified
+`<directory>`. It spawns three Unix processes. It is equivalant to
+run: `du -ah <directory> | sort -hr | head -n <N>`.
+
+```rust,no_run
+# extern crate clap;
+
+# use clap::{App, Arg, ArgMatches};
+use std::error::Error;
```
budziq: One we finish, the only `use` that will be needed here will be `use std::process::{Command, Stdio};`

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        exit(2);
+    });
+
+    let sort_process = match Command::new(SORT)
+        .arg(SORT_ARGUMENTS)
+        .stdin(du_stdout)
+        .stdout(Stdio::piped())
+        .spawn()
+    {
+        Err(why) => {
```
budziq: ditto on error handling

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+            println!("couldn't spawn {}: {}", SORT, why.description());
+            exit(3)
+        }
+        Ok(process) => process,
+    };
+
+    let sort_stdout = sort_process.stdout.unwrap_or_else(|| {
```
budziq: ditto on error handling and we do not need a separate variable binding. please see my comment on `head_output` below

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        println!("Could not capture {} standard output.", SORT);
+        exit(4);
+    });
+
+    let head_process = match Command::new(HEAD)
+        .arg(HEAD_ARGUMENTS)
```
budziq: ditto on error handling

src/basics.md
```diff
@@ -326,13 +327,185 @@ fn run() -> Result<()> {
...
+        .spawn()
+    {
+        Err(why) => {
+            println!("couldn't spawn {}: {}", HEAD, why.description());
+            exit(5)
+        }
+        Ok(process) => process,
+    };
+
+    let head_output = head_process.wait_with_output().unwrap_or_else(|err| {
```
budziq: No need fo the error boilerplate here and we do not need separate variable binding here. We can use 
```rust
let head_output = Command::new("head")
.arg("-n")
...
.spawn()?
.wait_with_output()?;
```

ludwigpacifici:
Thanks for your detailed review! I agree with your comments ; the code is shorter and to the point. Let me know what you think.

src/basics.md
```diff
@@ -326,13 +327,77 @@ fn run() -> Result<()> {
...
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#         Utf8(std::string::FromUtf8Error);
+#     }
+# }
+#
+fn run() -> Result<()> {
+    let directory = ".";
+    let du_process = Command::new("du")
```
budziq: We might be tempted to further DRY the code by extracting the builder into a separate fn like 

```rust
fn spawn_piped<P, I, A, IN>(program: P, args: I, input: IN) -> Result<Child>
where
    P: AsRef<OsStr>,
    I: IntoIterator<Item = A>,
    A: AsRef<OsStr>,
    IN: Into<Stdio>,
{
...
}
```

But I'm not sure if this does not make the code harder to read. I'll let you decide here

src/basics.md
```diff
@@ -326,13 +327,77 @@ fn run() -> Result<()> {
...
+    let head_process = Command::new("head")
+        .arg("-n")
+        .arg(max_results)
+        .stdin(sort_process)
+        .stdout(Stdio::piped())
+        .spawn()?
+        .wait_with_output()?;
+
+    println!(
+        "Top {} biggest files and directories in '{}':\n{}",
```
budziq: I guess that w can hardcode the "10" and lose the variable alltoheter

src/basics.md
```diff
@@ -326,13 +327,77 @@ fn run() -> Result<()> {
...
 
+[ex-run-piped-external-commands]: #ex-run-piped-external-commands
+<a name="ex-run-piped-external-commands"></a>
+## Run piped external commands
+
+[![std-badge]][std] [![cat-os-badge]][cat-os]
+
+Shows up to the 10<sup>th</sup> biggest files and subdirectories in
+the specified the current working directory. It spawns three Unix
+processes as external [`Command`]. It is equivalant to run: `du -ah . | sort -hr | head -n
```
budziq: I would mention the `Stdio::piped` in the description as it's critical here

budziq:
Also could you fix the merge conflict and squash the commits into one?

src/basics.md
ludwigpacifici: I gave it a try:
* [`pipe-process.rs`](https://gist.github.com/ludwigpacifici/d94dc241c2362270e9950aced26545dd)
* [`pipe-process-factorized.rs`](https://gist.github.com/ludwigpacifici/bd67799a15e8624008f7bcf268d482cc)

I prefer `pipe-process.rs`:
1. The goal is to demonstrate pipe processes, however, the `spawn_piped` function tends to bury it.
2. If the targeted reader is a beginner, `spawn_piped` introduce concepts of iterators. 
3. `spawn_piped` will DRY the code if more processes would be spawn.

If I miss something, let me know, otherwise I will skip `spawn_piped`.

ludwigpacifici:
Knowing that my commits are already pushed, can I squash and rebase them in this PR or should I create a new PR?

src/basics.md
```diff
@@ -326,13 +327,78 @@ fn run() -> Result<()> {
...
+    let sort_output = Command::new("sort")
+        .arg("-hr")
+        .stdin(du_output)
+        .stdout(Stdio::piped())
+        .spawn()?
+        .stdout
+        .ok_or_else(|| "Could not capture `sort` standard output.")?;
+
+    let head_output = Command::new("head")
+        .arg("-n")
```
budziq: you can also use `args(&["-n", "10"])`

src/basics.md
```diff
@@ -851,39 +917,40 @@ fn run() -> Result<()> {
...
-[`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
-[`Mutex`]: https://doc.rust-lang.org/std/sync/struct.Mutex.html
-[`RwLock`]: https://doc.rust-lang.org/std/sync/struct.RwLock.html
-[`MutexGuard`]: https://doc.rust-lang.org/std/sync/struct.MutexGuard.html
-[`Mmap::as_slice`]: https://docs.rs/memmap/*/memmap/struct.Mmap.html#method.as_slice
 [`seek`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.seek
-[`digest::Context`]: https://docs.rs/ring/*/ring/digest/struct.Context.html
-[`digest::Digest`]: https://docs.rs/ring/*/ring/digest/struct.Digest.html
 [rand-distributions]: https://doc.rust-lang.org/rand/rand/distributions/index.html
 [replacement string syntax]: https://docs.rs/regex/0.2.2/regex/struct.Regex.html#replacement-string-syntax
```
budziq: same here

budziq:
in regard to the suggested refactoring, I would not be hesitant to introduce iterators as these are already present very early in other examples (please think about cookbook as an appendix to RBE, you know the basics and would like to have an inkling how to solve your problem). I agree that the `spawn_piped` function might dilute the example somewhat, as discussed I'll leave the decision to you.

as for the [refactoring proposed by you](https://gist.github.com/ludwigpacifici/bd67799a15e8624008f7bcf268d482cc#file-pipe-process-factorized-rs-L49) you do not have to iterate in a for loop, the [`Command::args`](
https://doc.rust-lang.org/std/process/struct.Command.html#method.args) takes an iterator directly in a single call.

budziq:
> Knowing that my commits are already pushed, can I squash and rebase them in this PR or should I create a new PR?

Sure you can! You modify the history of your local branch (rebase/squash and whatnot) and just `git push --force` to you remote, this will automatically update the PR :+1:

src/basics.md
ludwigpacifici: I don't understand this one. Are you referring to this code, from your above comment?

```rust
        .arg("-ah")
        .arg(&directory)
```

ludwigpacifici:
Thank you for the advice 👍

src/basics.md
budziq: @ludwigpacifici 
Well that was not really descriptive on my part :disappointed: (I was doing 3x reviews concurrently at that time, teaches me not to multitask sorry!). 

What I mean was:
Please split this commit into one editing the example and second one editing the links 
and the "same here" was about making the regex links use wildcard "*" instead of "0.2.2" which happens in two places

ludwigpacifici: No worries @budziq, you do a good job 👍

ludwigpacifici:
Commits splitted. Any other change left to do?

budziq:
> Any other change left to do?

I would prefer for the links to be sorted case insensitively. otherwise we are good to merge once travis finishes :)

ludwigpacifici:
@budziq I tweaked my editor to change the sort.

budziq:
@ludwigpacifici Nicely done!

ludwigpacifici:
Thanks for your reviews @budziq !

budziq:
No Problem. Keep them comming :+1: 
You can see your example [already deployed](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#run-piped-external-commands)

Issue #292, opened 2017-09-25
add "Compile and link statically to a bundled C library" example


budziq:
Use [cc-rs](https://crates.io/crates/cc-rs) crate to compile and link statically to a bundled C library inside of a build.rs file.

Please note that it will be important to mention that this is a **build time** crate.
Most likely we will dedicate a separate (last) chapter like "build_time.md"

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/236

jaemk:
Note: The [`trust`](https://github.com/japaric/trust) repo is itself a basic example that could be copied and simplified.

budziq:
Right! @jaemk Would you be interested in taking lead on this example?

I see this as the first example that would have the longest explanation followed by the next 4 examples listed in https://github.com/rust-lang-nursery/rust-cookbook/issues/236 showing the possible minor differences (with much shorter descriptions)

jaemk:
Sure! I can take a look at this tomorrow and setup a new `build_tools` file

budziq:
merged

PR #298, opened 2017-09-27
Add build tools section & basic `cc` example

jaemk:
issue #292

@budziq Here's what I've got so far. Were you thinking of a more in-depth build-tools intro? Does the wording make sense?

### DIFF
--- a/src/build_tools.md
+++ b/src/build_tools.md
@@ -1,0 +1,128 @@
+# Build Time Tooling
+
+This section covers "build-time" tooling, or code that is run prior to compiling a crate's source code.
+Conventionally, build-time code lives in a **build.rs** file and is commonly referred to as a "build script".
+Common use cases include rust code generation and compilation of bundled C/C++/asm code.
+See crates.io's [documentation on the matter][build-script-docs] for more information.
+
+
+| Recipe | Crates | Categories |
+|--------|--------|------------|
+| [Compile and link statically to a bundled C library][ex-cc-static-bundled] | [![cc-badge]][cc] | [![cat-development-tools-badge]][cat-development-tools] |
+
+
+[ex-cc-static-bundled]: #ex-cc-static-bundled
+<a name="ex-cc-static-bundled"></a>
+## Compile and link statically to a bundled C library
+
+[![cc-badge]][cc] [![cat-development-tools-badge]][cat-development-tools]
+
+To accommodate scenarios where additional C, C++, or assembly is required in a project, the [**cc**][cc] crate
+offers a simple api for compiling bundled C/C++/asm code into static libraries (**.a**) that can be statically linked to by **rustc**.
+
+The following example has some bundled C code (**src/hello.c**) that will be used from rust.
+Before compiling our rust source code, the "build" file (**build.rs**) specified in **Cargo.toml** will run.
+Using the [**cc**][cc] crate, a static library file will be produced (in this case, **libhello.a**, see
+[`compile` docs][cc-build-compile]) which can then be used from rust by declaring the external function signatures in an `extern` block.
+
+Since the bundled C is very simple, only a single source file needs to be passed to [`cc::Build`][cc-build].
+For more complex build requirements, [`cc::Build`][cc-build] offers a full suite of builder methods for specifying
+[`include`][cc-build-include] paths and extra compiler [`flag`][cc-build-flag]s.
+
+### `Cargo.toml`
+
+```toml
+[package]
+...
+build = "build.rs"
+
+[build-dependencies]
+cc = "1"
+
+[dependencies]
+error-chain = "0.11"
+```
+
+### `build.rs`
+
+```rust,no_run
+extern crate cc;
+
+fn main() {
+    cc::Build::new()
+        .file("src/hello.c")
+        .compile("hello");   // outputs `libhello.a`
+}
+```
+
+### `src/hello.c`
+
+```c
+#include <stdio.h>
+
+
+void hello() {
+    printf("Hello from C!\n");
+}
+
+void greet(const char* name) {
+    printf("Hello, %s!\n", name);
+}
+```
+
+### `src/main.rs`
+
+```rust,no_run
+# #[macro_use] extern crate error_chain;
+use std::ffi::CString;
+use std::os::raw::c_char;
+
+#
+# error_chain! {
+#     foreign_links {
+#         NulError(::std::ffi::NulError);
+#         Io(::std::io::Error);
+#     }
+# }
+# 
+# 
+# fn prompt(s: &str) -> Result<String> {
+#     use std::io::Write;
+#     print!("{}", s);
+#     std::io::stdout().flush()?;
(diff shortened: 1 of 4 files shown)


src/build_tools.md
```diff
@@ -0,0 +1,122 @@
+# Build Tools
+
+This section covers "build-time" tooling, or code that is run prior to compiling a crate's source code.
+Conventionally, build-time code lives in a `build.rs` file and is commonly referred to as a "build script".
+Common use cases include `rust` code generation and compilation of bundled `C/C++/asm` code.
+See `crates.io`'s [documentation on the matter][build-script-docs] for more information.
+
+Notes:
```
budziq: I would remove the notes section here. The play button will not work anyway as these examples will be `no_run` anyway.

src/build_tools.md
```diff
@@ -0,0 +1,122 @@
+# Build Tools
```
budziq: In the hindsight I'm not sure if "Development Tools" would not be a better name but I'm willing to crowd source here. What do you think @jaemk ?

src/build_tools.md
```diff
@@ -0,0 +1,122 @@
...
+error-chain = "0.11"
+```
+
+### `build.rs`
+
+```rust,no_run
+extern crate cc;
+
+fn main() {
+    // builds a static lib: `libhello.a`
```
budziq: I would prefer to have this information in the textual description above instead of a comment here.

src/build_tools.md
```diff
@@ -0,0 +1,122 @@
...
+    cc::Build::new()
```
budziq: I'd love to see some hyperlinked references pointing to the used structs and methods as this is the firs formative example in this section.

jaemk:
Thanks @budziq! I'll get to this later tonight!

src/build_tools.md
jaemk: Now that I'm thinking about it, I'm wondering what other types of examples (other than `cc` stuff) will eventually be included here, if any. Does something of this nature deserve a more specifically named category? Maybe one of the following?
- `External Build Tools`
- `Custom Build Steps`
- `Building FFI Libraries`
- `Compiling and Integrating Non-Rust Code`
- `Build Script Support` (the conventional name [used by cargo](http://doc.crates.io/build-script.html))

If not, I think `Development Tools` might be better for covering a broader topic-space.

budziq: Hmm in time we might want to have here other types of support tooling such as
- [`syn`](https://crates.io/crates/syn) 
- [`cmake`](https://crates.io/crates/cmake)
- [`pkg-config`](https://crates.io/crates/pkg-config)

so imho `Development Tools`, `Compile time` or `Build time dependencies` might be appropriate (I'm leaning towards the last one). But I'll let you chose, its easy enough to change anytime ;)

jaemk: What do you think of `Build Time Tooling` or `Build Time Tools`?

src/build_tools.md
jaemk: I've made note of the output name in the example description, including a link to the `compile` method that talks about the naming scheme of output files. But I've also kept a comment nearby to make it clear that the compile output may not necessarily match the specified name.

jaemk:
@budziq I went with `Build Time Tooling` (thoughts?) and cleaned up / added extra info to the first example's description.

src/links.md
```diff
@@ -17,8 +17,12 @@ Keep lines sorted.
 [cat-cryptography]: https://crates.io/categories/cryptography
 [cat-debugging-badge]: https://badge-cache.kominick.com/badge/debugging--x.svg?style=social
 [cat-debugging]: https://crates.io/categories/debugging
+[cat-development-tools-badge]: https://badge-cache.kominick.com/badge/development_tools--x.svg?style=social
+[cat-development-tools]: https://crates.io/categories/development-tools
 [cat-encoding-badge]: https://badge-cache.kominick.com/badge/encoding--x.svg?style=social
 [cat-encoding]: https://crates.io/categories/encoding
+[cat-external-ffi-bindings-badge]: https://badge-cache.kominick.com/badge/external_ffi_bindings--x.svg?style=social
+[cat-external-ffi-bindings]: https://crate.io/categories/external-ffi-bindings
```
budziq: typo in the domain ;)

src/build_tools.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+
+{{#include links.md}}
+
+<!-- Other Reference -->
+
+[build-script-docs]: http://doc.crates.io/build-script.html
+[playground]: https://play.rust-lang.org
+[cc-build]: https://docs.rs/cc/1.0.0/cc/struct.Build.html
```
budziq: I would suggest using wildcard version numbers in doc.rs links

src/build_tools.md
```diff
@@ -0,0 +1,128 @@
+# Build Time Tooling
+
+This section covers "build-time" tooling, or code that is run prior to compiling a crate's source code.
+Conventionally, build-time code lives in a `build.rs` file and is commonly referred to as a "build script".
```
budziq: Sorry to catch this this late but generally we strive to use backticks <code>``</code> for code /symbols or external commands and for other text we use different form of emphasis (italics or bold). So I would change the method of emphasis for the file/language/crate/website names and use the backticks only for symbols and language keywords (`extern`)

src/intro.md
```diff
@@ -117,6 +117,13 @@ community. It needs and welcomes help. For details see
 | [Log to the Unix syslog][ex-log-syslog] | [![log-badge]][log] [![syslog-badge]][syslog] | [![cat-debugging-badge]][cat-debugging] |
 | [Log messages to a custom location][ex-log-custom] | [![log-badge]][log] | [![cat-debugging-badge]][cat-debugging] |
 
+## [Build Time Tooling](build_tools.html)
+
+| Recipe | Crates | Categories |
+|--------|--------|------------|
+| [Compile and link statically to a bundled C library][ex-cc-static-bundled] | [![cc-badge]][cc] | [![cat-development-tools-badge]][cat-development-tools] [![cat-rust-patterns-badge]][cat-rust-patterns] [![cat-external-ffi-bindings-badge]][cat-external-ffi-bindings] |
```
budziq: I'm not sure about the "rust-patterns" badge is appropriate here (infact the the "external-ffi-bindings" to as these is pretty much reselved for "*sys" crates and the bindgen)

jaemk:
@budziq No problem, thanks for the thorough read-through!

src/intro.md
jaemk: I also wasn't too sure which categories were appropriate when I first start this, so I just chose a couple and forgot to ask you earlier! I agree, just `Development Tools` seems appropriate now.

src/links.md
jaemk: good catch!

jaemk:
@budziq let me know if you see anything else

budziq:
Excellent work! Thanks @jaemk

Issue #290, opened 2017-09-24
add "Check webpage for broken links" example


budziq:
This is basically an expanded version of [extract links webpage example](https://rust-lang-nursery.github.io/rust-cookbook/net.html#ex-extract-links-webpage)

Use [reqwest](https://crates.io/crates/reqwest) - in conjunction with [select](https://crates.io/crates/select) to check if links on a webpage are not broken (i.e not 404's.)

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/84

sb89:
I'll have a look at this.

budziq:
:+1:

budziq:
merged

PR #299, opened 2017-09-27
Added "Check webpage for broken links" example

sb89:
#290 

I've maybe made this too complicated. Let me know what you think.


src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+#       IoError(std::io::Error);
+#       UrlParseError(url::ParseError);
+#   }
+# }
+
+fn run() -> Result<()> {
+    let mut broken_urls = HashSet::new();
+
+    let url = Url::parse("https://www.rust-lang.org/en-US/")?;
+    let mut base_url = Url::parse(&format!(
```
budziq: this whole block could be substituted with 
```rust
let base_url = Url::parse(&parsed[..url::Position::BeforePath])?;
```
and base_url will not need to be mut here

also I would extrat the whole link acquisition and transforming into a separate function for clarity

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+        "{}://{}",
+        url.scheme(),
+        url.host_str().unwrap_or("")
+    ))?;
+
+    let res = reqwest::get(url.as_str())?;
```
budziq: no need for `as_str` here

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+        base_url = Url::parse(base_tag_href)?;
+    }
+
+    let links: HashSet<&str> = document
+        .find(Name("a"))
+        .filter_map(|n| n.attr("href"))
+        .collect();
+
+    for link in links {
+        let link_url = Url::parse(link).or_else(|e| match e {
```
budziq: You can use `ParseOptions` here to make the crate do all the work here (if the link is absolut then it will be used otherwise the `base_url` will be used )

```rust
 let base_parser = Url::options().base_url(Some(&base_url));
...
 for link in links { //or `map` here
     let final_url = base_parser.parse(link)?;
```

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+            .filter_map(|n| n.attr("href"))
+            .nth(0)
+    {
+        base_url = Url::parse(base_tag_href)?;
+    }
+
+    let links: HashSet<&str> = document
+        .find(Name("a"))
+        .filter_map(|n| n.attr("href"))
+        .collect();
```
budziq: we can do the transform (making the paths absolut) here before collecting.

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+
+    for link in links {
+        let link_url = Url::parse(link).or_else(|e| match e {
+            ParseError::RelativeUrlWithoutBase => base_url.join(link),
+            _ => Err(e),
+        });
+
+        match link_url {
```
budziq: instead of matching, I would suggest using iterator combinators here to extract only the error cases with their links

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+            Err(e) => println!("Error parsing URL ({}): {}", link, e.description()),
+
+            Ok(link_url) => {
+                let res = reqwest::get(link_url.as_str())?;
```
budziq: I would extract this super small logic to separate function for clarity.

src/net.md
```diff
@@ -957,8 +1058,10 @@ fn run() -> Result<()> {
 [`Cow`]: https://doc.rust-lang.org/std/borrow/enum.Cow.html
 [`Document::from_read`]: https://docs.rs/select/*/select/document/struct.Document.html#method.from_read
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
+[`HashSet`]: https://doc.rust-lang.org/std/collections/struct.HashSet.html
 [`Ipv4Addr`]: https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html
 [`Name`]: https://docs.rs/select/*/select/predicate/struct.Name.html
+[`ParseError::RelativeUrlWithoutBase`]: https://docs.rs/url/1.5.1/url/enum.ParseError.html#variant.RelativeUrlWithoutBase
```
budziq: please use wildcard version when pointing to docs.rs

src/net.md
```diff
@@ -876,6 +877,106 @@ fn run() -> Result<()> {
...
+<a name="ex-check-broken-links"/>
+## Check a webpage for broken links
+
+[![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] [![cat-net-badge]][cat-net] |
+
+Use [`Url::parse`] to create a [`Url`] and then create the base URL from that. Use [`reqwest::get`] to perform a HTTP GET request and then use [`Document::from_read`] to parse the response into a HTML document.
+
+We can then check if the document has a "base" tag. If so, we get the "href" [`attr`] from the first occurrence of the "base" tag. This is then used as the base URL.
+
+We can then retrieve all the links from the document by using [`find`] with the criteria of the [`Name`] being "a".
```
budziq: We already have a [extract all links from a webpage](https://rust-lang-nursery.github.io/rust-cookbook/net.html#extract-all-links-from-a-webpage-html) example, where nearly all of these information is presented. Lets build on it and not duplicate the information. This is a more advanced example and I would suggest to limit the description to only the new information not covered by earlier examples.

If you think that some of this information is not covered so far please consider updating the earlier example descriptions :+1:

sb89:
I've implemented changes in line with your comments.

sb89:
I believe the reviews you have made is on outdated code. I performed an additional commit about 40 minutes ago.

budziq:
> I believe the reviews you have made is on outdated code. I performed an additional commit about 40 minutes ago.

Yeah I know I'm on a conference and the wifi is flaky at best (these are reviews that were to be sent few hours ago). Standby I'll publish an updated review shortly. There should be only minor changes and were good to go.

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
...
+    let document = Document::from_read(res)?;
+
+    let base_url = get_base_url(&url, &document)?;
+
+    let links: HashSet<&str> = document
+        .find(Name("a"))
+        .filter_map(|n| n.attr("href"))
+        .collect();
+
+    let broken_links = get_broken_links(&base_url, &links)?;
```
budziq: I would extract only the link checking logic into a separate function `check_link`, it would take only a single `Url` (or better yet `AsRef<Url>`). Then such function could be `map`'ed over the collection iterator instead of collected into yet another `HashSet`

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
...
+fn get_base_url(url: &Url, doc: &Document) -> Result<Url> {
+    let base_tag_href = doc.find(Name("base")).filter_map(|n| n.attr("href")).nth(0);
+
+    let base_url = base_tag_href.map_or_else(|| Url::parse(&url[..Position::BeforePath]), Url::parse)?;
+    
+    Ok(base_url)
+}
+
+fn get_broken_links<'a>(base_url: &Url, links: &HashSet<&'a str>) -> Result<HashSet<&'a str>> {
+    let mut broken_links = HashSet::new();
```
budziq: I would for this function to operate only on a single `Url`  `check_link`, it would take only a single `Url` (or better yet `AsRef<Url>`) that was already made absolute (making the `base_url` param unnecessary). Then such function could be `map`'ed over the collection iterator instead of collected into yet another `HashSet`

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
...
+
+We parse the URL using [`Url::parse`] and then retrieve the document using reqwest.
+
+We call "get_base_url" to retrieve the base URL. If the document has a "base" tag, we get the "href" [`attr`] from the first occurrence of the "base" tag. This is then used as the base URL. Otherwise, we can use [`Position::BeforePath`] with the original URL to get the base of that URL.
+
+We then retrieve all the links from the document, and [`collect`] them into a [`HashSet`] to remove any duplicates.
+
+Now that we have our URL, we pass them into the "get_broken_links" function along with the base URL. This will iterate over the URL, use [`url::ParseOptions`] and [`Url::parse`] for each to get the absolute URL and then perform a HTTP GET request. If the response status is 404, the URL will be added to the HashSet which is returned by the function.
+
+Finally, we check the contents of the HashSet. If it is empty, we report that no links are broken, otherwise we iterate over the HashSet and report the broken links.
```
budziq: I would suggest to describe only the finer details, i.e. things that might not be obvious from the code.

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-check-broken-links]: #ex-check-broken-links
+<a name="ex-check-broken-links"/>
+## Check a webpage for broken links
+
+[![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] [![cat-net-badge]][cat-net]
+
+We parse the URL using [`Url::parse`] and then retrieve the document using reqwest.
```
budziq: The user should ba already familiar with this information from the previous examples.

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
...
+
+We call "get_base_url" to retrieve the base URL. If the document has a "base" tag, we get the "href" [`attr`] from the first occurrence of the "base" tag. This is then used as the base URL. Otherwise, we can use [`Position::BeforePath`] with the original URL to get the base of that URL.
```
budziq: I would try to pare the explanation down to only the information that is not mentioned in earlier examples.

src/net.md
budziq: Also there seam to be conflicts making the merge impossible. Could you look into that before squashing the PR commits into single one?

src/net.md
```diff
@@ -876,6 +877,99 @@ fn run() -> Result<()> {
...
+
+    let res = reqwest::get(url.as_ref())?;
+    let document = Document::from_read(res)?;
+
+    let base_url = get_base_url(&url, &document)?;
+
+    let links: HashSet<&str> = document
+        .find(Name("a"))
+        .filter_map(|n| n.attr("href"))
+        .collect();
```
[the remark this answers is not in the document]
sb89: How would I do this as `parse` returns a `Result` and therefore I cannot use the "?" operator when iterating? Should I just ignore any URL that cannot be parsed (and report no errors) like this?

    let links: HashSet<Url> = document
        .find(Name("a"))
        .filter_map(|n| n.attr("href"))
        .filter_map(|link| base_parser.parse(link).ok())
        .collect();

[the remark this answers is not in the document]
budziq: Well that entirely depends on what you'd like to achieve. Test only the correct links on best effort basis (your suggestion) or error-out even if there is at least one error
by collecting into `Result`
```rust
    let hs: Result<HashSet<_>> =  document
    .find(Name("a"))
    .filter_map(|n| n.attr("href"))
        .map(|link| base_parser.parse(link).map_err(|e| e.into()))
        .collect();
```

Unfortunately to do this trick with error_chain you'll have to jump through hoops with `map_err` as above (normally this would not be required)

src/net.md
```diff
@@ -876,6 +877,86 @@ fn run() -> Result<()> {
...
+
+    let base_parser = Url::options().base_url(Some(&base_url));
+
+    let links: HashSet<Url> = document
+        .find(Name("a"))
+        .filter_map(|n| n.attr("href"))
+        .filter_map(|link| base_parser.parse(link).ok())
+        .collect();
+
+    let broken_links = links.iter().filter(|link| {
```
budziq: no need for intermediate `broken_links` variable. we can iterate over the filtered iterator in a for loop

src/net.md
```diff
@@ -876,6 +877,86 @@ fn run() -> Result<()> {
...
+        let link_ok = check_link(link);
+
+        link_ok.is_ok() && !link_ok.unwrap()
```
budziq: please avoid unwrapping. 
`check_link` could return a custom error defined in `error_chain!` when link is broken,
then you could extract all the errored values with `filter_map(|res| res.err())` also

src/net.md
```diff
@@ -876,6 +877,86 @@ fn run() -> Result<()> {
...
+extern crate reqwest;
+extern crate select;
+extern crate url;
+
+use url::{Url, Position};
+use reqwest::StatusCode;
+use select::document::Document;
+use select::predicate::Name;
+
+use std::collections::HashSet;
```
budziq: std use would go above the crate ones

src/net.md
```diff
@@ -876,6 +877,86 @@ fn run() -> Result<()> {
...
+#
+# error_chain! {
+#   foreign_links {
+#       ReqError(reqwest::Error);
+#       IoError(std::io::Error);
+#       UrlParseError(url::ParseError);
+#   }
+# }
+fn get_base_url(url: &Url, doc: &Document) -> Result<Url> {
```
budziq: we would need a new line between the `error_chain` block and the fn

sb89:
I've implemented changes as requested.

Instead of removing the unwrap and using a custom error, I've done something slightly different with the filter. Let me know what you think and if everything is OK I'll squash the commits.

budziq:
@sb89 very cool! Please squash + rebase the commits and we're ready to merge :+1:

budziq:
merged! Thanks for sticking with it @sb89 !

Issue #283, opened 2017-09-24
add "Parse and modify a version string" example


budziq:
Parse a simple hardcoded version string such as "0.2.6" and increment the major, minor or patch version. Use asserts to verify correct serialization and deserailization.

Use the [semver](https://github.com/steveklabnik/semver) crate

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/276

mykalu:
I'd like to try taking this on!

budziq:
merged

PR #305, opened 2017-09-30
Add "Parse and increment a version string" example

mykalu:
#283 

I'm actually not certain if this follows best practices or even the intended goal of the example.
Feedback appreciated!


mykalu:
Potential points to note:
- Changed name of example from "modify" to "increment" because incrementing is the only modification the crate enables.
- Prematurely assigned the `semver` crate to the [`Configuration`](https://crates.io/categories/config) category; the crate is not currently in any category, though I have mentioned this on [the crate's evaluation thread](https://internals.rust-lang.org/t/out-of-band-crate-evaluation-for-2017-10-09-semver/5929)!
- Not entirely sure if I'm implementing `error-chain` correctly...

budziq:
That's excellent work @mykalu !
I've merged it after minimal cleanup (squashed commits into single one, small wording update and moved the the external reference link to the bottom of the file)

Issue #291, opened 2017-09-24
add "Make a partial download with HTTP range headers" example


budziq:
Use [reqwest](https://crates.io/crates/reqwest) - in conjunction with HTTP range headers to make a partial download of a large file with progress.

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/84

sofiageo:
Seems like @jbolila beat me to it :) I didn't make any request comment because i didn't know how far I could go, but if was my first try with rust. 

I will still post my code for any comments 
```
#[macro_use]
extern crate error_chain;
extern crate reqwest;

use std::io::Read;
use reqwest::header::{Headers, Range};

error_chain! {
    foreign_links {
        Io(std::io::Error);
        HttpRequest(reqwest::Error);
        ParseRequest(reqwest::UrlError);
    }
}

fn run() -> Result<()> {
    let url = reqwest::Url::parse("https://rust-lang-nursery.github.io/rust-cookbook/")?;
    let client = reqwest::Client::new()?;

    print!("Making a HEAD request: ");
    let mut req = reqwest::Request::new(reqwest::Method::Head, url.clone());
    let mut res = client.execute(req)?;
    println!("done!");
    let headers= res.headers();
    let supported = headers.get_raw("Accept-Ranges");
    if supported.is_some() {
            let mut headers = Headers::new();
            headers.set(Range::bytes(1, 100));

            println!("Making a GET request: ");
            let mut res = client.get(url.clone())?.headers(headers).send()?;
            let mut body = String::new();
            res.read_to_string(&mut body)?;

            println!("Status: {}", res.status());
            println!("Headers:\n{}", res.headers());
            println!("Body:\n{}", body);
    }

    Ok(())
}

quick_main!(run);
```

budziq:
@sofiageo no reason to shy away from claiming an issue, we are not pressed for any deadline here! Just pick another issue and start working on it.

In regard to how this issue should be done, please checkout @jbolila 's PR once it's ready (I'll probably do a review there tomorrow). Your code is quite nice but with few suggestions:
- use convenience method https://docs.rs/reqwest/0.8.0/reqwest/struct.Client.html#method.head instead of `reqwest::Request::new(reqwest::Method::Head`
- use typed api `headers.get` instead of `headers.get_raw`
- no need to create temporary `mut headers` variable it can be constructed within the `?.headers(..).send` call
[...]

PR #309, opened 2017-10-01
Add "Make a partial download with HTTP range headers" example #291

jbolila:
This is my first PR in a Rust project, probably not the most idiomatic Rust but I'm here to learn.

### DIFF
--- a/src/net.md
+++ b/src/net.md
@@ -1069,8 +1069,8 @@ fn run() -> Result<()> {
     println!("content-lenght: {} download in progress..", content_length);
 
     let mut chunk_starts: u64 = 0;
-    let mut chunk_ends = if (chunk_starts + 100) > content_length {
-        content_length % 100 - 1
+    let mut chunk_ends = if 100 > content_length {
+        content_length - 1
     } else {
         99
     };
--- a/src/net.md
+++ b/src/net.md
@@ -19,7 +19,7 @@
 | [Extract all links from a webpage HTML][ex-extract-links-webpage] | [![reqwest-badge]][reqwest] [![select-badge]][select] | [![cat-net-badge]][cat-net] |
 | [Check webpage for broken links][ex-check-broken-links] | [![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Extract all unique links from a MediaWiki markup][ex-extract-mediawiki-links] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
-| [Make a partial download with HTTP range headers][ex-progress-with-range] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
+| [Make a partial download with HTTP range headers][ex-progress-with-range] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
 
 [ex-url-parse]: #ex-url-parse
 <a name="ex-url-parse"/>
@@ -1026,8 +1026,11 @@ fn run() -> Result<()> {
 
 [![reqwest-badge]][reqwest] [![cat-net-badge]][cat-net]
 
-Download content using [`reqwest::get`] and setting the [`reqwest::header::Range`] to do partial
-downloads, and between these show a basic progress bar.
+Uses [`reqwest::Client::head`] to get the content-length and validate if the server sets the header
+[`reqwest::header::ContentRange`], required to confirm the support of partial downloads.
+
+If supported downloads the content using [`reqwest::get`], setting the [`reqwest::header::Range`]
+to do partial downloads in chunks of 100 bytes, between these writes basic progress messages.
 
 Range header, defined in [RFC7233][HTTP Range RFC7233].
 
@@ -1035,65 +1038,68 @@ Range header, defined in [RFC7233][HTTP Range RFC7233].
 # #[macro_use]
 # extern crate error_chain;
 extern crate reqwest;
-extern crate url;
+
+use std::io::Read;
 
 use reqwest::header::{Range, ContentRange, ContentRangeSpec};
-use url::Url;
-
-use std::io::{Read, Write};
 
 # error_chain! {
 #     foreign_links {
 #         Io(std::io::Error);
 #         Reqwest(reqwest::Error);
-#         UrlParse(url::ParseError);
 #     }
 # }
 
-const TOTAL_CHUNKS: u64 = 20;
-
 fn run() -> Result<()> {
-    let client = reqwest::Client::new()?;
+    let client = reqwest::Client::new();
 
-    let url = Url::parse("https://httpbin.org/range/1024?duration=2")?;
-    let response = client.head(url.to_owned())?.send()?;
+    let url = "https://httpbin.org/range/1024?duration=2";
+    let response = client.head(url).send()?;
 
-    if let Some(range) = response.headers().get::<ContentRange>() {
-        if let ContentRangeSpec::Bytes { instance_length, .. } = **range {
-            let content_length = instance_length.unwrap_or(0);
-            let chunk_size = content_length / TOTAL_CHUNKS;
+    let range = response.headers().get::<ContentRange>().ok_or(
+        "response doesn't include the expected ranges",
+    )?;
 
-            let mut content: Vec<u8> = Vec::with_capacity(content_length as usize);
-            print!("content-lenght {} download in progress", content_length);
+    let content_length = match **range {
+        ContentRangeSpec::Bytes { instance_length: Some(len), .. } => len,
+        _ => bail!("response doesn't include the expected ranges"),
+    };
 
-            for chunk in 0..TOTAL_CHUNKS {
-                print!(".");
-                std::io::stdout().flush()?;
+    let mut content: Vec<u8> = Vec::with_capacity(content_length as usize);
+    println!("content-lenght: {} download in progress..", content_length);
 
-                let mut response = client
-                    .get(url.to_owned())?
-                    .header(Range::bytes(
-                        chunk * chunk_size,
-                        (chunk + 1) * chunk_size - 1,
-                    ))
-                    .send()?;
-                response.read_to_end(&mut content)?;
-            }
+    let mut chunk_starts: u64 = 0;
+    let mut chunk_ends = if (chunk_starts + 100) > content_length {
+        content_length % 100 - 1
+    } else {
+        99
+    };
 
-            if content_length % TOTAL_CHUNKS != 0 {
-                let mut response = client
-                    .get(url.to_owned())?
-                    .header(Range::bytes(TOTAL_CHUNKS * chunk_size, content_length - 1))
-                    .send()?;
-                response.read_to_end(&mut content)?;
-            }
+    while chunk_starts < content_length {
+        println!("{}-{}", chunk_starts, chunk_ends);
 
-            assert_eq!(content_length as usize, content.len());
-            println!("finished with success!");
-        } else {
-            bail!("response doesn't include the expected ranges");
+        let mut response = client
+            .get(url)
+            .header(Range::bytes(chunk_starts, chunk_ends))
+            .send()?;
+
+        if response.status() != reqwest::StatusCode::PartialContent &&
+            response.status() != reqwest::StatusCode::Ok
+        {
+            bail!("Unexpected server response: {}", response.status())
         }
+
+        response.read_to_end(&mut content)?;
+
+        chunk_starts += 100;
+        chunk_ends += if (chunk_ends + 100) > content_length {
+            content_length % 100
+        } else {
+            100
+        };
     }
+
+    println!("finished with success!");
 
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


src/net.md
```diff
@@ -1070,6 +1152,7 @@ fn run() -> Result<()> {
 [`reqwest::RequestBuilder`]: https://docs.rs/reqwest/0.6.2/reqwest/struct.RequestBuilder.html
 [`reqwest::Response`]: https://docs.rs/reqwest/*/reqwest/struct.Response.html
 [`reqwest::get`]: https://docs.rs/reqwest/*/reqwest/fn.get.html
+[`reqwest::header::Range`]: https://docs.rs/reqwest/0.7.3/reqwest/header/enum.Range.html
```
budziq: please use wildcard version `*` in docs.rs links (please also update any other links with hardcoded version that you find in this file)

src/net.md
```diff
@@ -1022,6 +1023,87 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-progress-with-range]: #ex-progress-with-range
+<a name="ex-progress-with-range"/>
+## Make a partial download with HTTP range headers
+
+[![reqwest-badge]][reqwest] [![cat-net-badge]][cat-net]
+
+Download content using [`reqwest::get`] and setting the [`reqwest::header::Range`] to do partial
```
budziq: I would mention about getting the `ContentRange` with `head` request.

jbolila:
Thank you @budziq for your comments, relay appreciated. I'm no sure about the rebase I did in my fork to get the latest changes and the update of the `Cargo.toml` .. (i don't have much experience with PR)

budziq:
@jbolila It seams that there was a problem with the rebase. You have pulled and rebased some of the already merged commits making this PR not mergable. 
I would suggest backtracking a little.

```bash
git remote add upstream git@github.com:rust-lang-nursery/rust-cookbook.git
git remote update                                # get latest changes
git checkout f46683b -b jbolila_pr_fix  # get your original commit
git rebase upstream/master                # place it on top of current master
git cherry-pick b1bf49dec8                  # add your latest fixes
```
then in the branch "jbolila_pr_fix" you'll have the state that you will want to push to your origin/master (with --force) to update the PR into a proper condition.

Then we will continue with the review.
If you'll have any further problems, do not hesitate to ask.

jbolila:
I hope is right, now.

```
*   a0370ed (HEAD -> master) Merge branch 'jbolila_pr_fix'
|\  
| * fcacb9f (jbolila_pr_fix) Changes requested in the PR 309
| * 4e675fb Add "Make a partial download with HTTP range headers" example #291
| * b7ab129 (upstream/master) Parse a complex version string.
| * 6ffdf8e Minor fixes
..
```

budziq:
unfortunately not yet. instead of merging jbolila_fix_pr you should overwrite master with it.

```bash
git checkout jbolila_fix_pr
git branch -D master
git checkout -b master
git push -u origin master --force
```

src/net.md
```diff
@@ -1019,6 +1020,93 @@ fn run() -> Result<()> {
...
+    let content_length = match **range {
+        ContentRangeSpec::Bytes { instance_length: Some(len), .. } => len,
+        _ => bail!("response doesn't include the expected ranges"),
+    };
+
+    let mut content: Vec<u8> = Vec::with_capacity(content_length as usize);
+    println!("content-lenght: {} download in progress..", content_length);
+
+    let mut chunk_starts: u64 = 0;
+    let mut chunk_ends = if 100 > content_length {
```
budziq: Although I would suggest not to hardcode "100" here but have some value/constant indicating the buffer size

Also are you upto converting the offset calculation code into an iterator returning `Range::bytes`? That would be awesome!

src/net.md
```diff
@@ -1019,6 +1020,93 @@ fn run() -> Result<()> {
...
+    let range = response.headers().get::<ContentRange>().ok_or(
+        "response doesn't include the expected ranges",
+    )?;
+
+    let content_length = match **range {
+        ContentRangeSpec::Bytes { instance_length: Some(len), .. } => len,
+        _ => bail!("response doesn't include the expected ranges"),
+    };
+
+    let mut content: Vec<u8> = Vec::with_capacity(content_length as usize);
```
budziq: most of the time when we use partial downloading it's due to the fact that we either want to resume the download later or the whole content is to larg to hold in a contiguous memory buffer. 

I would suggest using `content` buffer length as the each partial content request length (upto the last possibly shorter one). and each time writing it into a file with given offset.

src/net.md
jbolila: I can try that one, although looks like in the `std` `Range::step_by()` is still in the nightly build only [#27741](https://github.com/rust-lang/rust/issues/27741). I'll do it for now manually inside the for loop.

src/net.md
jbolila: I'll use the value of the constant to set the size of the buffer. 

What will be the filename? wouldn't that be to much for this example, I'm just concerned about the amount of code for the reqwest/ranges compared to the file handle.

src/net.md
budziq: the `step_by` would not help here much as it would just return every N'th value of the range (treated as iterator here) so
```rust
 for s in (0..1024).step_by(100) {
    println!("{:?}", s);
 }
```
would give 
```
0
100
200
300
400
500
600
700
800
900
1000
```

Which would help with the implementation but would not give a full solution. 
I would propose to make own iterator returning 
`hyper::header::Range::bytes` with each iteration containing the next range we want to obtain

src/net.md
budziq: opening and writing to a file would be only two additional lines which we could hide with `#` anyway. the filename can be anything like "download.bin"

src/net.md
budziq: Anyhow, if you'll have any problems or concerns I'll gladly help!

jbolila: @budziq please let me know if it's ok, and if you wanna hide some of these extra lines. Thanks!

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+
+        let mut response = client.get(url).header(range).send()?;
+
+        if response.status() != reqwest::StatusCode::PartialContent &&
+            response.status() != reqwest::StatusCode::Ok
+        {
+            bail!("Unexpected server response: {}", response.status())
+        }
+
+        let mut content: Vec<u8> = Vec::with_capacity(BUFFER_SIZE as usize);
```
budziq: no need to allocate the buffer on each iteration. Lets allocate it before the loop and use `Vec::clear` after each iteration

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+    let mut output_file = File::create("download.bin")?;
+    let ranges = PartialRange::new(range);
+
+    println!("starting download...");
+    for range in ranges.iter() {
+        println!("range {}", range);
+
+        let mut response = client.get(url).header(range).send()?;
+
+        if response.status() != reqwest::StatusCode::PartialContent &&
```
budziq: Id suggest using equivalent boolean expression that might be easier to read
(!a && !b) => !(a || b)

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+use reqwest::header::{ContentRange, ContentRangeSpec, Range};
+
+# error_chain! {
+#     foreign_links {
+#         Io(std::io::Error);
+#         Reqwest(reqwest::Error);
+#     }
+# }
+
+const BUFFER_SIZE: u64 = 100;
```
budziq: the example is no_run anyway. we can go wild with the buffer and request size 
the biggest request handled by httpbin seams to be around 102400 but I did not check precisely. The buffer can be set to something like 10KB to still show partial content (this is still unreasonably small. we might have to add a comment that normally it would be bigger )

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+
+struct PartialRange {
```
budziq: I propose to merge the `PartialRange` and `PartialRangeIter` into one which will be much more concise.

All we will need will be:

- the struct 
```rust
struct PartialRangeIter {
    start: u64,  // content range start
    end: u64,   // content range end
    buffer_size: usize, // external buffer size 
}
```
- The constructor `new` function below
- iter impl which will be much simplified

We should not need anything more complicated than
```rust
impl Iterator for PartialRangeIter {
    type Item = Range;

    fn next(&mut self) -> Option<Self::Item> {
        if self.start >= self.end {
            None
        } else {
            let old_start = self.start;
            self.start += cmp::min(self.buffer_size as u64, self.end - self.start +1);
            Some(Range::bytes(old_start, self.start -1))
        }
    }
}
```

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+impl PartialRange {
+    pub fn new(content_range: &ContentRangeSpec) -> PartialRange {
+        let (content_length, range_starts, range_ends) = match *content_range {
+            ContentRangeSpec::Bytes {
+                range: Some(range),
+                instance_length: Some(content_length),
+            } => (content_length, range.0, range.1),
+            _ => (0, 0, 0),
+        };
+
```
budziq: anything below here upto the `run` function will not be neaded

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+}
+
+impl PartialRange {
+    pub fn new(content_range: &ContentRangeSpec) -> PartialRange {
+        let (content_length, range_starts, range_ends) = match *content_range {
+            ContentRangeSpec::Bytes {
+                range: Some(range),
+                instance_length: Some(content_length),
+            } => (content_length, range.0, range.1),
+            _ => (0, 0, 0),
```
budziq: the iterator with (0,0,0) does not make sense/ would be invalid so I would suggest returning `Err("invalid range speccification")?` in this match leg,

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+const BUFFER_SIZE: u64 = 100;
+
+struct PartialRange {
+    content_length: u64,
+    range_ends: u64,
+    ranges: Vec<u64>,
+}
+
+impl PartialRange {
+    pub fn new(content_range: &ContentRangeSpec) -> PartialRange {
```
budziq: this looks like a failable operation I would suggest returning `Result<PartialRangeIter>` from this function

also we will need the buffer_size as an argument. I would strongly advise against using global constants throughout the code. This leads to bad refactoring experience. 

we will also need to check the buffer_size for being > 0 and return Err if it is not

src/net.md
```diff
@@ -1019,6 +1020,151 @@ fn run() -> Result<()> {
...
+        let (content_length, range_starts, range_ends) = match *content_range {
+            ContentRangeSpec::Bytes {
+                range: Some(range),
+                instance_length: Some(content_length),
```
budziq: how about directly constructing the `Ok(PartialRangeIter{....})` here and leaving the match as the last statement so it is naturally returned from the function. If it will be to dense, we can always return to the parameter extraction in match and returning after it.

src/net.md
jbolila: I have tried that approach during this end up with the following error, and now is back again with this code you're posting here:

```text
error[E0599]: no method named `iter` found for type `PartialRangeIter` in the current scope
  --> src/main.rs:77:25
   |
77 |     for range in ranges.iter() {
   |                         ^^^^
```

here is the complete `struct` + `impl`:

```rust
#[derive(Debug)]
struct PartialRangeIter {
    start: u64, // content range start
    end: u64, // content range end
    buffer_size: usize, // external buffer size
}

impl PartialRangeIter {
    pub fn new(content_range: &ContentRangeSpec, buffer_size: usize) -> Result<PartialRangeIter> {
        if buffer_size == 0 {
            Err("invalid buffer_size, give a value greater than zero.")?;
        }

        let (start, end) = match *content_range {
            ContentRangeSpec::Bytes { range: Some(range), .. } => (range.0, range.1),
            _ => Err("invalid range specification")?,
        };

        Ok(PartialRangeIter {
            start,
            end,
            buffer_size,
        })
    }
}

impl Iterator for PartialRangeIter {
    type Item = Range;

    fn next(&mut self) -> Option<Self::Item> {
        if self.start >= self.end {
            None
        } else {
            let old_start = self.start;
            let inc = std::cmp::min(self.buffer_size as u64, self.end - self.start + 1);
            self.start += std::cmp::min(self.buffer_size as u64, self.end - self.start + 1);
            Some(Range::bytes(old_start, self.start - 1))
        }
    }
}
```

can you explain?

src/net.md
jbolila: I'll do it as you wish, although for me it's not easier to read this way.

```rust
        if !(response.status() == reqwest::StatusCode::PartialContent ||
                 response.status() == reqwest::StatusCode::Ok)
        {
            bail!("Unexpected server response: {}", response.status())
        }
```

src/net.md
budziq: > for range in ranges.iter() {

There is nothing to provide the `iter` method nor is it required.
Som types provide `iter`, `into_iter` or both to return a separate helper iterator type.
In this situation we already have an iterator type instance as it can be constructed with `PartialRangeIter::new` (not being tied to any other type).

So following would simply work 

```rust
for range in ranges {
```

I hope that helpes :+1:

budziq:
Merged, Excellent work @jbolila ! 
I've squashed the commits and I'll post minor cleanup shortly.

jbolila:
Thank you @budziq !! I have learned a few things during this days, thanks for your patience and guidance.

budziq:
No problem! That is exactly what this effort is for!

Issue #268, opened 2017-08-15
Add "Run an external command passing it stdin and check for an error code" example


budziq:
cc #266

See [CONTRIBUTING.md](https://github.com/rust-lang-nursery/rust-cookbook/blob/master/CONTRIBUTING.md) for details.

oldmanmike:
@budziq 

I'd like to start off with this one, if it's okay.

budziq:
Sure @oldmanmike !

budziq:
merged!

PR #310, opened 2017-10-01
Add "Run an external command passing it stdin and check for an error code" example

oldmanmike:
For issue #268

Relatively new to Rust, so feedback would be amazing.


src/basics.md
```diff
@@ -326,6 +327,80 @@ fn run() -> Result<()> {
...
+# }
+
+fn run() -> Result<()> {
+    let mut child = Command::new("python").stdin(Stdio::piped())
+        .stderr(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()?;
+
+    {
+        match child.stdin.as_mut() {
```
budziq: we actually don't need the surrounding braces as long as we do not really need to use the `stdin` more than once. We can just
```rust
 child.stdin.as_mut()?.write_all(b"import this; exit()")
``` 

As for error handling I would normally use `chain_err` or `map_err` if I would like to add more context. But in this example I would try to keep it as simple as possible so just `?` might suffice.

src/basics.md
```diff
@@ -326,6 +327,80 @@ fn run() -> Result<()> {
...
+            Some(stdin) => {
+                stdin.write_all(b"import this; copyright(); credits(); exit()")?;
+            }
+        }
+    }
+
+    let output = child.wait_with_output()?;
+
+    if output.status.success() {
+        match String::from_utf8(output.stdout) {
```
budziq: I would use result's `map` combinator here istead of the match

src/basics.md
```diff
@@ -326,6 +327,80 @@ fn run() -> Result<()> {
...
+for execution. [`Output`] of said statement is then parsed using [`Regex`] and aggregated for all unique words used in output.
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+extern crate regex;
+
+use regex::Regex;
+use std::io::Write;
+use std::ops::Deref;
```
budziq: the deref is not needed here

src/basics.md
```diff
@@ -326,6 +327,80 @@ fn run() -> Result<()> {
...
+                    result.sort();
+                    result.dedup();
+                    println!("Found {} unique words:", result.len());
+                    println!("{:#?}", result);
+                }
+                Ok(())
+            }
+        }
+    } else {
+        match String::from_utf8(output.stderr) {
```
budziq: Instead of this whole match you can just use `bail!` here. It even takes care of formatting

src/basics.md
```diff
@@ -326,6 +327,61 @@ fn run() -> Result<()> {
...
+for execution. [`Output`] of said statement is then parsed.
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+
+use std::collections::HashSet;
+use std::io::Write;
+use std::process::{Command, Stdio};
+
```
budziq: lets add "#" here to avoid needless newline when example is collapsed

src/basics.md
```diff
@@ -326,6 +327,61 @@ fn run() -> Result<()> {
...
+            .collect::<HashSet<_>>();
+        println!("Found {} unique words:", words.len());
+        println!("{:#?}", words);
+        Ok(())
+    } else {
+        let err = String::from_utf8(output.stderr)?;
+        bail!("External command failed:\n {}", err)
+    }
+}
+
```
budziq: lets add "#" here to avoid trailing newline when example is collapsed

budziq:
Thanks @oldmanmike !

PR #308, opened 2017-10-01
Add "Parse a complex version string" example

FaultyRAM:
Fixes #284

PR #315, opened 2017-10-04
Match semver crate examples' styling

mykalu:
At the risk of bikeshedding, I revised the styling I used for #305 to match the styling used in #308.

Probably not even necessary, though. Feelings will not be hurt if the request is closed :')
(should I have opened an issue before PR'ing?)

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -606,26 +606,30 @@ use semver::Version;
 # }
 
 fn run() -> Result<()> {
-    let mut release = Version::parse("0.2.6")?;
-    assert_eq!(release, Version {
-        major: 0,
-        minor: 2,
-        patch: 6,
-        pre: vec!(),
-        build: vec!(),
-    });
+    let mut parsed_version = Version::parse("0.2.6")?;
 
-    release.increment_patch();
-    assert_eq!(release, Version::parse("0.2.7")?);
-    println!("New patch release: v{}", release);
+    assert_eq!(
+        parsed_version,
+        Version {
+            major: 0,
+            minor: 2,
+            patch: 6,
+            pre: vec![],
+            build: vec![],
+        }
+    );
 
-    release.increment_minor();
-    assert_eq!(release, Version::parse("0.3.0")?);
-    println!("New minor release: v{}", release);
+    parsed_version.increment_patch();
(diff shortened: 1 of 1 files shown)


src/app.md
```diff
@@ -606,25 +606,30 @@ use semver::Version;
...
+            minor: 2,
+            patch: 6,
+            pre: vec![],
+            build: vec![],
+        }
+    );
 
     release.increment_patch();
-    assert_eq!(release, Version::parse("0.2.7")?);
+    assert_eq!("0.2.7", release.to_string());
```
budziq: If I were to mark my preference here I would suggest putting the hardcoded string against we are testing on the right.

I probably liked your original version better but it's a matter of taste, so I'll let you chose.

src/app.md
```diff
@@ -606,25 +606,30 @@ use semver::Version;
...
 fn run() -> Result<()> {
-    let mut release = Version::parse("0.2.6")?;
-    assert_eq!(release, Version {
-        major: 0,
-        minor: 2,
-        patch: 6,
-        pre: vec!(),
-        build: vec!(),
-    });
+    let version_str = "0.2.6";
```
budziq: i see no reason for separate `version_str`.

mykalu: Hmmm you're right; I separated it to match the other example, but there's no practical reason for it.

mykalu:
Suggestions implemented!

Hmm, I'm using
`assert_eq!(release.to_string(), "0.3.0");`
but Complex Version is using
`let serialized_version = parsed_version.to_string(); `
`assert_eq!(&serialized_version, version_str);`

I feel that's a styling mismatch.
But to create intermediate variables in Simple Version would be excessive,
and removing the intermediate variable in Complex Version would remove the clarity that we're checking serialization.

Also concerned about the nomenclature difference between using `release` vs `parsed_version` as the variable names.

@FaultyRAM do you have any thoughts on this? :D

budziq:
I would say that clarity+conciseness of given examples are more important tan total stylistic regime. Also I feel that `parsed_version` is more informative than `release`

mykalu:
Suggestions implemented!

budziq:
thanks!

PR #317, opened 2017-10-05
Added "Check if given version is pre-release" example.

### LINKED ISSUE #316: Add "Check if given version is pre-release" example
Use asserts to show that out of two given version strings one is pre-release and second is not.

sb89:
Fixes #316.


budziq:
Nicely done @sb89 !

PR #320, opened 2017-10-07
Add "read csv records"

### LINKED ISSUE #228: add "read csv records"
should show both reading into a StringRecord and typed deserialization with Serde (preferably in two separate short snippets)

cc https://github.com/brson/rust-cookbook/issues/212

ludwigpacifici:
fixes #228

### DIFF
--- a/src/encoding.md
+++ b/src/encoding.md
@@ -11,6 +11,7 @@
 | [Serialize records to CSV][ex-serialize-csv] | [![csv-badge]][csv] | [![cat-encoding-badge]][cat-encoding] |
 | [Serialize records to CSV using Serde][ex-csv-serde] | [![csv-badge]][csv] [![serde-badge]][serde] | [![cat-encoding-badge]][cat-encoding] |
 | [Handle invalid CSV data with Serde][ex-invalid-csv] | [![csv-badge]][csv] [![serde-badge]][serde] | [![cat-encoding-badge]][cat-encoding] |
+| [Read CSV records][ex-csv-read] | [![csv-badge]][csv] | [![cat-encoding-badge]][cat-encoding] |
 | [Read CSV records with different delimeter][ex-csv-delimiter] | [![csv-badge]][csv] | [![cat-encoding-badge]][cat-encoding] |
 | [Filter CSV records matching a predicate][ex-csv-filter] | [![csv-badge]][csv] | [![cat-encoding-badge]][cat-encoding] |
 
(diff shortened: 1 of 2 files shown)


src/encoding.md
```diff
@@ -491,6 +492,84 @@ alisha,colombo,xyz";
...
+    for record in reader.deserialize() {
+        let record: Record = record?;
+        println!("{:?}", record);
+    }
+
+    Ok(())
+}
+
+#[derive(Debug, Deserialize)]
+struct Record {
```
budziq: I would move the struct definition to the beginning of the snippet. this will improve the reading flow somewhat

src/encoding.md
```diff
@@ -491,6 +492,84 @@ alisha,colombo,xyz";
 # quick_main!(run);
 ```
 
+[ex-csv-read]: #ex-csv-read
+<a name="ex-csv-read"></a>
+## Read CSV records
+
+[![csv-badge]][csv] [![cat-encoding-badge]][cat-encoding]
+
+Reads standard CSV records into [`csv::StringRecord`] (valid UTF-8 bytes).
```
budziq: > (valid UTF-8 bytes).

the meaning of this last part is unclear. Could you elaborate the description?

src/encoding.md
```diff
@@ -491,6 +492,84 @@ alisha,colombo,xyz";
...
+fn run() -> Result<()> {
+    let csv = "year,make,model,description
+1948,Porsche,356,Luxury sports car
+1967,Ford,mustang fastback 1967,American car";
+
+    let mut reader = csv::Reader::from_reader(csv.as_bytes());
+
+    for record in reader.deserialize() {
+        let record: Record = record?;
+        println!("{:?}", record);
```
budziq: how about we emphasize the strongly typed nature of the deserialization by also using one of the fields directly?

src/encoding.md
```diff
@@ -491,6 +492,84 @@ alisha,colombo,xyz";
...
+# }
+
+fn run() -> Result<()> {
+    let csv = "year,make,model,description
+1948,Porsche,356,Luxury sports car
+1967,Ford,mustang fastback 1967,American car";
+
+    let mut reader = csv::Reader::from_reader(csv.as_bytes());
+    for record in reader.records() {
+        println!("{:?}", record?);
```
budziq: how about we emphesize the dynamic nature of this variant by using the record fields by its index?

src/encoding.md
ludwigpacifici: From the documentation: https://docs.rs/csv/1.0.0-beta.4/csv/struct.StringRecord.html

> A string record permits reading or writing CSV rows that are valid UTF-8. [...] If you do need to read possibly invalid UTF-8 data, then you should prefer using a ByteRecord, since it makes no assumptions about UTF-8.

Is it part of the cookbook to paraphrase the rust/crates documentation? I think it is not a good idea to duplicate it, and actually I prefer to remove:

> (valid UTF-8 bytes).

The reader can follow the link for more precise and up to date documentation.

src/encoding.md
ludwigpacifici: Yes, good idea

src/encoding.md
ludwigpacifici: Good idea

src/encoding.md
budziq: Hmm it might be quite surprising to the users that their valid csv will not work with this example due to uncommon encoding. As this is the most basic csv example it should be the most informative and ordered as first (we will reorder the deserialization examples later on). I would actually suggest to mentioning this distinction between StringRecord and ByteRecord.

ludwigpacifici: Make sense. Should it be mentioned in the description and in an additional snippet?

budziq: the end of the main description should be enough but I'll leave it up to you to put it where it will make mist sense

ludwigpacifici: I will mention it in the description. However, the extra snippet is not useful. In the first example code, it's only a matter of replacing from `for record in reader.records() {` to `for record in reader.byte_records() {`

budziq: I wouldn't say that the second snippet is not useful. But handling non UTF-8 data there would be more involved so lets leave it as its is.

budziq:
@ludwigpacifici Nicely done!
I'll just remove the derive debug as it is no longer needed when cleaning up/reordering the csv examples

ludwigpacifici:
It's a pleasure! Thanks for your reviews @budziq

PR #323, opened 2017-10-09
Add 'Compile a C library while setting custom defines' (#295)

### LINKED ISSUE #295: add "Compile a C library while setting custom defines" example
Use [cc](https://crates.io/crates/cc) crate to compile a C library from a build.rs file while setting custom defines/CFLAGS.

Please note that it will be important to mention that this is a **build time** crate.
Most likely we will dedicate a separate (last) chapter like "build_time.md"

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/236

bobbo:
Fixes #295

### DIFF
--- a/src/build_tools.md
+++ b/src/build_tools.md
@@ -177,11 +177,11 @@ fn main(){
 
 [![cc-badge]][cc] [![cat-development-tools-badge]][cat-development-tools]
 
-It is simple to build bundled C code with custom defines using [`cc::Build::define`][cc-build-define]. 
-It takes an [`Option`] value, so it is possible to create defines such as `#define FLAG`
-as well as `#define VALUE 123`. This example builds a bundled C file with dynamic defines set in 
-`build.rs` and prints `Welcome to foo - version 1.0.2` when run. Cargo sets some [environment 
-variables][cargo-env] which may be useful for some custom defines.
+It is simple to build bundled C code with custom defines using [`cc::Build::define`]. 
+It takes an [`Option`] value, so it is possible to create defines such as `#define APP_NAME "foo"`
+as well as `#define WELCOME` (pass `None` as the value for a value-less defne). This example builds
+a bundled C file with dynamic defines set in `build.rs` and prints "**Welcome to foo - version 1.0.2**"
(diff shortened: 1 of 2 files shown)


src/build_tools.md
```diff
@@ -171,14 +171,84 @@ fn main(){
...
+```
+
 {{#include links.md}}
 
 <!-- Other Reference -->
 
 [build-script-docs]: http://doc.crates.io/build-script.html
 [playground]: https://play.rust-lang.org
 [cc-build]: https://docs.rs/cc/*/cc/struct.Build.html
+[cc-build-define]: https://docs.rs/cc/1.0.0/cc/struct.Build.html#method.define
```
budziq: please use wildcard version numbers
also we can use
<code>
[`Build::define`]: https://docs.rs/cc/*/cc/struct.Build.html#method.define
</code>
directly instead. 

It would be great help if you could cleanup the other examples here with this type of reference (but this will not be a blocker for this PR ;) )

src/build_tools.md
```diff
@@ -171,14 +171,84 @@ fn main(){
...
+[ex-cc-custom-defines]: #ex-cc-custom-defines
+<a name="ex-cc-custom-defines"></a>
+## Compile a C library while setting custom defines
+
+[![cc-badge]][cc] [![cat-development-tools-badge]][cat-development-tools]
+
+It is simple to build bundled C code with custom defines using [`cc::Build::define`][cc-build-define]. 
+It takes an [`Option`] value, so it is possible to create defines such as `#define FLAG`
+as well as `#define VALUE 123`. This example builds a bundled C file with dynamic defines set in 
+`build.rs` and prints `Welcome to foo - version 1.0.2` when run. Cargo sets some [environment
```
budziq: > `build.rs` and prints `Welcome to foo - version 1.0.2` 

Lets leave the backtick "`" emphasis for code. I would suggest **bold** emphasis here.

bobbo:
Hi @budziq, updated based on your feedback. How's it looking now?

budziq:
Thanks @bobbo!

Issue #216, opened 2017-07-04
add "Obtain backtrace of complex error scenarios" example


budziq:
Show some realistic 3 level error handling (possibly based on largely stripped down version of [Download a file example](https://brson.github.io/rust-cookbook/net.html#download-a-file-to-a-temporary-directory) without tempdir/headder noise)

Use the error-chain crate

cc https://github.com/brson/rust-cookbook/issues/184

ludwigpacifici:
Hello @budziq 

I would be interested to provide an example for this topic.

budziq:
Thanks @ludwigpacifici . the link was for the previous landing page sry.

ludwigpacifici:
To make sure I am heading in the right direction, can you tell me if the snippet highlights a backtrace with 3 levels of error handling?

```rust
fn run() -> Result<()> {
    f1().chain_err(|| "run failed beacuse f1")?;
    Ok(())
}

fn f1() -> Result<std::fs::File> {
    f2().chain_err(|| "f1 failed because of f2")
}

fn f2() -> Result<std::fs::File> {
    File::open("oops").chain_err(|| "f2 is unable to open 'oops' file")
}
```

Ouput:

```text
Error: run failed beacuse f1
Caused by: f1 failed because of f2
Caused by: f2 is unable to open 'oops' file
Caused by: No such file or directory (os error 2)
```

budziq:
@ludwigpacifici pretty much yeah but I would like for this example to have very simple but at least semi realistic case and a little nicer function names. Would you be able to come up with and idea for it? I'm thinking of some high level operation failing because obtaining a token (parsing possibly an int or just utf8 string) from file failed due to open error.

ludwigpacifici:
Write an implementation of [ROT13](https://en.wikipedia.org/wiki/ROT13). It takes a file as input, and print the encoded text on the standard output. To keep the code short, it will switch [a-zA-Z] and other valid characters are skipped. At first glance, it can fails if the file cannot be opened or a string from the file is not utf8.

budziq:
I would suggest against ROT13 I would hesitate advertising this manner of ~~encryption~~/obfuscation even in joking manner. But any other failable mode of text processing would do such as matching file contents against a regexp?

ludwigpacifici:
Then, what do you think of a basic calculator: `<number><op><number>`, where `<number>` is a decimal integer and `<op>` is one of `+`, `-`, `/` and `*`?

```
[...]

PR #325, opened 2017-10-09
add "Obtain backtrace of complex error scenarios" example

ludwigpacifici:
Issue #216

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -24,7 +24,7 @@
 | [Define and operate on a type represented as a bitfield][ex-bitflags] | [![bitflags-badge]][bitflags] | [![cat-no-std-badge]][cat-no-std] |
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Check number of logical cpu cores][ex-check-cpu-cores] | [![num_cpus-badge]][num_cpus] | [![cat-hardware-support-badge]][cat-hardware-support] |
-
+| [Obtain backtrace of complex error scenarios][ex-error-chain-backtrace] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
@@ -866,7 +866,7 @@ Uses [`ring::hmac`] to creates a [`hmac::Signature`] of a string then verifies t
 # #[macro_use]
 # extern crate error_chain;
 extern crate ring;
-# 
+#
 # error_chain! {
 #     foreign_links {
 #         Ring(ring::error::Unspecified);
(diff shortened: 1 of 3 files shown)


ludwigpacifici:
Hello @budziq 

Here is my attempt to show a "complex" backtrace with a relatively short code. Let me know if it is a good candidate for this issue.

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
...
+102,256,204";
+
+    let rgb = read_rgb_from_csv(csv.as_bytes()).chain_err(|| "Cannot read CSV data")?;
+    let hex = rgb_to_hex(&rgb);
+    println!("{:?} to hexadecimal #{:X}", rgb, hex);
+
+    Ok(())
+}
+
+quick_main!(run);
```
budziq: lets not use the `quick_main!` here as it will error out the application.
lets have a regular `main` in addition to `run` to showcase usage of
[`backtrace`](https://docs.rs/error-chain/0.11.0/error_chain/trait.ChainedError.html#tymethod.backtrace) in order to iterate over the backtrace

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
...
+    let color: Rgb = csv::Reader::from_reader(csv_data)
+        .deserialize()
+        .nth(0)
+        .ok_or("Cannot deserialize the first CSV record")?
+        .chain_err(|| "Cannot deserialize RGB color")?;
+
+    Ok(color)
+}
+
+# type Hex = u32;
```
budziq: we can substitute this fun tion with custom `Display` impl for `Rgb`

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
...
+#
+# fn rgb_to_hex(color: &Rgb) -> Hex {
+#     (color.red as Hex) << 16 | (color.blue as Hex) << 8 | (color.green as Hex)
+# }
+#
+fn run() -> Result<()> {
+    let csv = "red,blue,green
```
budziq: to bump the size of our backtrace we can use a file input. Provide an associate function `from_path` which would in turn call associated function `from_reader` (formerly `read_rgb_from_csv`)

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
 }
 ```
 
+[ex-error-chain-backtrace]: #ex-error-chain-backtrace
+<a name="ex-error-chain-backtrace"></a>
+## Obtain backtrace of complex error scenarios
+
+[![error-chain-badge]][error-chain] [![cat-rust-patterns-badge]][cat-rust-patterns]
+
+Extend errors by appending new errors via [`chain_err`]. It provides a
```
budziq: I would start with mentioning that we have a complex error handling scenario and we present a backtrace from it.

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
...
+better context to understand an error in a form of a backtrace. The
+below recipes attemps to deserialize the value `256` into a `u8`. An
```
budziq: Also I would present the example backtrace output after the code snippet.

src/basics.md
ludwigpacifici: I don't  understand you comment. Can you give more details?

ludwigpacifici:
Thanks @budziq for your review. I am on a weekend, I expect to correct the PR beginning of the week.

src/basics.md
ludwigpacifici: > custom Display impl for Rgb
 
Ah, I think I get it :-) Do you mean something like:

```rust
let color = Rgb{...};
println!("#{:X}", color); // Prints #66FFCC
```

`rgb_to_hex` will be replace by the `Display` impl for `Rgb`?

src/basics.md
ludwigpacifici: Isn't it redundant if the reader hit the play button?

src/basics.md
budziq: Good point :+1:

src/basics.md
budziq: We will be only able to (easily) simulate one error scenario (file not found). We might decide to show backtraces from other error scenarios too. But we will see how the example unfolds :)

src/basics.md
```diff
@@ -1101,6 +1101,68 @@ fn main() {
...
+    blue: u8,
+    green: u8,
+}
+
+fn read_rgb_from_csv(csv_data: &[u8]) -> Result<Rgb> {
+    let color: Rgb = csv::Reader::from_reader(csv_data)
+        .deserialize()
+        .nth(0)
+        .ok_or("Cannot deserialize the first CSV record")?
+        .chain_err(|| "Cannot deserialize RGB color")?;
```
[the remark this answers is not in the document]
ludwigpacifici: I have to use `ok_or` and `chain_err` to make it work. I tried to be concise, maybe too much? Here is a more verbose version:

```rust
let mut reader = csv::Reader::from_reader(csv_data);
let mut it = reader.deserialize();
let option = it.nth(0);
if let Some(inner) = option {
    // Notice that we need to provide a type hint for automatic deserialization.
    let color: Rgb = inner.chain_err(|| "Cannot deserialize RGB color")?;
    Ok(color)
} else {
    Err(From::from("Cannot deserialize the first CSV record"))
}
```

Documentation I used: https://docs.rs/csv/1.0.0-beta.4/csv/index.html#example-with-serde

src/basics.md
ludwigpacifici: Not sure I understand this comment. [`from_path`](https://docs.rs/csv/1.0.0-beta.4/csv/struct.ReaderBuilder.html#method.from_path) is already provided by the csv crate. 

The snippet would look like:

```rust
    let color: Rgb = csv::ReaderBuilder::new()
        .from_path("color.csv")
        .chain_err(|| "Cannot read CSV file")?
        .deserialize()
        .nth(0)
        .ok_or("Cannot deserialize the first CSV record")?
        .chain_err(|| "Cannot deserialize RGB color")?;
```

Then it leads to a shorter error stack:

```text
Error level - description
└> 0 - Cannot read CSV file
└> 1 - No such file or directory (os error 2)
└> 2 - No such file or directory (os error 2)
```

Am I missing something?

ludwigpacifici:
I updated the PR following most of your comments. Still open to discussion:
* usage of `ok_or` and `chain_err`
* usage of `from_path`

src/basics.md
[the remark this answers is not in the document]
budziq: Nope, the original is very good. Honestly my bad I have made this review a too cursory one.

src/basics.md
budziq: Again my bad for lack of diligence in the review.

src/basics.md
```diff
@@ -1101,14 +1101,109 @@ fn main() {
...
+        errors
+            .iter()
+            .enumerate()
+            .for_each(|(index, error)| eprintln!("└> {} - {}", index, error));
+
+        if let Some(backtrace) = errors.backtrace() {
+            eprintln!("{:?}", backtrace);
+        }
+
+        ::std::process::exit(1);
```
budziq: I'd just drop this line to make our CI happy or make the example `no_run`. We can always write a comment here mentioning that we would normally handle the errors or leave the app here with and error.

src/basics.md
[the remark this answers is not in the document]
ludwigpacifici: No worries @budziq :-) It's good you question my rust code because I am beginner and cookbook PR teaches me a lot.

src/basics.md
ludwigpacifici: That's alright. Maybe the following snippet will make the example more Rust idiomatic?

```rust
impl Rgb {
    fn from_reader(csv_data: &[u8]) -> Result<Rgb> {
        // move code of read_rgb_from_csv here
    }
}
```

and in the `run` function, the line will be:

```
let rgb = Rgb::from_reader(csv.as_bytes()).chain_err(|| "Cannot read CSV data")?;
```

The error stack is unchanged. What do you think?

src/basics.md
ludwigpacifici: The CI is trying to execute the recipes by default, meaning it checks the code compiles at least. If `no_run` is set, is the code still compiled?

I try to understand which is better `no_run` vs make the code return a success code.

budziq: `no_run` - compiles the code but does not run it
`ignore` - does not compile it at all

I'd probably go with leaving this example runnable but removing the `::std::process::exit(1);` but I'll leave it up to you.

ludwigpacifici: Thanks for the listing.

> I'd probably go with leaving this example runnable but removing the ::std::process::exit(1);

Yes, I'll do that

ludwigpacifici:
CI is happy :-)

One question left on my side: https://github.com/rust-lang-nursery/rust-cookbook/pull/325#discussion_r146101680, and waiting for your suggestions, if you have any.

src/basics.md
budziq: yep that's cool 👍 the comment slipped past me

ludwigpacifici:
`read_rgb_from_csv` was moved to `from_reader`. 

I noticed removing `process::exit(1)` make the CI happy, but when I click on the play button, the program output is empty (error backtrace is not displayed, but it is rendered as text just below the code snippet). Looks like the error output is redirected somewhere else. Should we keep this behavior or add a `no_run` instead?

budziq:
Stderr is printed only if playground returns [process failure](https://github.com/azerupi/mdBook/blob/cd711bfb1ca161e8a834f4527f3936f818a6f50f/src/theme/book.js#L386)

Let's keep the example runnable and possibly submit a bug report to mdbook (we can also modify it lolly in cookbook but I'd prefer to to avoid it)

ludwigpacifici:
Alright @budziq. In that case, the PR looks fine on my side. I will raise an issue to mdbook. Should I cc you?

budziq:
Thanks @ludwigpacifici!

Yep, please cc me on mdbook issue. I'll ad it to my ever expanding todo list ;).

budziq:
or better yet please create the issue in cookbook for now untill we come up with reasonable solution to propose.

ludwigpacifici:
Thanks for your time @budziq 

I will put it on the ever expanding list of the rust-cookbook :-)

Issue #93, opened 2017-05-17
Handle a rate-limited API


dtolnay:
Hitting a rate limit may either return an error for the request or return a JSON body that is different from the usual. Let's show how to deal with one or both of these cases.

Michael-F-Bryan:
What kind of example did you have in mind here? The way rate limiting is handled will necessarily vary depending on which API you are communicating with. For example, GitHub always replies with a `X-RateLimit-Remaining` header and you'll get a 403 if you go over your limit.

So a hypothetical request to some GitHub API might go like this:

- Send request
- deserialize response
- if deserializing fails, check header for `X-RateLimit-Remaining` and return `RateLimitError` if equal to "0"
- else check status code and do other error handling to return an appropriate error

Or were you thinking of patterns people can use to make sure they don't go over their rate limit (e.g. the [Leaky Bucket] algorithm)?

[Leaky Bucket]: https://en.wikipedia.org/wiki/Leaky_bucket

dtolnay:
@konstin you suggested this one initially. Any thoughts?

konstin:
The correct solution would be a crate that is hooked into hyper and/or rewquest. This crate would delay all requests until you're allowed to requests again while logging to the user. Unfortunately, no stable crate for that currently exists.

There a two custom solutions: Github's api is friendly enough telling you the number of remaining requests, so you'd just check that header and sleep if the number gets to low.

As most apis aren't that helpfull though, the best solution is essentially trial and error, which I've quickly sketched out below. Consider the code as pseudocode, it's only meant to explain the control flow. 

```rust
fn rate_limit_agnostic_request(url: Url) -> Result<String, Box<Error>> {
    loop {
        let mut reponse = reqwest::get(&url)?;
        if reponse.status().is_success() {  
            let mut string = String::new();
            reponse.read_to_string(&mut string)?;
            return Ok(string);
        } else if reponse.status() == TooManyRequests {
            info!("Rate limit hit, sleeping for 5 seconds");
            sleep(Seconds::new(5)); // TODO: Find the right functions for that
        } else { // Some other error
            return Err(from::From("Some other error"));
        }
    }
}
```

konstin:
[...]

PR #333, opened 2017-10-14
Adds example of handle a rate-limited API #93

jbolila:
a small example.

### DIFF
--- a/src/intro.md
+++ b/src/intro.md
@@ -94,6 +94,7 @@ community. It needs and welcomes help. For details see
 | [Check webpage for broken links][ex-check-broken-links] | [![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Extract all unique links from a MediaWiki markup][ex-extract-mediawiki-links] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
 | [Make a partial download with HTTP range headers][ex-progress-with-range] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
+| [Handling the Github rate limit error condition][ex-rate-limit-exceeded] | [![reqwest-badge]][reqwest] [![hyper-badge]][hyper] | [![cat-net-badge]][cat-net] |
 
 ## [Application development](app.html)
 
@@ -225,6 +226,7 @@ community. It needs and welcomes help. For details see
 [ex-url-rm-frag]: net.html#ex-url-rm-frag
 [ex-urlencoded]: encoding.html#ex-urlencoded
 [ex-progress-with-range]: net.html#ex-progress-with-range
+[ex-rate-limit-exceeded]: net.html#ex-rate-limit-exceeded
 [ex-random-file-access]: basics.html#ex-random-file-access
 [ex-regex-replace-named]: basics.html#ex-regex-replace-named
 [ex-csv-filter]: encoding.html#ex-csv-filter
--- a/src/net.md
+++ b/src/net.md
@@ -20,6 +20,7 @@
 | [Check webpage for broken links][ex-check-broken-links] | [![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Extract all unique links from a MediaWiki markup][ex-extract-mediawiki-links] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
 | [Make a partial download with HTTP range headers][ex-progress-with-range] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
+| [Handling the Github rate limit error condition][ex-rate-limit-exceeded] | [![reqwest-badge]][reqwest] [![hyper-badge]][hyper] | [![cat-net-badge]][cat-net] |
 
 [ex-url-parse]: #ex-url-parse
 <a name="ex-url-parse"/>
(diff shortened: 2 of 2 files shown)


src/intro.md
```diff
@@ -94,6 +94,7 @@ community. It needs and welcomes help. For details see
 | [Check webpage for broken links][ex-check-broken-links] | [![reqwest-badge]][reqwest] [![select-badge]][select] [![url-badge]][url] | [![cat-net-badge]][cat-net] |
 | [Extract all unique links from a MediaWiki markup][ex-extract-mediawiki-links] | [![reqwest-badge]][reqwest] [![regex-badge]][regex] | [![cat-net-badge]][cat-net] |
 | [Make a partial download with HTTP range headers][ex-progress-with-range] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
+| [Handling the Github rate limit error condition][ex-rate-limit-exceeded] | [![reqwest-badge]][reqwest] [![hyper-badge]][hyper] | [![cat-net-badge]][cat-net] |
```
j-haj: I think the title should be more general to the problem we are solving, rather than Github specific. For example "Handle a rate-limited API"

Similarly, `ex-handle-rate-limited-api` may be more clear (although arguably we are talking about the same thing)

src/net.md
```diff
@@ -1123,6 +1124,99 @@ fn run() -> Result<()> {
...
+This example uses the [GitHub API - Rate limiting], as an example of how to handle error conditions.
+
+
+```rust,no_run
+#[macro_use]
+extern crate error_chain;
+#[macro_use]
+extern crate hyper;
+extern crate reqwest;
+# extern crate rayon;
```
j-haj: You don't seem to be using the `rayon` crate in this example, so I would remove references to `rayon`.

src/net.md
```diff
@@ -1123,6 +1124,99 @@ fn run() -> Result<()> {
...
+
+use std::time::{Duration, UNIX_EPOCH};
+
+use reqwest::StatusCode;
+# use rayon::prelude::*;
```
j-haj: Again, this should be removed

src/net.md
```diff
@@ -1123,6 +1124,99 @@ fn run() -> Result<()> {
...
 ```
 
+[ex-rate-limit-exceeded]: #ex-rate-limit-exceeded
+<a name="ex-rate-limit-exceeded"/>
+## Handling the Github rate limit error condition
+
+[![reqwest-badge]][reqwest] [![hyper-badge]][hyper] [![cat-net-badge]][cat-net]
+
+This example uses the [GitHub API - Rate limiting], as an example of how to handle error conditions.
+
```
j-haj: I think it might be helpful to elaborate a little bit on what is going on in this example. It doesn't need to be an essay, but maybe just briefly spelling out the steps or any key points the reader should be aware of.

j-haj:
One additional item I wanted to mention is I get an error when running this locally:

```
Compiling playground v0.0.1 (file:///playground)
    Finished dev [unoptimized + debuginfo] target(s) in 4.66 secs
     Running `target/debug/playground`
Error: https://api.github.com/users/rust-lang-nursery: failed to lookup address information: Temporary failure in name resolution
```

The URL appears to be correct, so I'm not entirely sure what the issue is here... It is quite possible it is user error on my end (maybe due to running the cookbook locally?).

src/net.md
jbolila: is using Rayon to do calls in parallel and reach the limit of API calls, that block of code is hidden as is this one.

j-haj: Ah yes I missed that `into_par_iter` on line 1179... my apologies!

jbolila:
@j-haj thank you for your comments. The url seems to be correct https://developer.github.com/v3/users/

j-haj:
@jbolila yes the URL is definitely correct - I got a valid response if I tested it out independently, but for some reason when I was running in mdbook I got that error. I suspect it has more to do with either my environment or mdbook

src/net.md
```diff
@@ -1123,6 +1124,102 @@ fn run() -> Result<()> {
...
+    let rate_reset_within = Duration::from_secs(**rate_reset_at) - UNIX_EPOCH.elapsed()?;
+
+    if response.status() == StatusCode::Forbidden && **rate_remaining == 0 {
+        bail!(ErrorKind::RateLimitExceded(
+            **rate_limit,
+            rate_reset_within.as_secs(),
+        ));
+    }
+
+    println!(
```
budziq: I'd suggest we try to handle the ratelimit (possibly wait and retry?) here instead of just printing and bailing.

I know that some other contributors had opinions on this slightly unclear example idea.
@konstin and @clarcharr  would like to add some thoughts on this example?

src/net.md
```diff
@@ -1123,6 +1124,102 @@ fn run() -> Result<()> {
...
+[![reqwest-badge]][reqwest] [![hyper-badge]][hyper] [![cat-net-badge]][cat-net]
+
+This example uses the [GitHub API - Rate limiting], as an example of how to handle error conditions.
+
+For that defines three custom headers using the [`hyper::header!`] macro to read the values sent by
+the server, and checks for [`reqwest::StatusCode::Forbidden`]. If the rate limit was excided raises
+an exception with a detailed message containing the information provided in the custom headers.
+
+```rust,no_run
+#[macro_use]
```
budziq: I'd suggest to hide all of the error_chain related code

src/net.md
```diff
@@ -1123,6 +1124,102 @@ fn run() -> Result<()> {
...
+header! { (XRateLimitRemaining, "X-RateLimit-Remaining") => [usize] }
+header! { (XRateLimitReset, "X-RateLimit-Reset") => [u64] }
+
+fn run() -> Result<()> {
+    let url = "https://api.github.com/users/rust-lang-nursery ";
+    let client = reqwest::Client::new();
+
+#   let response = client.get(url).send()?;
+#
+#   // for the purpose of this example will exaust the limit off calls
```
budziq: lets drop the limit exhaustion code from the example

src/net.md
jbolila: Was important to develop and test this code sample, and it's now up to you how do you want to present this.

src/net.md
jbolila: I'll wait a few days to get more clear directions what do you wanna have in the cookbook, thank you.

budziq: @jbolila looks like there are no suggestions ;) So lets go with the simplest possible implementation (wait and retry).

budziq:
Ping @jbolila . Do you need any help in getting this through the finish line?

jbolila:
@budziq if you can do the changes you have in mind will be great. Thank you.

budziq:
@jbolila thanks for the info. probably will wait for someone to take this PR up. Most likely it would be best to implement this using async hyper once cookbook async chapter is good to go

PR #334, opened 2017-10-15
Update examples with "for_each"

### LINKED ISSUE #332: modernize examples for rust 1.21 (`Iterator::for_each`)
Our examples have a lot of for loops that operate on chained iterators just to consume the values for their side effects (typically a single `println!`).

As the [`Iterator::for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each) has been stabilized in rust 1.21 we can make the examples a little more idiomatic. Lets use `for_each` instead of for loop when we have chained iterator calls.

sb89:
Fixes #332

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -309,8 +309,7 @@ fn run() -> Result<()> {
                                ([0-9a-fA-F]+) # commit hash
                                (.*)           # The commit message")?;
 
-    let stdout = String::from_utf8(output.stdout)?;
-    let commits = stdout
+    String::from_utf8(output.stdout)?
         .lines()
         .filter_map(|line| pattern.captures(line))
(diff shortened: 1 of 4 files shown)


src/basics.md
```diff
@@ -310,7 +310,7 @@ fn run() -> Result<()> {
                                (.*)           # The commit message")?;
 
     let stdout = String::from_utf8(output.stdout)?;
-    let commits = stdout
+    stdout
```
budziq: we can get rid of the intermediate temp variables such as `stdout` here

sb89:
Implemented changes as requested.

budziq:
Well done!

Issue #331, opened 2017-10-12
Add "Test if all/any items match predicate in parallel" example


budziq:
Showcase a [`any`](https://docs.rs/rayon/*/rayon/iter/trait.ParallelIterator.html#method.any) and [`all`](https://docs.rs/rayon/*/rayon/iter/trait.ParallelIterator.html#method.all) with [rayon](https://crates.io/crates/rayon) crate.

Please discuss the idea here before starting implementation.
cc https://github.com/rust-lang-nursery/rust-cookbook/issues/272

CJStadler:
I'm interested in working on this. Should this be one or two examples? Do we want something more complicated than "check whether all the elements in this array are even"? These functions are so basic that it feels like adding additional complexity might just obscure their use without adding value.

budziq:
I would go with the simplest approach possible (also mentioning the sequential versions from std in the description). We can showcase both `all` and `any` in the same snippet.

PR #335, opened 2017-10-15
Add rayon example with any and all

CJStadler:
fixes https://github.com/rust-lang-nursery/rust-cookbook/issues/331

edit: added `fixes #311` clause to description to close the issue on merge.

### DIFF
--- a/src/concurrency.md
+++ b/src/concurrency.md
@@ -54,11 +54,15 @@ fn main() {
 
     assert!(!vec.par_iter().any(|n| (*n % 2) != 0)); // None are odd.
     assert!(vec.par_iter().all(|n| (*n % 2) == 0)); // All are even.
+    assert!(!vec.par_iter().any(|n| *n > 8 )); // None are greater than 8.
+    assert!(vec.par_iter().all(|n| *n <= 8 )); // All are less than or equal to 8.
 
-    vec.push(3);
+    vec.push(9);
 
     assert!(vec.par_iter().any(|n| (*n % 2) != 0)); // At least 1 is odd.
     assert!(!vec.par_iter().all(|n| (*n % 2) == 0)); // Not all are even.
+    assert!(vec.par_iter().any(|n| *n > 8 )); // At least 1 is greater than 8.
+    assert!(!vec.par_iter().all(|n| *n <= 8 )); // Not all are less than or equal to 8.
 }
 ```
 
--- a/src/concurrency.md
+++ b/src/concurrency.md
@@ -3,6 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Mutate the elements of an array in parallel][ex-rayon-iter-mut] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
+| [Check the elements of an array against a predicate in parallel][ex-rayon-any-all] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Search items using given predicate in parallel][ex-rayon-parallel-search] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Sort a vector in parallel][ex-rayon-parallel-sort] | [![rayon-badge]][rayon] [![rand-badge]][rand] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Map-reduce in parallel][ex-rayon-map-reduce] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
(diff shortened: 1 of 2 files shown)


src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
 }
 ```
 
+[ex-rayon-any-all]: #ex-rayon-any-all
+<a name="ex-rayon-any-all"></a>
+## Check the elements of an array against a predicate in parallel
+
+[![rayon-badge]][rayon] [![cat-concurrency-badge]][cat-concurrency]
+
+The `rayon` provides a [`ParallelIterator`] trait which includes parallelized implementations of many of the methods from the standard library's [`Iterator`] trait.
```
j-haj: Missing "crate" in "The `rayon` [crate] provides..."

src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
...
+
+This example demonstrates using the [`any`] and [`all`] methods. [`any`] checks in parallel whether any element of the iterator matches the predicate, and returns as soon as one is found. [`all`] checks in parallel whether all elements of the iterator match the predicate, and returns as soon as a non-matching element is found.
```
j-haj: I think we should use `rayon::any` and `rayon::all` here, since there are also functions `any` and `all` in the standard library. The concern isn't so much clarity (I think this example is very well written) but more so about namespacing the function names in our links.

src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
...
+
+```rust
+extern crate rayon;
+
+use rayon::prelude::*;
+
+fn main() {
```
j-haj: There seems to be a mix of examples using a `main()` function and a those using `error_chain`. I'm not sure the Contributing guide is clear on this either, as it does not explicitly state contributions should use `error_chain`. On the other hand, there are no errors we need to worry about here, so maybe using `error_chain` unnecessarily complicates the examples.

@budziq what are your thoughts here? I am somewhat partial to using a consistent approach across examples.

j-haj:
Outstanding example! I think the description is especially well written. Just a minor fix on a missing word.

src/concurrency.md
budziq: @j-haj Thanks for taking this on :+1: . 
You are right our [note about error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling) does not specifically state what is the correct course of action if there are not errors to handle. 

The approach is as follows. If there are no errors to handle, there is no reason to increase the complexity of the example with `run` and `error_chain` and we just use plain `main`. We might be interested in adding this information to our contributing and error handling sections

CJStadler: I was also wondering about this. Thanks!

CJStadler:
Thanks for the review @j-haj!

src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
 }
 ```
 
+[ex-rayon-any-all]: #ex-rayon-any-all
+<a name="ex-rayon-any-all"></a>
+## Check the elements of an array against a predicate in parallel
+
+[![rayon-badge]][rayon] [![cat-concurrency-badge]][cat-concurrency]
+
+The `rayon` crate provides a [`ParallelIterator`] trait which includes parallelized implementations of many of the methods from the standard library's [`Iterator`] trait.
```
budziq: I'd suggest to omit this first paragraph as the reader should already be familiar with rayon basics via the previous examples. On the other hand we might want to mention that `rayon::any` and `rayon::all` are parallel versions of their std counterparts.

src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
...
+
+```rust
+extern crate rayon;
+
+use rayon::prelude::*;
+
+fn main() {
+    let mut vec = vec![2, 4, 6, 8];
+
+    let any_odd = vec.par_iter().any(|n| (*n % 2) != 0);
```
budziq: I'd suggest to write these just as a single line assertions
```rust
assert!(!vec.par_iter().any(|n| (*n % 2) != 0));
```
 this would be slightly more concise

src/concurrency.md
```diff
@@ -33,6 +34,40 @@ fn main() {
...
+    assert!(!any_odd);
+
+    let all_even = vec.par_iter().all(|n| (*n % 2) == 0);
+    assert!(all_even);
+
```
budziq: how about we add a little flavor to the second predicate like some inequality?
```rust
 assert!(vec.par_iter().all(|n| *n < 10));
  vec.push(13);
```

src/concurrency.md
```diff
@@ -3,6 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Mutate the elements of an array in parallel][ex-rayon-iter-mut] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
+| [Check the elements of an array against a predicate in parallel][ex-rayon-any-all] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
```
budziq: How about renaming the example to simply:
"Test in parallel if all or any collection elements match given predicate"

We are not limited to arrays here (and actually any collection/iterator viable for `par_Iter` will do)

CJStadler: Definitely 👍

src/concurrency.md
CJStadler: It seems like this loses some of the power of the example to demonstrate the relationship between the methods (`when not any are odd then all are even, when any are odd then not all are even`).

src/concurrency.md
CJStadler: 👍 I can mention the std counterparts in the second paragraph.

src/concurrency.md
CJStadler: Personally I think it's slightly more readable to break up the line and give this a name, but I'm happy to change it.

src/concurrency.md
```diff
@@ -33,6 +34,32 @@ fn main() {
...
+fn main() {
+    let mut vec = vec![2, 4, 6, 8];
+
+    assert!(!vec.par_iter().any(|n| (*n % 2) != 0)); // None are odd.
+    assert!(vec.par_iter().all(|n| (*n % 2) == 0)); // All are even.
+
+    vec.push(3);
+
+    assert!(vec.par_iter().any(|n| (*n % 2) != 0)); // At least 1 is odd.
+    assert!(!vec.par_iter().all(|n| (*n % 2) == 0)); // Not all are even.
```
budziq: Imho the [std examples](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) for all/any are a little more expressive as these are showing inequality instead of just equality operators.  We can always add another two asserts with different predicates. Please consider slightly modifying or expanding the predicates and we are ready to merge.

CJStadler:
Thanks @budziq! I added some examples with inequality.

budziq:
Nicely done @CJStadler  !

Please also checkout #209 and comment there if you consent to repository re-license to CC0.

PR #336, opened 2017-10-17
Updates the wording on error handling to improve clarity in CONTRIBUTING.md

j-haj:
It was not previously clear when a contributor should use `error_chain`
and when they should use `main()`. This patch adds a couple sentences to
the section on error handling to make that more explicit.


budziq:
Thanks a lot!

Issue #330, opened 2017-10-12
Add "Search unordered items in parallel" example


budziq:
Showcase [`find_any`](https://docs.rs/rayon/*/rayon/iter/trait.ParallelIterator.html#method.find_any) with [rayon](https://crates.io/crates/rayon) crate.

Please discuss the idea here before starting implementation.
cc https://github.com/rust-lang-nursery/rust-cookbook/issues/272

budziq:
> Do we care about the size of the data we are searching? 

Lets just show the general idea. Making the example data meaningfully parallelizable  is out of scope for such example.

> Since we are searching, maybe we should have examples for find_any, find_first, and find_last in this recipe since their use is similar.

find_last / find_first impose additional restrictions on the iterable and I'd say that their use is more specialized. I'd need a solid and specific use case here to add these.

j-haj:
I have a very simple example [here](https://github.com/j-haj/rayon-search/blob/master/rayon-search/src/main.rs). Everything is pretty straight forward except for the double reference part. Is it worth explaining why there is a `&&x` in the final example?

budziq:
Nice! Actually rayon shares this quirk with standard [sequential find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) from std we can always link to it in the example (it's already explained there but we'll see how the textual description unfolds :) )

In terms of the actual example we might go wild here and give some more interesting predicate than just direct equality. 
Hmm I would also go with assert_eq to the actual values

j-haj:
Sounds good! I will add a more complex predicate to the example and submit a PR sometime today or tomorrow hopefully

j-haj:
@budziq I am not sure how wild you were thinking we would get here so I added two different approaches to the example I linked earlier:

1. I added a slightly more complex search predicate (find an even number greater than 6) - a little more complex than straight equality, but not materially

2. Created a `Person` struct that implements the `PartialEq` trait, and search a `Vec<Person>` for a an entry with a specified `first_name` and `age` greater than some given value. This is quite a bit more complex, but might be too complex for the goals of the Cookbook. On the other hand, this example may be closer to "real world" than the integers example

budziq:
Both approaches seam fine and relatively minimal (We ca always hide the Vec<Person> creation) and I'm not sure we need the `PartialEq`. I'll let you chose which one you'd put into the cookbook

budziq:
merged

PR #337, opened 2017-10-17
Adds search unordered items in parallel example

j-haj:
cc #330

### DIFF
--- a/src/concurrency.md
+++ b/src/concurrency.md
@@ -3,6 +3,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Mutate the elements of an array in parallel][ex-rayon-iter-mut] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
+| [Search items using given predicate in parallel][ex-rayon-parallel-search] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Sort a vector in parallel][ex-rayon-parallel-sort] | [![rayon-badge]][rayon] [![rand-badge]][rand] | [![cat-concurrency-badge]][cat-concurrency] |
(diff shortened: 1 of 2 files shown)


j-haj:
Didn't mean for c3f4a42 do be included in this.. I will rebase and push and update

src/concurrency.md
```diff
@@ -72,6 +73,46 @@ fn main() {
...
+use rayon::prelude::*;
+
+fn main() {
+    let v = vec![6, 2, 1, 9, 3, 8, 11];
+
+    let f1 = v.par_iter().find_any(|&&x| x == 9);
+    let f2 = v.par_iter().find_any(|&&x| x % 2 == 0 && x > 6);
+    let f3 = v.par_iter().find_any(|&&x| x > 8);
+
+    if let Some(found_9) = f1 {
```
budziq: How about writing it more concisely as
```rust
    assert_eq!(f1, Some(&9));
    assert_eq!(f2, Some(&8));
    assert!(f3 > Some(&8))
```

src/concurrency.md
```diff
@@ -72,6 +73,46 @@ fn main() {
 }
 ```
 
+[ex-rayon-parallel-search]: #ex-rayon-parallel-search
+<a name="ex-rayon-parallel-search"></a>
+## Search unordered items in parallel
```
budziq: Hmm we might want to change the example title to something along the lines of
"Search items with given predicate in parallel". Originally the "unordered" part was to emphasize that we are not doing binary search but I think that mentioning the arbitrary predicate is a stronger usecase.

src/concurrency.md
j-haj: This is much cleaner. It did not occur to me to convert the expected value into an `optional`. Very clever! 👍

j-haj:
Great suggestions! I'll push some updates later this evening

j-haj:
I slightly modified the suggested title to "Search items *using* given predicate in parallel" to make it clear the predicate is associated with the search rather than the items.

budziq:
Nicely done @j-haj !

Issue #324, opened 2017-10-09
Cookbook ideas for chrono


j-haj:
Let's add some recipes for the [`chrono`](https://crates.io/crates/chrono) crate. Feel free to contribute additional examples/ideas or PRs!

Here are some ideas to start us off:

- [ ] Compute elapsed time to execute a segment of code
- [ ] Display formatted date and time using RFC 2822, RFC 3339, and custom formats
- [ ] Create a `DateTime` struct from input using RFC 2822, RFC 3339, and custom formats
- [ ] Example showing checked arithmetic using dates
- [ ] Example showing use of functions from the [`Datelike`](https://docs.rs/chrono/0.4.0/chrono/trait.Datelike.html) and  [`Timelike`](https://docs.rs/chrono/0.4.0/chrono/trait.Timelike.html) traits such as month, weekday, hour, minute, etc.
- [ ] Example showing serialization and deserialization of `DateTime` using [`serde`](https://crates.io/crates/serde)

budziq:
Initial ideas by @j-haj 
Here are some ideas to start us off:

- [ ] Compute elapsed time to execute a segment of code
- [ ] Display formatted date and time using RFC 2822, RFC 3339, and custom formats
- [ ] Create a `DateTime` struct from input using RFC 2822, RFC 3339, and custom formats
- [ ] Example showing checked arithmetic using dates
- [ ] Example showing use of functions from the [`Datelike`](https://docs.rs/chrono/*/chrono/trait.Datelike.html) and  [`Timelike`](https://docs.rs/chrono/*/chrono/trait.Timelike.html) traits such as month, weekday, hour, minute, etc.
- [ ] Example showing serialization and deserialization of `DateTime` using [`serde`](https://crates.io/crates/serde)

Once we get some attention here, I'll stsrt adding final example ideas to the tracking list.

Thanks a lot @j-haj ! :1st_place_medal:

hegza:
I've done something like what you're describing here @budziq / @j-haj, for both formatting and clocking execution time. Could be cool trying to make them as idiomatic as I can. I believe I'll be able to get back to you with a PR about the three first points later down the week.

j-haj:
@hegza Sounds great - can't wait to see the examples!

One thing I noticed the other day is that chrono seems to state that it is not really best used for tracking elapsed time to execute a segment of code:

[...]

PR #338, opened 2017-10-23
Add example for formatting chrono::DateTime

hegza:
Checking in to see if I remembered everything that should be included in a new example.

* Add link to date-and-time-badge at badge-cache.kominick.com
* Add link to date-and-time category at crates.io
* Add all necessary document structure related to the aformentioned

See also: #324

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -1223,8 +1223,8 @@ fn main() {
 ## Display formatted date and time
 [![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
 
-Gets and displays the current time in UTC in the well-known formats RFC 2822 and
-RFC 3339, and in a custom format.
+Gets and displays the current time in UTC using [`Utc::now`] in the well-known
+formats RFC 2822 and RFC 3339, and in a custom format.
 
 ```rust
 extern crate chrono;
@@ -1291,6 +1291,7 @@ fn main() {
 [`SecureRandom::fill`]: https://docs.rs/ring/*/ring/rand/trait.SecureRandom.html#tymethod.fill
 [`seek`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.seek
 [`Stdio::piped`]: https://doc.rust-lang.org/std/process/struct.Stdio.html
+[`Utc::now`]: https://docs.rs/chrono/*/chrono/offset/struct.Utc.html#method.now
 [rand-distributions]: https://doc.rust-lang.org/rand/rand/distributions/index.html
 [replacement string syntax]: https://docs.rs/regex/*/regex/struct.Regex.html#replacement-string-syntax
 
--- a/src/basics.md
+++ b/src/basics.md
@@ -1223,8 +1223,10 @@ fn main() {
 ## Display formatted date and time
 [![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
 
-Gets and displays the current time in UTC using [`Utc::now`] in the well-known
-formats RFC 2822 and RFC 3339, and in a custom format.
+Gets and displays the current time in UTC using [`Utc::now`]. Formats the
+current time in the well-known formats [RFC 2822] using [`DateTime::to_rfc2822`]
+and [RFC 3339] using [`DateTime::to_rfc3339`], and in a custom format using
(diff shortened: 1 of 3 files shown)


src/basics.md
```diff
@@ -1092,6 +1093,35 @@ fn main() {
...
+
+Displays UTC in the well-known formats RFC 2822 and RFC 3339, and in a custom
+format.
+
+```rust
+extern crate chrono;
+use chrono::{DateTime, Utc};
+
+fn main() {
+    // Get coordinated universal time (UTC)
```
budziq: I'd move all the comments to the textual description above with key identifiers hyperlinked to the chrono docs.rs (please remember to link to wildcard "*" version)

src/basics.md
```diff
@@ -1092,6 +1093,35 @@ fn main() {
...
+    println!("UTC is: {}", utc);
+
+    // Display formatted date and time using RFC 2822
+    println!("UTC in RFC 2822 is: {}", utc.to_rfc2822());
+
+    // Display formatted date and time using RFC 3339
+    println!("UTC in RFC 3339 is: {}", utc.to_rfc3339());
+
+    // Display formatted date and time using a custom format
+    const CUSTOM_FORMAT: &str = "%a %b %e %T %Y";
```
budziq: Probably we do not need a separate variable here. Let's keep the example as simple as possible!

src/basics.md
```diff
@@ -1092,6 +1093,35 @@ fn main() {
...
+Displays UTC in the well-known formats RFC 2822 and RFC 3339, and in a custom
+format.
+
+```rust
+extern crate chrono;
+use chrono::{DateTime, Utc};
+
+fn main() {
+    // Get coordinated universal time (UTC)
+    let utc: DateTime<Utc> = Utc::now();
```
budziq: I'd suggest renaming the `utc` to something like `now`

src/basics.md
```diff
@@ -1195,27 +1195,20 @@ Run the recipe with `RUST_BACKTRACE=1` to display a detailed [`backtrace`] assoc
 ## Display formatted date and time
 [![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
 
-Displays UTC in the well-known formats RFC 2822 and RFC 3339, and in a custom
-format.
+Gets and displays the current time in UTC in the well-known formats RFC 2822 and
```
budziq: I'd suggest adding hyperlinks to key identifiers such as [`Utc::now`](https://docs.rs/chrono/*/chrono/offset/struct.Utc.html#method.now)

src/basics.md
```diff
@@ -1188,6 +1190,28 @@ Error level - description
 
 Run the recipe with `RUST_BACKTRACE=1` to display a detailed [`backtrace`] associated with this error.
 
+[ex-format-datetime]: #ex-format-datetime
+<a name="ex-format-datetime"></a>
+## Display formatted date and time
+[![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
+
+Gets and displays the current time in UTC using [`Utc::now`] in the well-known
+formats RFC 2822 and RFC 3339, and in a custom format.
```
budziq: I also suggest to add hyperlinks to other important identifiers such as `to_rfc2822`, `to_rfc3339` and `format`. Sorry if I was not clear previously.

I'd also hyperlink the mentioned RFC's like

>in the standard date formats [RFC 2822](https://www.ietf.org/rfc/rfc2822.txt) and [RFC 3339] as well as in a custom format(https://www.ietf.org/rfc/rfc3339.txt), and in a custom format.

hegza: I was mostly trying to get a feel for the granularity that we want to have for the things to be linked. That's why I was trying to do exactly what you said instead of trying to use my brain. That's likely the source of confusion here. Anyway, thanks for taking the time 👍 I'll be able to do it much better next time.

budziq: No problem!

hegza:
I think this should be okay. Do tell me if there's still something I can do.

budziq:
@hegza nicely done!

Please also checkout https://github.com/rust-lang-nursery/rust-cookbook/issues/209 and comment there if you consent to repository re-license to CC0.

PR #342, opened 2017-10-24
Adds parallel map-reduce recipe

### LINKED ISSUE #329: Add "Parallel map reduce" example
Showcase a basic map-reduce scheme with [rayon](https://crates.io/crates/rayon) crate.

Please discuss the idea here before starting implementation.
cc https://github.com/rust-lang-nursery/rust-cookbook/issues/272

j-haj:
fixes #329

* Adds example in `concurrency.md`
* Adds links to `intro.md`

### DIFF
--- a/src/concurrency.md
+++ b/src/concurrency.md
@@ -5,6 +5,7 @@
 | [Mutate the elements of an array in parallel][ex-rayon-iter-mut] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Search items using given predicate in parallel][ex-rayon-parallel-search] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Sort a vector in parallel][ex-rayon-parallel-sort] | [![rayon-badge]][rayon] [![rand-badge]][rand] | [![cat-concurrency-badge]][cat-concurrency] |
+| [Map-reduce in parallel][ex-rayon-map-reduce] | [![rayon-badge]][rayon] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Generate jpg thumbnails in parallel][ex-rayon-thumbnails] | [![rayon-badge]][rayon] [![glob-badge]][glob] [![image-badge]][image] | [![cat-concurrency-badge]][cat-concurrency][![cat-filesystem-badge]][cat-filesystem] |
 | [Spawn a short-lived thread][ex-crossbeam-spawn] | [![crossbeam-badge]][crossbeam] | [![cat-concurrency-badge]][cat-concurrency] |
 | [Draw fractal dispatching work to a thread pool][ex-threadpool-fractal] | [![threadpool-badge]][threadpool] [![num-badge]][num] [![num_cpus-badge]][num_cpus] [![image-badge]][image] | [![cat-concurrency-badge]][cat-concurrency][![cat-science-badge]][cat-science][![cat-rendering-badge]][cat-rendering] |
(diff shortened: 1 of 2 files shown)


src/concurrency.md
```diff
@@ -73,6 +74,63 @@ fn main() {
...
+        Person { age: 17 },
+        Person { age: 31 },
+        Person { age: 30 },
+    ];
+
+    let gt_30 = |x: u32| x > 30;
+    let num_over_30 = v.par_iter().filter(|&ref x| gt_30(x.age)).count();
+    let sum_over_30 = v.par_iter()
+        .filter(|&ref x| gt_30(x.age))
+        .map(|ref x| x.age)
```
budziq: I'd suggest mapping first and then using `filter(gt_30)` instead of closure in closure

src/concurrency.md
```diff
@@ -73,6 +74,63 @@ fn main() {
...
+        .reduce(|| 0, |x, y| x + y);
+    
+    // Rayon also has a sum() method, which yields the same results as the
+    // reduction on the previous line
```
budziq: I'd move this comment to the textual description above

src/concurrency.md
```diff
@@ -73,6 +74,63 @@ fn main() {
...
+    let alt_sum_30: u32 = v.par_iter()
+        .filter(|&ref x| gt_30(x.age))
+        .map(|ref x| x.age)
+        .sum();
+
+    let avg_over_30 = sum_over_30 as f32 / num_over_30 as f32;
+    let alt_avg_over_30 = alt_sum_30 as f32 / num_over_30 as f32;
+
+    assert!((avg_over_30 - alt_avg_over_30).abs() < 1e-8);
```
budziq: I'd suggest using `std::f32::EPSILON` instead of `1e-8`

src/concurrency.md
```diff
@@ -73,6 +74,63 @@ fn main() {
...
+        .reduce(|| 0, |x, y| x + y);
+    
+    // Rayon also has a sum() method, which yields the same results as the
+    // reduction on the previous line
+    let alt_sum_30: u32 = v.par_iter()
+        .filter(|&ref x| gt_30(x.age))
+        .map(|ref x| x.age)
+        .sum();
+
+    let avg_over_30 = sum_over_30 as f32 / num_over_30 as f32;
```
budziq: I'd probably move casts to above and store the variables as `f32` but it's a matter of taste.

src/concurrency.md
```diff
@@ -73,6 +74,63 @@ fn main() {
...
+        Person { age: 19 },
+        Person { age: 42 },
+        Person { age: 17 },
+        Person { age: 17 },
+        Person { age: 31 },
+        Person { age: 30 },
+    ];
+
+    let gt_30 = |x: u32| x > 30;
+    let num_over_30 = v.par_iter().filter(|&ref x| gt_30(x.age)).count();
```
budziq: the `&ref` and `ref` are not needed anywhere in this example

src/concurrency.md
j-haj: I run into a type inference issue with the `alt_sum_30` variable if I moved `as f32` to the line where `alt_sum_30` is initialized. So far the easiest way to get rid of that error is to make `alt_sum_30` a `u32` and then convert to `f32` later. Any suggestions?

budziq: right `sum` is a template so type inference gets wonky here. 

```rust
 let alt_sum_30: f32 = v.par_iter()
        .filter(|x| gt_30(x.age))
        .map(|x| x.age as f32)
        .sum();
```
```rust
    let alt_sum_30 = v.par_iter()
        .filter(|x| gt_30(x.age))
        .map(|x| x.age as f32)
        .sum::<f32>();
```

or the cleanest solution, leaving is as it was, summed into u32 and casting later on ;)

j-haj:
Thanks for all the comments! I have pushed an updated version

budziq:
Nicely done!

Issue #341, opened 2017-10-24
Add "Measure elapsed time" example


budziq:
Use  `time::Instant` to measure execution time between two code sections.

cc #266

AlexEne:
I gave this a go here: https://github.com/rust-lang-nursery/rust-cookbook/pull/343

PR #343, opened 2017-10-24
Added measure-elapsed-time example

AlexEne:
I just explored a bit this repo, read the contribution guidelines, one thing lead to another and I added the measure-elapsed-time example.

fixes https://github.com/rust-lang-nursery/rust-cookbook/issues/341

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -25,6 +25,7 @@
 | [Access a file randomly using a memory map][ex-random-file-access] | [![memmap-badge]][memmap] | [![cat-filesystem-badge]][cat-filesystem] |
 | [Check number of logical cpu cores][ex-check-cpu-cores] | [![num_cpus-badge]][num_cpus] | [![cat-hardware-support-badge]][cat-hardware-support] |
 | [Obtain backtrace of complex error scenarios][ex-error-chain-backtrace] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
+| [Measure elapsed time][ex-measure-elapsed-time] | [![std-badge]][std] | [![cat-time-badge]][cat-time] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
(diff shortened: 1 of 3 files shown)


src/basics.md
```diff
@@ -1188,6 +1189,48 @@ Error level - description
...
+    println!("Time elapsed in expensive_function() is: {:?}", duration);
+}
+```
+Another way to get the time [`Duration`] is as a difference of two 
+time [`Instant`] objects.
+
+```rust,no_run
+let start = Instant::now();
+expensive_function();
+let duration = Instant::now() - start();
```
budziq: `start()` is not callable, this is why the CI complains

src/basics.md
```diff
@@ -1188,6 +1189,48 @@ Error level - description
...
+    expensive_function();
+    let duration = start.elapsed();
+    
+    println!("Time elapsed in expensive_function() is: {:?}", duration);
+}
+```
+Another way to get the time [`Duration`] is as a difference of two 
+time [`Instant`] objects.
+
+```rust,no_run
```
budziq: I'd remove this second approach all together.

src/basics.md
```diff
@@ -1188,6 +1189,48 @@ Error level - description
 
 Run the recipe with `RUST_BACKTRACE=1` to display a detailed [`backtrace`] associated with this error.
 
+[ex-measure-elapsed-time]: #ex-measure-elapsed-time
+<a name="ex-measure-elapsed-time"></a>
+## Measure the elapsed time between two code sections
+
+[![std-badge]][std] [![cat-time-badge]][cat-time]
+
+One way to get the elapsed time is to directly use the method
```
budziq: Let's reformat the description somewhat to a form like "Measures [`time::Instant::elapsed`] since [`time::Instant::now`]"

src/basics.md
```diff
@@ -1188,6 +1189,48 @@ Error level - description
...
+fn expensive_function() {
+    thread::sleep(Duration::from_millis(10));
+}
+
+fn main() {
+    let start = Instant::now();
+    expensive_function();
+    let duration = start.elapsed();
+    
+    println!("Time elapsed in expensive_function() is: {:?}", duration);
```
budziq: One possible tripwire here is that `elapsed` does not reset the start in any way.

We can suggest following or just mention the fact in the description.

```rust
let start = Instant::now();
expensive_function();
let (start, duration) = (Instant::now(), start.elapsed());
println!("Time elapsed in expensive_function() is: {:?}", duration);
some_other_function();
println!("Time elapsed in some_other_function() is: {:?}", start.elapsed());
```

AlexEne: I will just mention it in the description since I think this example will be a bit harder to follow and since start is reset before the println!(), makes it a bit imprecise.

src/basics.md
AlexEne: Done

src/basics.md
AlexEne: Ah, I see, I believed that it has to be a full rust program so it compiles (with main and all that)

src/basics.md
```diff
@@ -1188,6 +1189,48 @@ Error level - description
...
+Another way to get the time [`Duration`] is as a difference of two 
+time [`Instant`] objects.
+
+```rust,no_run
+let start = Instant::now();
+expensive_function();
+let duration = Instant::now() - start();
+```
+
+This is the fastest way of getting some quick and dirty
```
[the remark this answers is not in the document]
AlexEne: Ok, I missed the point a bit since the description was a bit generic. I will do a PR with these fixes later today.

budziq:
Thanks @AlexEne !

Please also checkout https://github.com/rust-lang-nursery/rust-cookbook/issues/209 and comment there if you consent to repository re-license to CC0.

AlexEne:
I commented on that thread, I agree with the idea and thanks for all the feedback ;)

budziq:
Thanks! Not that much feedback was needed :)

PR #344, opened 2017-10-25
added measure time example

Ophirr33:
I threw this together, but it looks like I got beat out by  #343. Feel free to steal any of my ideas.


budziq:
@Ophirr33 thanks! It's quite nice but much more elaborate than what I was aiming for. We wanted to present a way to report elapsed time and not start building a benchmarking framework (especially as reliable benchmarking is much more sophisticated than one might expect, please see https://github.com/bluss/bencher/) so I'll stick with @AlexEne 's PR for now, sorry!

PR #349, opened 2017-10-28
Add "Redirect both stdout and stderr of child process to the same fle"

### LINKED ISSUE #345: Add "Redirect both stdout and stderr of child process to the same file" example
see https://users.rust-lang.org/t/double-redirection-stdout-stderr/13554

cc #266

ludwigpacifici:
fixes #345

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -11,6 +11,7 @@
 | [Run an external command and process stdout][ex-parse-subprocess-output] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run an external command passing it stdin and check for an error code][ex-parse-subprocess-input] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run piped external commands][ex-run-piped-external-commands] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
+| [Redirect both stdout and stderr of child process to the same file][ex-redirect-stdout-stderr-same-file] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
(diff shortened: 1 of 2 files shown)


ludwigpacifici:
Hello @budziq 

I change the command from `find /` (what I suggested in #345) to`ls`. The execution time is quicker.

Also, I hardcoded the filename. I saw a crate to create temporaries, but the example would be less minimal.

Let me know your thoughts.

src/basics.md
```diff
@@ -448,6 +449,50 @@ fn run() -> Result<()> {
...
+[ex-redirect-stdout-stderr-same-file]: #ex-redirect-stdout-stderr-same-file
+<a name="ex-redirect-stdout-stderr-same-file"></a>
+## Redirect both stdout and stderr of child process to the same file
+
+[![std-badge]][std] [![cat-os-badge]][cat-os]
+
+Spawns a Unix command and redirects `stdout` and `stderr` to the same
+file. It follows the same idea as [run piped external
+commands](#ex-run-piped-external-commands), however [`process::Stdio`]
+will be constructed from a file. In addition, [`File::try_clone`] is
```
budziq: the statement might be slightly misleading. we might want to rewrite to emphasize that `process::Stdio` will write to provided `File`and that `File::try_clone` will duplicate the file handle ensuring the write pointer is synchronized for stderr and stdout

ludwigpacifici:
@budziq, PR updated with your comments.

I also changed changed the equivalent  Unix command from `ls . oops &> out.txt` to
`ls . oops >out.txt 2>&1`. The former [may not be compatible](https://unix.stackexchange.com/questions/176216/redirection-differences-between-and-21) and the later is more [standard](https://stackoverflow.com/questions/876239/how-can-i-redirect-and-append-both-stdout-and-stderr-to-a-file-with-bash).

budziq:
Well done!

ludwigpacifici:
Thanks!

PR #351, opened 2017-10-29
Add example for parsing into chrono::DateTime

hegza:
Add an example for parsing strings in different formats into chrono::DateTime structs.

See also: #324

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -27,6 +27,7 @@
 | [Obtain backtrace of complex error scenarios][ex-error-chain-backtrace] | [![error-chain-badge]][error-chain] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Measure elapsed time][ex-measure-elapsed-time] | [![std-badge]][std] | [![cat-time-badge]][cat-time] |
 | [Display formatted date and time][ex-format-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
+| [Parse string into DateTime struct][ex-parse-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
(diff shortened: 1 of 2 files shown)


src/basics.md
```diff
@@ -1242,6 +1243,49 @@ fn main() {
...
+## Parse string into DateTime struct
+[![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
+
+Parses a [`DateTime`] struct from strings representing the well-known formats
+[RFC 2822], [RFC 3339], and a custom format, using
+[`DateTime::parse_from_rfc2822`], [`DateTime::parse_from_rfc3339`], and
+[`DateTime::parse_from_str`] respectively. Escape sequences that are available
+for the [`DateTime::parse_from_str`] can be found at
+[`chrono::format::strftime`]. Note that the [`DateTime::parse_from_str`]
+requires that such a DateTime struct must be creatable that it uniquely
```
budziq: I would also introduce the Naive{Data, Time, DateTime} family here https://docs.rs/chrono/*/chrono/struct.DateTime.html?search=parse_from_str these are less restrictive in regard to the user input (allow for only date and only time without timezone)

src/basics.md
```diff
@@ -1242,6 +1243,49 @@ fn main() {
...
+[ex-datetime-from-input]: #ex-parse-datetime
+<a name="ex-parse-datetime"></a>
+## Parse string into DateTime struct
+[![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
+
+Parses a [`DateTime`] struct from strings representing the well-known formats
+[RFC 2822], [RFC 3339], and a custom format, using
+[`DateTime::parse_from_rfc2822`], [`DateTime::parse_from_rfc3339`], and
+[`DateTime::parse_from_str`] respectively. Escape sequences that are available
+for the [`DateTime::parse_from_str`] can be found at
```
budziq: We might want to split it into two paragraphs (at least the part about escape sequences)

hegza:
Extended based on feedback. Is this what we are going after? Do comment if there's still something missing.

src/basics.md
```diff
@@ -1242,6 +1243,52 @@ fn main() {
...
+Parses a [`DateTime`] struct from strings representing the well-known formats
+[RFC 2822], [RFC 3339], and a custom format, using
+[`DateTime::parse_from_rfc2822`], [`DateTime::parse_from_rfc3339`], and
+[`DateTime::parse_from_str`] respectively.
+
+Escape sequences that are available for the [`DateTime::parse_from_str`] can be
+found at [`chrono::format::strftime`]. Note that the [`DateTime::parse_from_str`]
+requires that such a DateTime struct must be creatable that it uniquely
+identifies a date and a time. For parsing dates and times without timezones, see
+[`NaiveDate::parse_from_str`], [`NaiveTime::parse_from_str`], and
```
budziq: I would suggest to also use these functions in the actual code and just mention here about the types `NaiveDate` 
 / `Naivetime`  / `NaiveDateTime`

budziq: > Is this what we are going after?

Pretty much. :+1:  Just the last nit listed above.

hegza:
Should be it. I noticed the PR-template as well and checked all things and squashed the commits in accordance with that.

budziq:
Merged! Thanks @hegza !

PR #352, opened 2017-10-31
Add example for checked arithmetic of chrono

### LINKED ISSUE #359: Add  "checked arithmetic using dates" recipe
With the [`chrono`](https://crates.io/crates/chrono) crate, write a recipe showing how to do date arithmetic.

cc #324

See [CONTRIBUTING.md](https://github.com/rust-lang-nursery/rust-cookbook/blob/master/CONTRIBUTING.md) for details.

hegza:
See also: #324

fixes #359

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -29,6 +29,7 @@
 | [Measure elapsed time][ex-measure-elapsed-time] | [![std-badge]][std] | [![cat-time-badge]][cat-time] |
 | [Display formatted date and time][ex-format-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Parse string into DateTime struct][ex-parse-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
+| [Perform checked date and time calculations][ex-datetime-arithmetic] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
@@ -1343,6 +1344,47 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-datetime-arithmetic]: #ex-datetime-arithmetic
+<a name="ex-datetime-arithmetic"></a>
+## Perform checked date and time calculations
+[![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
+
+Calculates and displays the date and time two weeks from now using
+[`DateTime::checked_add_signed`] and the date of the day before that using
+[`DateTime::checked_sub_signed`]. The methods return None if the date and time
+cannot be calculated.
+
+Escape sequences that are available for the
+[`DateTime::format`] can be found at [`chrono::format::strftime`].
+
+```rust
+extern crate chrono;
+use chrono::{DateTime, Duration, Utc};
+
+fn day_earlier(date_time: DateTime<Utc>) -> Option<DateTime<Utc>> {
+    date_time.checked_sub_signed(Duration::days(1))
+}
+
+fn main() {
+    let now = Utc::now();
+    println!("{}", now);
+
+    let almost_three_weeks_from_now = now.checked_add_signed(Duration::weeks(2))
+            .and_then(|in_2weeks| in_2weeks.checked_add_signed(Duration::weeks(1)))
+            .and_then(day_earlier);
+
+    match almost_three_weeks_from_now {
+        Some(x) => println!("{}", x),
+        None => eprintln!("Almost three weeks from now overflows!"),
+    }
+
+    match now.checked_add_signed(Duration::max_value()) {
+        Some(x) => println!("{}", x),
+        None => eprintln!("We can't use chrono to tell the time for the Solar System to complete more than one full orbit around the galactic center."),
+    }
+}
+```
+
 {{#include links.md}}
 
 <!-- API Reference -->
@@ -1361,9 +1403,10 @@ fn run() -> Result<()> {
 [`DateTime::parse_from_rfc3339`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.parse_from_rfc3339
 [`DateTime::parse_from_str`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.parse_from_str
 [`DateTime::to_rfc2822`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc2822
-[`DateTime::to_rfc2822`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc2822
 [`DateTime::to_rfc3339`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc3339
-[`DateTime::to_rfc3339`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc3339
+[`DateTime::format`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.format
+[`DateTime::checked_add_signed`]: https://docs.rs/chrono/*/chrono/struct.Date.html#method.checked_add_signed
+[`DateTime::checked_sub_signed`]: https://docs.rs/chrono/*/chrono/struct.Date.html#method.checked_sub_signed
 [`DateTime`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html
 [`digest::Context`]: https://docs.rs/ring/*/ring/digest/struct.Context.html
 [`digest::Digest`]: https://docs.rs/ring/*/ring/digest/struct.Digest.html
--- a/src/intro.md
+++ b/src/intro.md
@@ -47,6 +47,7 @@ community. It needs and welcomes help. For details see
 | [Measure elapsed time][ex-measure-elapsed-time] | [![std-badge]][std] | [![cat-time-badge]][cat-time] |
 | [Display formatted date and time][ex-format-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Parse string into DateTime struct][ex-parse-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
+| [Perform checked date and time calculations][ex-datetime-arithmetic] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 
 ## [Encoding](encoding.html)
 
@@ -164,11 +165,13 @@ community. It needs and welcomes help. For details see
 [ex-csv-filter]: encoding.html#ex-csv-filter
 [ex-csv-read]: encoding.html#ex-csv-read
 [ex-csv-serde]: encoding.html#ex-csv-serde
+[ex-datetime-arithmetic]: basics.html#ex-datetime-arithmetic
 [ex-dedup-filenames]: app.html#ex-dedup-filenames
 [ex-error-chain-backtrace]: basics.html#ex-error-chain-backtrace
 [ex-extract-hashtags]: basics.html#ex-extract-hashtags
 [ex-extract-links-webpage]: net.html#ex-extract-links-webpage
 [ex-extract-mediawiki-links]: net.html#ex-extract-mediawiki-links
+[ex-format-datetime]: basics.html#ex-format-datetime
 [ex-file-post]: net.html#ex-file-post
 [ex-file-predicate]: app.html#ex-file-predicate
 [ex-file-sizes]: app.html#ex-file-sizes


src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+<a name="ex-datetime-arithmetic"></a>
+## Add or subtract dates and times using checked arithmetic
+[![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
+
+Calculates and displays the date and time two weeks from now using
+[`DateTime::checked_add_signed`] and the date of yesterday using
+[`DateTime::checked_sub_signed`]. The methods return None if the date and time
+cannot be calculated.
+
+Escape sequences that are available for the
```
ludwigpacifici: [Display formatted date and time](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#display-formatted-date-and-time) is already described in the cookbook. if you want to have one line explaining this, I would forward the reader to the relevant recipe.

src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+use chrono::{Duration, Utc};
+
+fn main() {
+    let now = Utc::now();
+    println!("The date and time are {}", now);
+
+    if let Some(two_weeks_from_now) = now.checked_add_signed(Duration::weeks(2)) {
+        println!("Two weeks from now is {}", two_weeks_from_now);
+    } else {
+        println!("Two weeks from now overflows!");
```
ludwigpacifici: Usually errors are handled via `error_chain` crate. Not sure here it would make the code shorter. However, one way to do it is to convert the `Option` returned by `checked_add_signed` and convert it into an error via `ok_or` and use `?` to bubble up the error. Also the code has to be moved into a function `fn run() -> Result<()>`. Check other recipes for this kind of example.

Another way to improve errors is use `eprintln!` instead of `println!`.

src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+    }
+
+    if let Some(yesterday) = now.checked_sub_signed(Duration::days(1)) {
+        println!("Yesterday was {}", yesterday.format("%a %b %e %Y"));
+    } else {
+        println!("Yesterday overflows!");
```
ludwigpacifici: Same as my previous comment.

src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+    }
+
+    if let Some(heat_death) = now.checked_add_signed(Duration::max_value()) {
+        println!(
+            "Heat death of the universe: {}",
+            heat_death
+        );
+    } else {
+        println!("We can't use chrono to tell the date of the heat death of the universe.");
```
ludwigpacifici: Same as my previous comment.

src/basics.md
```diff
@@ -1297,6 +1342,9 @@ fn main() {
 [`DateTime::to_rfc2822`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc2822
 [`DateTime::to_rfc3339`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.to_rfc3339
 [`DateTime::format`]: https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.format
+[`DateTime::checked_add_signed`]: https://docs.rs/chrono/*/chrono/struct.Date.html#method.checked_add_signed
```
ludwigpacifici: Please, sort links.

> non rendered items are in sorted order (links, reference, identifiers, Cargo.toml.

from [pull request template](https://github.com/rust-lang-nursery/rust-cookbook/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

ludwigpacifici:
Thanks @hegza for your PR! I suggested some modifications. Tell me what you think. Wait for @budziq for his official review.

One last comment about your commit message, can you reword it to follow the below:

> PR contains correct "fixes #ISSUE_ID" clause to autoclose the issue once PR is merged

from [pull request template](https://github.com/rust-lang-nursery/rust-cookbook/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

src/basics.md
hegza: If I understood correctly what you meant, I feel like this would indeed reduce redundancy, but it would also disorient the reader. Here, it cannot be assumed that the reader is familiar with all the recipes. That's why I would probably prefer to just repeat something as simple as this, rather than link to another recipe.

src/basics.md
hegza: I did wonder a bit about chrono's decision to use `Option` here instead of Result. I think `Result` would be more correct here. But as it is, I felt that it's not appropriate to use `error_chain` with `Option` and went for the pattern matching approach.

I would love to know if my reasoning has got something to improve here. I will definitely do something about it, at least add the `eprintln!`, but do tell if you still have something to add to this.

src/basics.md
hegza: The links in that whole section are generally unordered. That's why I followed chronological ordering grouped by topic / crate. Should I perhaps alpha-sort the whole list of API reference links, or just the ones belonging to the `chrono` crate:

hegza:
The PR does not fix any specific issue, although it is related to, and partially fixes issue #324. In this case, I would prefer not to put an auto-closing tag. If I happen to be the person to commit the last fix to that issue #324, I promise to format the commit message accordingly ^^

ludwigpacifici:
My mistake, I though you were solving a specific issue. In that case, it make sense to not autoclose the issue :-)

@budziq is not inconvenient in case someone can autoclose a todo list by solving one item? Is it worth it to always add a specific issue for solving one item of the list?

src/basics.md
ludwigpacifici: Can you alpha-sort the whole API Reference list in a separate commit in this PR?

hegza: Yes, can do. I'll get on it as soon as I'm able.

src/basics.md
ludwigpacifici: > I did wonder a bit about chrono's decision to use Option here instead of Result.

I agree with you. And I think it's worth to discuss it with the `chrono` team. there is [a similar talk](https://github.com/chronotope/chrono/issues/110).

> do tell if you still have something to add to this.

I though of a third solution, you keep the `if let Some` and remove the else statement. However, it's an ugly because it hides errors.

In order to be homogeneous with the rest of the cookbook and be slightly more concise, I'll go with the `error_chain` solution.

@budziq any thoughts?

src/basics.md
ludwigpacifici: > If I understood correctly what you meant

Sorry for not being clear. You guessed it right :-)

My thoughts behind forwarding the reader: it makes the cookbook more discoverable (help the reader to be familiar with several recipes) and each recipe focuses solely one item (less redundancy).

I let you discuss it with @budziq.

src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+
+    if let Some(yesterday) = now.checked_sub_signed(Duration::days(1)) {
+        println!("Yesterday was {}", yesterday.format("%a %b %e %Y"));
+    } else {
+        println!("Yesterday overflows!");
+    }
+
+    if let Some(heat_death) = now.checked_add_signed(Duration::max_value()) {
+        println!(
+            "Heat death of the universe: {}",
```
budziq: Heat death of the universe is estimated at ~10e103 years while Duration::max_value() is 2^63 (i64::MAX) in milliseconds making it roughly 2.9243e8 years. So our statement would be a little inaccurate (94 orders of magnitude give or take ;)
Which is sad cause the idea was awesome :)

I would suggest to change it to something like ("maximal representable time" ) unless you come up with something equally awesome as heat death of the universe.

src/basics.md
```diff
@@ -1242,6 +1243,50 @@ fn main() {
...
+    println!("The date and time are {}", now);
+
+    if let Some(two_weeks_from_now) = now.checked_add_signed(Duration::weeks(2)) {
+        println!("Two weeks from now is {}", two_weeks_from_now);
+    } else {
+        println!("Two weeks from now overflows!");
+    }
+
+    if let Some(yesterday) = now.checked_sub_signed(Duration::days(1)) {
+        println!("Yesterday was {}", yesterday.format("%a %b %e %Y"));
```
budziq: lets simplify the example and description by removing the `format`

src/basics.md
budziq: > I did wonder a bit about chrono's decision to use Option

It follows from the [stdlibs decision about checked integer arithmetic](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_add). I agree that both of these decisions are debatable. 

If the `Option` `None` variant here is an `Err` is somewhat context dependent (personally I'd just go with `match` here instead of `if let Some()` or `error_chain` in the default case) but if we build some additional narrative (ie. put the calculations into a well named function indicating that it could fail) I think that we could justify `error_chain` usage. What are you thoughts?

I'd also suggest adding a series of calculations with [`and_then`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then) combinator to spice up the example.

src/basics.md
hegza: No problem.

src/basics.md
hegza: Yeah, I tried to do it with the correct numbers at first, but realized that I'd have to delegate the heavy-lifting to a new crate, and that was a place I wasn't prepared to go to. :D I'll figure it out.

hegza:
I made another iteration. Feel free to check it out. I'll be on stand by until the merges are required.

budziq:
@ludwigpacifici 

> @budziq is not inconvenient in case someone can autoclose a todo list by solving one item? Is it worth it to always add a specific issue for solving one item of the list?

Eh my bad. The "fixes" clause is only meant for PR's addressing a specific "example idea" issues but its not stated anywhere (mostly due to PR template being overly verbose and restrictive already). If anyone has a suggestion to make it clearer in a form that is not even more TL;NR I'm all ears :) In general I see no point in autoclosing the tracking issues as these are meant to gather example ideas from readers and contributors.

Also my bad again for not making the "example idea" issues for chrono untill now haven't had time much time near PC lately (currently writing on mobile).

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+cannot be calculated.
+
+Escape sequences that are available for the
+[`DateTime::format`] can be found at [`chrono::format::strftime`].
+
+```rust
+extern crate chrono;
+use chrono::{DateTime, Duration, Utc};
+
+fn yesterday(date_time: DateTime<Utc>) -> Option<DateTime<Utc>> {
```
budziq: how about renaming it to `day_earlier` making it a little more accurate as we are not necessarily anchored in now?

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+    date_time.checked_sub_signed(Duration::days(1))
+}
+
+fn main() {
+    let now = Utc::now();
+    println!("{}", now);
+
+    let two_weeks_from_now = now.checked_add_signed(Duration::weeks(2));
+    match two_weeks_from_now {
```
budziq: how about pulling this calculation directly into the almost_three_weeks_from_now variable or at least dropping the whole match+printout here as the example gets a little noisy with it.

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+        Some(x) => println!("{}", x),
+        None => println!("Two weeks from now overflows!"),
+    }
+
+    let almost_three_weeks_from_now =
+            two_weeks_from_now.unwrap()
```
budziq: Lets not use `unwrap` in our examples this is exactly why `and_then` was introduced :+1: 

```rust
let almost_three_weeks_from_now = now.checked_add_signed(Duration::weeks(2))
        .and_then(|in_2weeks| in_2weeks.checked_add_signed(Duration::weeks(1)))
        .and_then(day_earlier);
```

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+            .checked_add_signed(Duration::weeks(1)).and_then(yesterday);
+    match almost_three_weeks_from_now {
+        Some(x) => println!("{}", x),
+        None => println!("Almost three weeks from now overflows!"),
```
budziq: if we are simulating onsite error handling lets use `eprintln!`

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+    }
+
+    let time_until_full_galactic_orbit = now.checked_add_signed(Duration::max_value());
+    match time_until_full_galactic_orbit {
```
budziq: how about compacting these two lines into one?
```rust
match now.checked_add_signed(Duration::max_value()) {
```

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+
+    let two_weeks_from_now = now.checked_add_signed(Duration::weeks(2));
+    match two_weeks_from_now {
+        Some(x) => println!("{}", x),
+        None => println!("Two weeks from now overflows!"),
+    }
+
+    let almost_three_weeks_from_now =
+            two_weeks_from_now.unwrap()
+            .checked_add_signed(Duration::weeks(1)).and_then(yesterday);
```
budziq: I would introduce a newline between the variable creation and `match` to improve the readability.

ludwigpacifici:
FYI, @hegza you have your dedicated issue: #359. If you don't mind to reword your commit message with you next update :-)

src/basics.md
hegza: A very good idea. I wanted something like that but wasn't able to properly express myself.

src/basics.md
```diff
@@ -1242,6 +1243,53 @@ fn main() {
...
+    match almost_three_weeks_from_now {
+        Some(x) => println!("{}", x),
+        None => println!("Almost three weeks from now overflows!"),
+    }
+
+    let time_until_full_galactic_orbit = now.checked_add_signed(Duration::max_value());
+    match time_until_full_galactic_orbit {
+        Some(x) => println!("{}", x),
+        None => println!("We can't use chrono to tell the date of the Solar System completing one full orbit around the galactic center."),
```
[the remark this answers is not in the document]
hegza: I would expect no less scientific accuracy here. My mistake :D

hegza:
Fixed, reworded. Looking fine now, to me at least. Drop me a comment if you disagree!

src/basics.md
```diff
@@ -1343,6 +1344,47 @@ fn run() -> Result<()> {
...
+            .and_then(day_earlier);
+
+    match almost_three_weeks_from_now {
+        Some(x) => println!("{}", x),
+        None => eprintln!("Almost three weeks from now overflows!"),
+    }
+
+    match now.checked_add_signed(Duration::max_value()) {
+        Some(x) => println!("{}", x),
+        None => println!("We can't use chrono to tell the time for the Solar System to complete more than one full orbit around the galactic center."),
```
budziq: Lets use `eprintln!` here  consistently with the previous match and we are ready to merge. I'd do it myself by I'm mobile only this weekend.

src/basics.md
```diff
@@ -1343,6 +1344,47 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-datetime-arithmetic]: #ex-datetime-arithmetic
+<a name="ex-datetime-arithmetic"></a>
+## Add or subtract dates and times using checked arithmetic
```
budziq: I'd suggest renaming the example to "Perform checked date and time calculations". It will be slightly more discoverable.

src/basics.md
hegza: Oh, weird that I missed that while updating the other eprintln!

hegza:
Ready once more.

budziq:
Thanks for sticking with it. Nicely done @hegza 👍 
And thanks for help with the review @ludwigpacifici :)

PR #366, opened 2017-11-13
Add Datelike and Timelike example

### LINKED ISSUE #357: Add "use of functions from the Datelike and Timelike traits such as month, weekday, hour, minute" example
With the [`chrono`](https://crates.io/crates/chrono) crate, write a recipe showing the use of [`chrono::Datelike`](https://docs.rs/chrono/*/chrono/trait.Datelike.html) and [`chrono::Timelike`](https://docs.rs/chrono/*/chrono/trait.Timelike.html) traits.

cc #324

See [CONTRIBUTING.md](https://github.com/rust-lang-nursery/rust-cookbook/blob/master/CONTRIBUTING.md) for details.

CrockAgile:
fixes #357

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -30,7 +30,7 @@
 | [Display formatted date and time][ex-format-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Parse string into DateTime struct][ex-parse-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Perform checked date and time calculations][ex-datetime-arithmetic] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
-| [Examine a DateTime struct via Datelike and Timelike traits][ex-examine-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
+| [Examine the date and time][ex-examine-date-and-time] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
@@ -1386,9 +1386,9 @@ fn main() {
 }
 ```
 
-[ex-examine-datetime]: #ex-examine-datetime
-<a name="ex-examine-datetime"></a>
-## Examine a DateTime struct via Datelike and Timelike traits
+[ex-examine-date-and-time]: #ex-examine-date-and-time
+<a name="ex-examine-date-and-time"></a>
+## Examine the date and time
 [![chrono-badge]][chrono] [![cat-date-and-time-badge]][cat-date-and-time]
 
 Gets the current UTC [`DateTime`] and its hour/minute/second via [`Timelike`]
--- a/src/basics.md
+++ b/src/basics.md
@@ -1401,7 +1401,6 @@ use chrono::{Datelike, Timelike, Utc};
 fn main() {
     let now = Utc::now();
 
-    // available due to chrono::Timelike
     let (is_pm, hour) = now.hour12();
     println!(
         "The current UTC time is {:02}:{:02}:{:02} {}",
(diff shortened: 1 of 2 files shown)


src/basics.md
```diff
@@ -1385,6 +1386,44 @@ fn main() {
...
+
+```rust
+extern crate chrono;
+use chrono::{Datelike, Timelike, Utc};
+
+fn main() {
+    let now = Utc::now();
+
+    // available due to chrono::Timelike
+    let second = now.second();
```
budziq: I'd suggest:
- running rustfmt on the example
- dropping the intermediate variables and making the queries directly in the `println!`body
```rust
    let (is_pm, hour) = now.hour12();
    println!(
        "The current UTC time is {:02}:{:02}:{:02} {}",
        hour,
        now.minute(),
        now.second(),
        if is_pm { "PM" } else { "AM" },
    );
```

The example should ges slightly less noisy.

src/basics.md
```diff
@@ -1385,6 +1386,44 @@ fn main() {
 }
 ```
 
+[ex-examine-datetime]: #ex-examine-datetime
+<a name="ex-examine-datetime"></a>
+## Examine a DateTime struct via Datelike and Timelike traits
```
budziq: I'd probably avoid using the tpe and trait names in the example title. How about something along the lines of "Examine the date and time"?

CrockAgile: Sounds good! I was trying to be explicit toward the issue, but you are right that including it makes it a little verbose/wordy. Thanks for the review

src/basics.md
CrockAgile: I agree. Good change

ludwigpacifici:
Will it be useful to show also the other way around, i.e. from a string date (example "2017/12/31T12:34:56-1:00") get the [DateTime](https://docs.rs/chrono/*/chrono/struct.DateTime.html) via [parse_from_rfc3339](https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.parse_from_rfc3339) or [parse_from_str](https://docs.rs/chrono/*/chrono/struct.DateTime.html#method.parse_from_str)?

budziq:
> Will it be useful to show also the other way around, i.e. from a string date 

We already have date [parsing example](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#parse-string-into-datetime-struct) I'd suggest to keep this example focused as it is.

src/basics.md
```diff
@@ -1385,6 +1386,52 @@ fn main() {
...
+and its year/month/day/weekday via [`Datelike`].
+
+```rust
+extern crate chrono;
+use chrono::{Datelike, Timelike, Utc};
+
+fn main() {
+    let now = Utc::now();
+
+    // available due to chrono::Timelike
```
ludwigpacifici: The comment repeats what is said in recipe description (few lines above). Does it make sense to remove it (to make example slightly shorter)?

src/basics.md
```diff
@@ -1385,6 +1386,52 @@ fn main() {
...
+        now.minute(),
+        now.second(),
+        if is_pm { "PM" } else { "AM" }
+    );
+    println!(
+        "And there have been {} seconds since midnight",
+        now.num_seconds_from_midnight()
+    );
+
+    // available due to chrono::Datelike
```
ludwigpacifici: The comment repeats what is said in recipe description (few lines above). Does it make sense to remove it (to make example slightly shorter)?

ludwigpacifici:
> We already have date parsing example I'd suggest to keep this example focused as it is.

My bad, I have not seen it. Thanks!

src/basics.md
budziq: Yeah, this is a good idea. 
Lets drop the comments and we are ready to merge!

Also, the example title is still a little off but I can't come up with a better one at the moment. @CrockAgile If you can come up with something slightly more natural please propose one but it will not block the PR. 
Well done!

CrockAgile: @budziq I also can't get a better title in my head, but I agree it is lacking. "Get current date and time"? I am open to suggestions. Hard to be more detailed/exciting when there are already examples for parsing and modifying datetimes.

budziq:
Thanks @CrockAgile!

Please drop a comment on #209 if you consent to repository being relicensed with CC0 license!

PR #368, opened 2017-11-15
Add "Avoid discarding errors during error conversions" example

### LINKED ISSUE #215: add "Avoid discarding errors during error conversions" example
A slightly more complex example than https://github.com/brson/rust-cookbook/issues/214 showing aggregation of custom and crate errors that might use chain_err

Example should show off error-chain crate

cc https://github.com/brson/rust-cookbook/issues/184

ludwigpacifici:
fixes #215


ludwigpacifici:
Hello @budziq 

To me, the use case (get a random number via HTTP) is not very useful, but it shows the error handling with a relatively short code and covers what was discussed in #215.

If you have some ideas to improve it, let me know.

src/basics.md
```diff
@@ -1144,6 +1145,74 @@ fn main() {
...
+
+[Matching] on different error types returned by a function is possible
+and relatively compact with [error-chain] crate. To do so,
+[`ErrorKind`] is used to determine the error type.
+
+To illustrate the error matching, a random integer generator web
+service will be queried via [reqwest] and then the response will be
+parsed. Errors can be generated by the Rust standard library,
+[reqwest] and by the web service. [`foreign_links`] are used for well
+defined Rust errors. The `errors` macro is used to create an
```
budziq: `errors` is actually not a macro but a clause inside a `error_chain!` macro

src/basics.md
```diff
@@ -1144,6 +1145,74 @@ fn main() {
...
+
+    println!("a random number between 0 and 10: {}", random_value);
+
+    Ok(())
+}
+
+fn main() {
+    if let Err(error) = run() {
+        match *error.kind() {
+            ErrorKind::Io(_) => println!("{:?}", error),
```
budziq: We might want to differentiate the `println!`s here a little. 1-2 word preamble should be enough

src/basics.md
ludwigpacifici: Thanks for the clarification! I did not knew that. Is it called a [DSL](https://rustbyexample.com/macros/dsl.html#domain-specific-languages-dsls)?

budziq: > Is it called a DSL?

Yep, the error-chain crate defines a DSL via its macros.

ludwigpacifici:
Thanks for the comments. I updated the PR.

budziq:
Nicely done!

ludwigpacifici:
Thank you @budziq !

PR #371, opened 2017-11-25
semver example: find the latest version satsifying given range

### LINKED ISSUE #285: add "Find the latest version satisfying given range" example
Given a set of version strings, find the latest version in a set that satisfies the given range (like docs.rs does in its urls). Use asserts to verify correct serialization and deserailization.

anna-liao:
fixes #285 r? @budziq

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -17,6 +17,7 @@
 | [Parse and increment a version string][ex-semver-increment] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Parse a complex version string][ex-semver-complex] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check if given version is pre-release][ex-semver-prerelease] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
+| [Find the latest version satisfying given range][ex-semver-latest] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check external command version for compatibility][ex-semver-command] | [![semver-badge]][semver] | [![cat-text-processing-badge]][cat-text-processing] [![cat-os-badge]][cat-os]
 
 
@@ -725,6 +726,69 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-semver-latest]: #ex-semver-latest
+<a name="ex-semver-latest"></a>
+## Find the latest version satisfying given range
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given a list of version &strs, finds the latest [`semver::Version`] that satisfying a given [`semver::VersionReq`] using [`VersionReq::matches`].
+
+```rust
+# #[macro_use]
+# extern crate error_chain;
+extern crate semver;
+
+use semver::{Version, VersionReq};
+#
+# error_chain! {
+#     foreign_links {
+#         SemVer(semver::SemVerError);
+#         SemVerReq(semver::ReqParseError);
+#     }
+# }
+
+fn find_max_matching_version<'a, I>(version_req_str: &str, iterable: I) -> Result<Option<Version>>
+where
+    I: IntoIterator<Item = &'a str>,
+{
+    let vreq = VersionReq::parse(version_req_str)?;
+
+    Ok(
+        iterable
+            .into_iter()
+            .filter_map(|s| Version::parse(s).ok())
+            .filter(|s| vreq.matches(s))
+            .max(),
+    )
+}
+
+fn run() -> Result<()> {
+    assert_eq!(
+        find_max_matching_version("<= 1.0.0", vec!["0.9.0", "1.0.0", "1.0.1"])?,
+        Some(Version::parse("1.0.0")?)
+    );
+
+    // Shows Semver precedence for pre-release tags
+    assert_eq!(
+        find_max_matching_version(
+            ">1.2.3-alpha.3",
+            vec![
+                "1.2.3-alpha.3",
+                "1.2.3-alpha.4",
+                "1.2.3-alpha.10",
+                "1.2.3-beta.4",
+                "3.4.5-alpha.9",
+            ]
+        )?,
+        Some(Version::parse("1.2.3-beta.4")?)
+    );
+
+    Ok(())
+}
+#
+# quick_main!(run);
+```
+
 [ex-semver-command]: #ex-semver-command
 <a name="ex-semver-command"></a>
 ## Check external command version for compatibility
@@ -807,6 +871,7 @@ fn run() -> Result<()> {
 [`tar::Entries`]: https://docs.rs/tar/*/tar/struct.Entries.html
 [`tar::Entry`]: https://docs.rs/tar/*/tar/struct.Entry.html
 [`Version::parse`]: https://docs.rs/semver/*/semver/struct.Version.html#method.parse
+[`VersionReq::matches`]: https://docs.rs/semver/*/semver/struct.VersionReq.html#method.matches
 [`WalkDir::depth`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html#method.depth
 [`WalkDir::DirEntry`]: https://docs.rs/walkdir/*/walkdir/struct.DirEntry.html
 [`WalkDir::max_depth`]: https://docs.rs/walkdir/*/walkdir/struct.WalkDir.html#method.max_depth
--- a/src/intro.md
+++ b/src/intro.md
@@ -127,6 +127,7 @@ community. It needs and welcomes help. For details see
 | [Parse and increment a version string][ex-semver-increment] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Parse a complex version string][ex-semver-complex] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check if given version is pre-release][ex-semver-prerelease] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
+| [Find the latest version satisfying given range][ex-semver-latest] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check external command version for compatibility][ex-semver-command] | [![semver-badge]][semver] | [![cat-text-processing-badge]][cat-text-processing] [![cat-os-badge]][cat-os]
 
 ## [Logging](logging.html)
@@ -238,6 +239,7 @@ community. It needs and welcomes help. For details see
 [ex-semver-command]: app.html#ex-semver-command
 [ex-semver-complex]: app.html#ex-semver-complex
 [ex-semver-increment]: app.html#ex-semver-increment
+[ex-semver-latest]: app.html#ex-semver-latest
 [ex-semver-prerelease]: app.html#ex-semver-prerelease
 [ex-serialize-csv]: encoding.html#ex-serialize-csv
 [ex-sha-digest]: basics.html#ex-sha-digest


src/app.md
```diff
@@ -725,6 +726,47 @@ fn run() -> Result<()> {
...
+# error_chain! {
+#     foreign_links {
+#         SemVer(semver::SemVerError);
+#     }
+# }
+
+fn run() -> Result<()> {
+    let mut ver = Version::from_str("0.0.0").unwrap();
+    let vset = ["0.9.0", "1.0.0", "1.0.1"];
+    let r = VersionReq::parse("<= 1.0.0").unwrap();
```
ludwigpacifici: Let's use `error_chain` and replace `unwrap()` by `?`. If the reader copy paste the snippet and modify the version with a wrong value, the program will exit with a message error more useful than a panic.

ludwigpacifici:
Hello @anna-liao 

Thank you for your PR!

Here are some suggestions to help you improve your semver example:
* do not forget to update `intro.md` file so that it also appears in the table of contents
* @budziq, should the example be enriched by adding pre-release tag into the versions? If yes, the crate as a [specific behaviour](https://docs.npmjs.com/misc/semver#prerelease-tags) that is not obvious and might be worth to highlight ([crate code](https://github.com/steveklabnik/semver/blob/c9fd3667fd5837fd12216e375a81c7e9a08d9d1d/src/version_req.rs#L375)).

src/app.md
```diff
@@ -725,6 +726,47 @@ fn run() -> Result<()> {
...
+
+    for v in vset.iter() {
```
ludwigpacifici: Let's try to make this `for` loop more Rust idiomatic by using iterators. For example:
1. `filter_map` to convert valid `vset` string versions to `semver::Version`
2. `filter` versions which does not match your `VersionReq`
3. and use `max` to get the latest version from the range

anna-liao:
@ludwigpacifici Thanks for the review!  I have updated my PR, though have not added the pre-release tag yet, pending response from @budziq

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+
+```rust
+#[macro_use]
+extern crate error_chain;
+extern crate semver;
+
+use semver::Version;
+use semver::VersionReq;
+
+mod errors {
```
ludwigpacifici: You can make your example more concise by removing `mod errors` scope.

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+        SemVer(semver::SemVerError);
+    }
+}
+
+fn run() -> Result<()> {
+    let vset = ["0.9.0", "1.0.0", "1.0.1"];
+    let r = VersionReq::parse("<= 1.0.0")
+        .chain_err(|| "unable to parse VersionReq")?;
+
+    let vmax = vset.iter()
```
ludwigpacifici: You can make your example more concise by removing `vset` and go `["0.9.0", "1.0.0", "1.0.1"].iter()...`

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-semver-latest]: #ex-semver-latest
+<a name="ex-semver-latest"></a>
+## Find the latest version satisfying given range
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given an array of unparsed version `&str`s, iterate and filter against a semver `VersionReq` and find the latest `Version`.
```
ludwigpacifici: I would give more details in the description. For example, explain the range is obtain via the `matches` and the `max`.

ludwigpacifici:
Thanks @anna-liao. I added some more comments, but I would suggest you to wait for @budziq feedback and then proceed to the changes. I just saw he [mentioned another solution (with sets)](https://github.com/rust-lang-nursery/rust-cookbook/issues/285#issuecomment-340681964) to implement it. Let's see which is more suitable for this example.

src/app.md
budziq: this whole mod can be deleted.

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+extern crate semver;
+
+use semver::Version;
+use semver::VersionReq;
+
+mod errors {
+    error_chain! { }
+}
+
+error_chain! {
```
budziq: please checkout our [note about error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling).

The error boilerplate (error definition and error-chain import) should be hidden with '#'

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+    foreign_links {
+        SemVer(semver::SemVerError);
```
budziq: SemVerError is not actually used in this example

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+    }
+}
+
+fn run() -> Result<()> {
+    let vset = ["0.9.0", "1.0.0", "1.0.1"];
+    let r = VersionReq::parse("<= 1.0.0")
+        .chain_err(|| "unable to parse VersionReq")?;
```
budziq: you can drop the `chain_err` if you add `semver::ReqParseError` to the `foreign_links`

src/app.md
budziq: agreed :+1:

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+
+error_chain! {
+    foreign_links {
+        SemVer(semver::SemVerError);
+    }
+}
+
+fn run() -> Result<()> {
+    let vset = ["0.9.0", "1.0.0", "1.0.1"];
+    let r = VersionReq::parse("<= 1.0.0")
```
budziq: lets improve the readability by using longer variable names

src/app.md
budziq: also we can use slightly longer variable names to improve readabaility

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+        .chain_err(|| "unable to parse VersionReq")?;
+
+    let vmax = vset.iter()
+                   .filter_map(|s| Version::parse(s).ok())
+                   .filter(|s| r.matches(&s))
+                   .max()
+                   .chain_err(|| "unable to find max Version")?;
```
budziq: I'd suggest replacing the `chain_err` with `ok_or_else` to identical effect or just leave it as an `Option` without any unwrapping and test the version_max (suggested variable name) with `Some(Version { ..... })`

Also, I'd recommend to replace the "unable to find max Version" with something along the lines of "Unable to find version satisfying given range".

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+
+    assert_eq!(vmax, Version { major: 1, minor: 0, patch: 0, pre: vec!(), build: vec!() });
+
+    Ok(())
+}
+
+quick_main!(run);
```
budziq: lets hide the error boilerplate

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given an array of unparsed version `&str`s, iterate and filter against a semver `VersionReq` and find the latest `Version`.
+
+```rust
+#[macro_use]
+extern crate error_chain;
+extern crate semver;
+
+use semver::Version;
```
budziq: lets import both `Version` and `VersionReq` with a single statement

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+    let r = VersionReq::parse("<= 1.0.0")
+        .chain_err(|| "unable to parse VersionReq")?;
+
+    let vmax = vset.iter()
+                   .filter_map(|s| Version::parse(s).ok())
+                   .filter(|s| r.matches(&s))
+                   .max()
+                   .chain_err(|| "unable to find max Version")?;
+
+    assert_eq!(vmax, Version { major: 1, minor: 0, patch: 0, pre: vec!(), build: vec!() });
```
budziq: lets run rustfmt against the example

src/app.md
```diff
@@ -725,6 +726,50 @@ fn run() -> Result<()> {
...
+}
+
+fn run() -> Result<()> {
+    let vset = ["0.9.0", "1.0.0", "1.0.1"];
+    let r = VersionReq::parse("<= 1.0.0")
+        .chain_err(|| "unable to parse VersionReq")?;
+
+    let vmax = vset.iter()
+                   .filter_map(|s| Version::parse(s).ok())
+                   .filter(|s| r.matches(&s))
```
budziq: the `&` is unnecessary here

src/app.md
budziq: Agreed. I'd remove the "an array" and "unparsed". I'd also add the hyperlinks to key identifiers like @ludwigpacifici suggests, Imho these would be `VersionReq::parse`, `VersionReq::matches` and optionally `Version::parse`.

Additionally "latest" is quite imprecise, I'd suggest rewriting it as 'satisfying the given range "<= 1.0.0"' .

src/app.md
anna-liao: I tried just removing the `chain_err` and leaving it as `.max()?;` but it results in a compiler error `the trait bound `Error: std::convert::From<std::option::NoneError>` is not satisfied`.  So I just replaced the `chain_err` with `ok_or_else` instead.

anna-liao:
Thanks for the reviews, @ludwigpacifici and @budziq !  Really appreciate you both taking the time to go over my PR with a fine tooth comb.

I am having two issues while revising the PR:
1) When I click on the link I added in `intro.md` after running `mdbook serve -o`, it jumps to the wrong example.  Specifically, it jumps to the `ex-semver-command` example.  I am pretty sure I wrote the link format correctly.  I am running this on Ubuntu 16.04
2) In the description, I can't hyperlink `VersionReq::matches` and `max`.  I tried specifying as `[`VersionReq::matches`]` and `[`semver::VersionReq::matches`]`, as well as `[`std::max`]` and `[`max`]`.

src/app.md
```diff
@@ -725,6 +726,56 @@ fn run() -> Result<()> {
...
 ```
 
+[ex-semver-latest]: #ex-semver-latest
+<a name="ex-semver-latest"></a>
+## Find the latest version satisfying given range
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given a list of version `&str`s, find the latest [`semver::Version`] that satisfies the given range
+"<= 1.0.0". A [`semver::VersionReq`] is used to compare the parsed version to a 
+minimum version requirement.  Filter for version `&str` that are verified as [`semver::Version`] with [`Version::parse`] and satisfies the [`semver::VersionReq`] using [`semver::VersionReq::matches`](https://docs.rs/semver/*/semver/struct.VersionReq.html#method.matches).  Return the latest [`semver::Version`] using [`std::cmp::max`](https://doc.rust-lang.org/std/cmp/fn.max.html).
```
ludwigpacifici: To ease readability, links are placed at the bottom of the file in alphabetic order.

No need for double space after a dot. Maybe I am too picky? :-)
- "requirement.  Filter"
- "matches).  Return"

ludwigpacifici:
> I am having an issue while revising the PR:
When I click on the link I added in intro.md after running mdbook serve -o, it jumps to the wrong example. Specifically, it jumps to the ex-semver-command example. I am pretty sure I wrote the link format correctly. I am running this on Ubuntu 16.04

Nice catch @anna-liao! I do not see something obvious in your PR that can explain that. Also, I can reproduce your issue with Firefox with different links:
- https://rust-lang-nursery.github.io/rust-cookbook/net.html#ex-extract-links-webpage
- https://rust-lang-nursery.github.io/rust-cookbook/basics.html#ex-measure-elapsed-time

With Chrome it is fine. Not sure if the issue is cookbook or mdbook related.

@budziq, in case you missed it, did you have time to look at the below (from top of this thread)? :-)

> @budziq, should the example be enriched by adding pre-release tag into the versions? If yes, the crate as a specific behaviour that is not obvious and might be worth to highlight (crate code).

budziq:
> With Chrome it is fine. Not sure if the issue is cookbook or mdbook related.

I can reproduce it only on Firefox 57 which due to stylo is more prone to incomplete style flashes. What we experience here is firefox pointing to correct page location but then a highlight.js and mdbooks custom js kicks-in to style the code snippets and hide the lines starting with '#'. The links you are mentioning are in places particularly heavy with '#'. Hiding the '#' lines causes the page to get shorter but  firefox viewport does not move anymore. I'm not versed in cross-browser compatible html+js+css so I cannot say if that this is a problem with either firefox or mdbook but it is certainly not a problem with the cookbook. I guess that we should start with posting the issue to mdbook bugtracker. Anyone care to do that? :)

>  should the example be enriched by adding pre-release tag into the versions? 

I'm not particularly sure if we should focus on teaching semver intricacies as the mentioned behavior is just part of the standard but if we were able to present it in a compact form then I'm game! If we were to go that way I'd suggest to extract the comparing/searching logic into a function and call it two times (once on a dataset that we already have and once on a version with prerelease tag)

src/app.md
```diff
@@ -725,6 +726,55 @@ fn run() -> Result<()> {
...
 ```
 
+[ex-semver-latest]: #ex-semver-latest
+<a name="ex-semver-latest"></a>
+## Find the latest version satisfying given range
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given a list of version `&str`s, find the latest [`semver::Version`] that satisfies the given range
+"<= 1.0.0". A [`semver::VersionReq`] is used to compare the parsed version to a 
+minimum version requirement. Filter for version `&str` that are verified as [`semver::Version`] with [`Version::parse`] and satisfies the [`semver::VersionReq`] using [`VersionReq::matches`]. Return the latest [`semver::Version`] using [`max`].
```
budziq: I'd probably refrain from describing all of the code step by step (especially the last sentence might be omitted).

src/app.md
```diff
@@ -725,6 +726,55 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-semver-latest]: #ex-semver-latest
+<a name="ex-semver-latest"></a>
+## Find the latest version satisfying given range
+[![semver-badge]][semver] [![cat-config-badge]][cat-config]
+
+Given a list of version `&str`s, find the latest [`semver::Version`] that satisfies the given range
+"<= 1.0.0". A [`semver::VersionReq`] is used to compare the parsed version to a
```
budziq: I'd probably try to keep the description more compact. Haw about something along these lines?:

Given a list of version `&str`s, finds the latest [`semver::Version`] that satisfying given [`semver::VersionReq`] "<= 1.0.0" using [`VersionReq::matches`].

src/app.md
anna-liao: @budziq @ludwigpacifici I updated the description so that it is concise.

Regarding an example with the pre-release tags, I am working on that in a separate branch and will submit as a PR when it is working.

ludwigpacifici:
Thanks @anna-liao for raising an mdBook issue.

> I'm not particularly sure if we should focus on teaching semver intricacies as the mentioned behavior is just part of the standard but if we were able to present it in a compact form then I'm game! If we were to go that way I'd suggest to extract the comparing/searching logic into a function and call it two times (once on a dataset that we already have and once on a version with prerelease tag) 

I raised this idea because I tried this example with these inputs `["0.9.0", "1.0.1", "0.9.0-abc"]` and I was surprised to see `0.9.0` as a result. I though it was bug. In the end I found out, it was by definition. I agree, it's a bit border line between cookbook and semver specifications. Refactor the code in a function and call it on two different examples is a good approach and not too intrusive.

anna-liao:
Added new version with pre-release example.  r? @ludwigpacifici @budziq

src/app.md
```diff
@@ -725,6 +726,80 @@ fn run() -> Result<()> {
...
+
+fn find_max_matching_version(
+    version_req_str: &str,
+    version_str_intoiter: IntoIter<&str>,
+) -> Result<Option<Version>> {
+    let vreq = VersionReq::parse(version_req_str)?;
+
+    Ok(
+        version_str_intoiter
+        //.iter()
```
ludwigpacifici: You can remove this comment

src/app.md
```diff
@@ -725,6 +726,80 @@ fn run() -> Result<()> {
...
+            ">1.2.3-alpha.3",
+            vec![
+                "1.2.3-alpha.3",
+                "1.2.3-alpha.4",
+                "1.2.3-alpha.10",
+                "1.2.3-beta.4",
+                "3.4.5-alpha.9",
+            ].into_iter()
+        )?,
+        Some(Version {
```
ludwigpacifici: Since the expected result can be surprising, I wonder if a comment should be added like: "Shows Semver precedence for pre-release tags"

src/app.md
```diff
@@ -725,6 +726,80 @@ fn run() -> Result<()> {
...
+use semver::{AlphaNumeric, Numeric, Version, VersionReq};
+use std::vec::IntoIter;
+#
+# error_chain! {
+#     foreign_links {
+#         SemVer(semver::ReqParseError);
+#     }
+# }
+
+fn find_max_matching_version(
```
budziq: If you go with signature such as this one:

```rust
fn find_max_matching_version<'a, I>(
    version_req_str: &str,
    iterable: I,
) -> Result<Option<Version>>
where
    I: IntoIterator<Item = &'a str>,
{
```

Then you will be able to omit both `into_iter()` calls below and 

also rustfmt allows us to put whole function declaration in a single line so lets do it for readability.

anna-liao: When I run the code with this signature, I get this error:

```
error[E0599]: no method named `filter_map` found for type `I` in the current scope
  --> src/main.rs:33:10
   |
33 |         .filter_map(|s| Version::parse(s).ok())
   |          ^^^^^^^^^^
   |
   = note: the method `filter_map` exists but the following trait bounds were not satisfied:
           `&mut I : std::iter::Iterator`
   = help: items from traits can only be used if the trait is implemented and in scope
   = note: the following trait defines an item `filter_map`, perhaps you need to implement it:
           candidate #1: `std::iter::Iterator`
```
I couldn't figure out how to resolve this compiler error.  Do I need to implement filter_map for I?

ludwigpacifici: @anna-liao , have you tried to remove the two below `into_iter()` calls (in the asserts) and go `iterable.into_iter().filter_map(...` in `find_max_matching_version` function?

budziq: > Do I need to implemDo I need to implement filter_map for I?

Nope. Actually the error message could be a lot better here.

Let me try to explain what is going on here. `IntoIterator` is a trait bound here making the `find_max_matching_version`  function generic over type implementing the trait. So it will accept now things like `Vec`, `HashSet` (and almost any other container) or any `Iterator` as long as their `Item` is `&str`.

But while all these types are accepted as the parameter, they are not an iterator yet (which itself implement `filter_map`). You will have to actually use the `into_iter` trait method to turn the argument into an `Iterator`.

This is how almost all conversion trait bounds are used (`AsRef<T>`,  `Into<T>`, etc). You use the trait bound to accept a range of types and use the trait method inside the function to get parameter converted into the type you want. I hope this helps :)

anna-liao: @ludwigpacifici @budziq Ah thanks for the explanation!  I have tested the code as suggested,  and updated the PR

src/app.md
```diff
@@ -725,6 +726,81 @@ fn run() -> Result<()> {
...
+
+Given a list of version &strs, finds the latest [`semver::Version`] that satisfying a given [`semver::VersionReq`] using [`VersionReq::matches`].
+
+```rust
+# #[macro_use]
+# extern crate error_chain;
+extern crate semver;
+
+use semver::{AlphaNumeric, Numeric, Version, VersionReq};
+use std::vec::IntoIter;
```
budziq: this import is unused leading to warnings.

src/app.md
```diff
@@ -725,6 +726,81 @@ fn run() -> Result<()> {
...
+            ">1.2.3-alpha.3",
+            vec![
+                "1.2.3-alpha.3",
+                "1.2.3-alpha.4",
+                "1.2.3-alpha.10",
+                "1.2.3-beta.4",
+                "3.4.5-alpha.9",
+            ]
+        )?,
+        Some(Version {
```
budziq: I'm not sure if we should not replace this multiline `Some(Version{...})` with just `Some(Version::parse("1.0.0")?)` for brevity in both cases (also reducing number of required imports). But it will not be a disaster if we leave it as it is. What are your thoughts on it @anna-liao , @ludwigpacifici ?

anna-liao:
@budziq I think that's a good idea, as it is more concise and readable.  I have updated the PR with the revision.

ludwigpacifici:
I have one question below, otherwise looks good to me!

src/app.md
```diff
@@ -725,6 +726,69 @@ fn run() -> Result<()> {
...
+use semver::{Version, VersionReq};
+#
+# error_chain! {
+#     foreign_links {
+#         SemVer(semver::SemVerError);
+#         SemVerReq(semver::ReqParseError);
+#     }
+# }
+
+fn find_max_matching_version<'a, I>(version_req_str: &str, iterable: I) -> Result<Option<Version>>
```
ludwigpacifici: The semantic of the return type `Result<Option<T>>` looks off to me. However, I have no better suggestion. Does it mean something specific?

budziq: > Does it mean something specific?

Yep. Typically `Result` is used to handle possibility of error while `Option` expresses possibility of nonexistence. So as long as nonexistence of searched value is not considered a hard error it would be best expressed as an `Option` while having to deal with parsing errors leaves us with `Result` so `Result<Option<T>>` gives the reader of our API a clear indication that the operation can both fail as well as yield no usable return value even if there is no error.

budziq:
Nicely done @anna-liao !
Thanks for the help with the review @ludwigpacifici !

src/app.md
ludwigpacifici: Thanks @budziq for your explanation. I appreciate it!

ludwigpacifici:
Well done @anna-liao !
You are welcome @budziq !

anna-liao:
Thanks @budziq and @ludwigpacifici  :)

PR #372, opened 2017-11-27
Issue287: create random password from a set of allowed characters

### LINKED ISSUE #276: Cookbook ideas for semver
Come up with ideas for nice introductory examples of using the [`semver`](https://github.com/steveklabnik/semver) crate, possibly in combination with other crates, that would be good to show in the Rust Cookbook. Please leave a comment here with your ideas! You don't necessarily have to write the example code yourself but PRs are always welcome!

- [x] [Parse and modify a version string](https://github.com/rust-lang-nursery/rust-cookbook/issues/283)
- [x] [Parse a complex version string](https://github.com/rust-lang-nursery/rust-cookbook/issues/284)
- [x] [Check if given version is pre-release](https://github.com/rust-lang-nursery/rust-cookbook/issues/316)
- [x] [Find the latest version satisfying given range](https://github.com/rust-lang-nursery/rust-cookbook/issues/285)
- [x] [Check external command version for compatibility](https://github.com/rust-lang-nursery/rust-cookbook/issues/286)

### LINKED ISSUE #287: add "Create random passwords from a set of allowed characters" example
Given a set of allowed ASCII chars, generate a set of passwords with given length.

Use [rand](https://crates.io/crates/rand) or [ring](https://crates.io/crates/ring) (the ring will most likely be a better idea although the implementation will be more involved) crate

cc https://github.com/rust-lang-nursery/rust-cookbook/issues/227

Thanks for the idea @Phaiax !

anna-liao:
fixes #287 r? @budziq 

I categorized this example under `text processing`.  Do you think that is the appropriate category?

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -8,6 +8,8 @@
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with given distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Create random passwords from a set of alphanumeric characters][ex-rand-passwd] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
+| [Create random passwords from a set of user-defined characters][ex-rand-choose] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
 | [Run an external command and process stdout][ex-parse-subprocess-output] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run an external command passing it stdin and check for an error code][ex-parse-subprocess-input] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run piped external commands][ex-run-piped-external-commands] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
@@ -274,6 +276,52 @@ fn main() {
     let rand_point: Point = rng.gen();
     println!("Random tuple: {:?}", rand_tuple);
     println!("Random Point: {:?}", rand_point);
+}
+```
+
+[ex-rand-passwd]: #ex-rand-passwd
+<a name="ex-rand-passwd"></a>
+## Create random passwords from a set of alphanumeric characters
+
+[![rand-badge]][rand] [![cat-os-badge]][cat-os]
+
+Randomly generates a string of given length ASCII characters in the range `A-Z, a-z, 0-9`, with [`gen_ascii_chars`].
+
+```rust
+extern crate rand;
+
+use rand::{thread_rng, Rng};
+
+fn main() {
+    let rand_string: String = thread_rng().gen_ascii_chars().take(30).collect();
+    println!("{}", rand_string);
+}
+```
+
+[ex-rand-choose]: #ex-rand-choose
+<a name="ex-rand-choose"></a>
+## Create random passwords from a set of user-defined characters
+
+[![rand-badge]][rand] [![cat-os-badge]][cat-os]
+
+Randomly generates a string of given length ASCII characters with custom user-defined bytestring, with [`choose`].
+
+```rust
+extern crate rand;
+
+use rand::{thread_rng, Rng};
+
+fn main() {
+    const CHARSET: &[u8] =  b"ABCDEFGHIJKLMNOPQRSTUVWXYZ\
+    abcdefghijklmnopqrstuvwxyz\
+    0123456789)(*&^%$#@!~";
+
+    let mut rng = thread_rng();
+    let password: Option<String> = (0..30)
+        .map(|_| Some(*rng.choose(CHARSET)? as char))
+        .collect();
+
+    println!("{:?}", password);
 }
 ```
 
@@ -1615,6 +1663,7 @@ fn main() {
 [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
 [`BufReader`]: https://doc.rust-lang.org/std/io/struct.BufReader.html
 [`chain_err`]: https://docs.rs/error-chain/*/error_chain/index.html#chaining-errors
+[`choose`]: https://docs.rs/rand/*/rand/trait.Rng.html#method.choose
 [`chrono::format::strftime`]: https://docs.rs/chrono/*/chrono/format/strftime/index.html
 [`Command`]: https://doc.rust-lang.org/std/process/struct.Command.html
 [`Datelike`]: https://docs.rs/chrono/*/chrono/trait.Datelike.html
@@ -1639,6 +1688,7 @@ fn main() {
 [`File::try_clone`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
 [`foreign_links`]: https://docs.rs/error-chain/*/error_chain/#foreign-links
+[`gen_ascii_chars`]: https://docs.rs/rand/*/rand/trait.Rng.html#method.gen_ascii_chars
 [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
 [`hmac::Signature`]: https://docs.rs/ring/*/ring/hmac/struct.Signature.html
 [`IndependentSample::ind_sample`]: https://doc.rust-lang.org/rand/rand/distributions/trait.IndependentSample.html#tymethod.ind_sample
--- a/src/intro.md
+++ b/src/intro.md
@@ -26,6 +26,8 @@ community. It needs and welcomes help. For details see
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with given distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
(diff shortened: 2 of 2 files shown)


src/basics.md
```diff
@@ -276,6 +277,26 @@ fn main() {
 }
 ```
 
+[ex-rand-passwd]: #ex-rand-passwd
+<a name="ex-rand-passwd"></a>
+## Create random passwords from a set of allowed characters
+
+[![rand-badge]][rand] [![cat-text-processing-badge]][cat-text-processing]
+
+Randomly generates a string of given length ASCII characters in the range `A-Z, a-z, 0-9`.
```
ludwigpacifici: Can you give more details, such as [`gen_ascii_chars`](https://docs.rs/rand/0.3.18/rand/trait.Rng.html#method.gen_ascii_chars) is the method which a hardcoded range of character.

Plus, I would show the [`choose`](https://docs.rs/rand/0.3.18/rand/trait.Rng.html#method.choose) method which is a user defined set of allowed characters. Maybe wait for the green flag of @budziq for this one.

src/basics.md
```diff
@@ -8,6 +8,7 @@
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with given distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Create random passwords from a set of allowed characters][ex-rand-passwd] | [![rand-badge]][rand] | [![cat-text-processing-badge]][cat-text-processing] |
```
budziq: I'd remove the "text-processing" category. We are not really doing any processing of input text.

src/basics.md
```diff
@@ -276,6 +277,26 @@ fn main() {
...
+
+Randomly generates a string of given length ASCII characters in the range `A-Z, a-z, 0-9`.
+
+```rust
+extern crate rand;
+
+use rand::{thread_rng, Rng};
+
+fn main() {
+    let s: String =
```
budziq: lets run rustfmt on the snippet

src/basics.md
budziq: Agreed on both arguments:
- lets add links to key identifiers in the description.
- I'd suggest to add two snippets here the second one being an iterator based in `choose` like the one used to implement `gen_ascii_chars` but taking arbitrary bytestring in the constructor, then used in identical fashion as the first trivial snippet. That would be very cool!

src/basics.md
```diff
@@ -1615,6 +1664,7 @@ fn main() {
 [`BufRead`]: https://doc.rust-lang.org/std/io/trait.BufRead.html
 [`BufReader`]: https://doc.rust-lang.org/std/io/struct.BufReader.html
 [`chain_err`]: https://docs.rs/error-chain/*/error_chain/index.html#chaining-errors
+[`choose`]: https://docs.rs/rand/0.3.18/rand/trait.Rng.html#method.choose
```
budziq: lets use wildcard version numbers

src/basics.md
```diff
@@ -1639,6 +1689,7 @@ fn main() {
 [`File::try_clone`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.try_clone
 [`File`]: https://doc.rust-lang.org/std/fs/struct.File.html
 [`foreign_links`]: https://docs.rs/error-chain/*/error_chain/#foreign-links
+[`gen_ascii_chars`]: https://docs.rs/rand/0.3.18/rand/trait.Rng.html#method.gen_ascii_chars
```
budziq: lets use wildcard version numbers

src/basics.md
```diff
@@ -277,6 +278,54 @@ fn main() {
...
+#     foreign_links {
+#         String(std::string::FromUtf8Error);
+#     }
+# }
+
+fn run() -> Result<()> {
+    let s: String = thread_rng().gen_ascii_chars().take(30).collect();
+    println!("{}", s);
+
+    let alphanum_bytestring =
```
budziq: I'd suggest implementing an iterator type like the [`AsciiGenerator`](https://docs.rs/rand/0.3.18/src/rand/lib.rs.html#670-684) instead that would take an rng in a constructor and for added merrit a "charset" (a hardcoded bytestr for our purposes without anych dynamic generation with ranges and chain) as an argument. 

Then it would be used almost identically as the `gen_ascii_chars`

src/basics.md
```diff
@@ -277,6 +278,54 @@ fn main() {
...
+        String::from_utf8((b'a'..b'z')
+        .chain(b'A'..b'Z')
+        .chain(b'0'..b'9')
+        //.chain([b'!', b'?', b'%'])
+        .collect())?;
+
+    let mut rand_bytevec: Vec<u8> = Vec::new();
+
+    for _ in 0..30 {
```
budziq: If the iterator approach would be used, pretty much all of this code would be not needed.

src/basics.md
```diff
@@ -8,6 +8,7 @@
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with given distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Create random passwords from a set of allowed characters][ex-rand-passwd] | [![rand-badge]][rand] | |
```
budziq: there should be a category here

src/intro.md
```diff
@@ -26,6 +26,7 @@ community. It needs and welcomes help. For details see
 | [Generate random numbers within a range][ex-rand-range] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random numbers with given distribution][ex-rand-dist] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
+| [Create random passwords from a set of allowed characters][ex-rand-passwd] | [![rand-badge]][rand] | |
```
budziq: there should be a category here

src/basics.md
```diff
@@ -277,6 +278,54 @@ fn main() {
...
+
+fn run() -> Result<()> {
+    let s: String = thread_rng().gen_ascii_chars().take(30).collect();
+    println!("{}", s);
+
+    let alphanum_bytestring =
+        String::from_utf8((b'a'..b'z')
+        .chain(b'A'..b'Z')
+        .chain(b'0'..b'9')
+        //.chain([b'!', b'?', b'%'])
```
budziq: It does not work because ranges give you `u8` while iterating over slice gives you `&u8` so the solution (if we wanted to go with this approach and we do not ;)) would be to clone the iterator contents.
        `.chain([b'!', b'?', b'%'].iter().cloned())` or even better  `.chain(b"!?%".iter().cloned())`.

On the other hand I would suggest to just go with bytestring literal
```
const CHARSET: &[u8] =  b"ABCDEFGHIJKLMNOPQRSTUVWXYZ\
              abcdefghijklmnopqrstuvwxyz\
              0123456789)(*&^%$#@!~";
```

anna-liao:
@budziq Thanks for the review.  I revised the PR based on your suggestions.  With the `choose` example, if I manually compose the `vec` it seems to work, but I am having issues writing an interator like for `AsciiGenerator`.  Can you provide some guidance on how to write `CustomGenerator` to work properly?

src/basics.md
```diff
@@ -277,6 +279,79 @@ fn main() {
...
+    fn next(&mut self) -> Option<char> {
+        const CHARSET: &'static [u8] =
+            b"ABCDEFGHIJKLMNOPQRSTUVWXYZ\
+              abcdefghijklmnopqrstuvwxyz\
+              0123456789)(*&^%$#@!~";
+        Some(*self.rng.choose(CHARSET).unwrap() as char)
+    }
+}
+
+fn gen_rand_chars<'a>(&'a mut self) -> CustomGenerator<'a, Self> where Self: Sized {
```
budziq: I guess that you are trying to emulate the `gen_ascii_chars` method here but it is not possible to extend the Rng in such way. All we can do is prepare a standard constructor that will take an rng and a charset.

Some along the lines of
```rust
impl<'a, R> CustomGenerator<'a, R> {
    fn new(rng: R, charset: &'a [u8]) -> Result<Self> {
        if charset.len() > 0 {
            Some(CustomGenerator {
                rng: rng,
                charset: charset,
            })
        } else {
                 Err(ErrorKind::InvalidCharset.into())
        }
    }
}
```

src/basics.md
```diff
@@ -277,6 +279,79 @@ fn main() {
...
+
+Randomly generates a string of given length ASCII characters with custom user-defined bytestring, with [`choose`].
+
+```rust
+extern crate rand;
+
+use rand::{thread_rng, Rng};
+
+struct CustomGenerator<'a, R:'a> {
+    rng: &'a mut R,
```
budziq: I'd probably go with owing the `rng` to make the lifetimes simpler. But I'll leave it up to you :)
We might go with having a refference to some outsude charset

anna-liao:
@budziq Thanks for the help!  I have learned more about idiomatic Rust syntax.  I decided to stick with the non-generator case for user-specified charset.  I couldn't get the generator case working.  I think the generator case would be good as an additional example?

src/basics.md
```diff
@@ -277,6 +279,52 @@ fn main() {
...
+use rand::{thread_rng, Rng};
+
+fn main() {
+    const CHARSET: &[u8] =  b"ABCDEFGHIJKLMNOPQRSTUVWXYZ\
+    abcdefghijklmnopqrstuvwxyz\
+    0123456789)(*&^%$#@!~";
+
+    let mut rng = thread_rng();
+    let password: String = (0..30)
+        .map(|_| *rng.choose(CHARSET).unwrap() as char)
```
budziq: we generally avoid `unwrap` in the examples, we should either document why the `unwrap` will be always ok as long as the `CHARSET` is nonempty or provide a generalized solution like:
```rust
    let password: Option<String> = (0..30)
        .map(|_| Some(*rng.choose(CHARSET)? as char))
        .collect();

    println!("{:?}", password);
```

anna-liao: @budziq Thanks for showing me how to implement the generator case!  I have updated the PR as suggested to remove the `unwrap()` and print an `Option` instead.

PR #373, opened 2017-11-27
Add "Process child process output continuously" example

### LINKED ISSUE #339: Add "Process child process output continuously" example
Please provide an example using std only that shows how to process child process output (stdout) continuously without waiting for the child to finish.

cc #266 

Original request below:

--------

In https://rust-lang-nursery.github.io/rust-cookbook/basics.html#ex-parse-subprocess-output it looks like the example waits for subprocess to finish before processing output. How to process program output in realtime similar to what `grep` does?

Also, if the reading operation is blocking and child process hangs without any output, how to implement a timeout in this situation?

ludwigpacifici:
fixes #339 

Hello @budziq ! Another PR :-) I also added a commit to update the std badge version.

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -12,6 +12,7 @@
 | [Run an external command passing it stdin and check for an error code][ex-parse-subprocess-input] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run piped external commands][ex-run-piped-external-commands] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
 | [Redirect both stdout and stderr of child process to the same file][ex-redirect-stdout-stderr-same-file] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
+| [Continuously process child process' outputs][ex-continuous-process-output] | [![std-badge]][std] | [![cat-os-badge]][cat-os][![cat-text-processing-badge]][cat-text-processing] |
 | [Filter a log file by matching multiple regular expressions][ex-regex-filter-log] | [![regex-badge]][regex] | [![cat-text-processing-badge]][cat-text-processing]
 | [Declare lazily evaluated constant][ex-lazy-constant] | [![lazy_static-badge]][lazy_static] | [![cat-caching-badge]][cat-caching] [![cat-rust-patterns-badge]][cat-rust-patterns] |
 | [Maintain global mutable state][ex-global-mut-state] | [![lazy_static-badge]][lazy_static] | [![cat-rust-patterns-badge]][cat-rust-patterns] |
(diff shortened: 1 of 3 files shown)


src/basics.md
```diff
@@ -502,6 +503,57 @@ fn run() -> Result<()> {
...
+
+    let reader = BufReader::new(stdout);
+
+    for line in reader.lines() {
+        let line = line?;
+        if line.find("usb").is_some() {
+            println!("{}", line);
+        }
+    }
+#
```
budziq: the `Ok(())` might be a boilerplate but lets not hide it as it is necessary to compile the example when copy pasted

src/basics.md
```diff
@@ -12,6 +12,7 @@
 | [Run an external command passing it stdin and check for an error code][ex-parse-subprocess-input] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
 | [Run piped external commands][ex-run-piped-external-commands] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
 | [Redirect both stdout and stderr of child process to the same file][ex-redirect-stdout-stderr-same-file] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
+| [Continuously process child process' outputs][ex-continuous-process-output] | [![std-badge]][std] | [![cat-os-badge]][cat-os] |
```
budziq: hmm text-processing badge might make sense. What are your thoughts @ludwigpacifici ?

src/basics.md
```diff
@@ -502,6 +503,57 @@ fn run() -> Result<()> {
...
+fn run() -> Result<()> {
+    let stdout = Command::new("journalctl")
+        .stdout(Stdio::piped())
+        .spawn()?
+        .stdout
+        .ok_or_else(|| "Could not capture standard output.")?;
+
+    let reader = BufReader::new(stdout);
+
+    for line in reader.lines() {
```
budziq: how about we try to express this in a functional style with `filter_map` `filter` and `for_each`?

src/basics.md
```diff
@@ -502,6 +503,57 @@ fn run() -> Result<()> {
...
+
+The below recipe is equivalent to run the Unix shell command
+`journalctl | grep usb`.
+
+```rust,no_run
+# #[macro_use]
+# extern crate error_chain;
+#
+use std::process::{Command, Stdio};
+use std::io::BufReader;
```
budziq: lets group these imports

src/basics.md
```diff
@@ -502,6 +503,57 @@ fn run() -> Result<()> {
...
+<a name="ex-continuous-process-output"></a>
+## Continuously process child process' outputs
+
+[![std-badge]][std] [![cat-os-badge]][cat-os]
+
+Contrary to [Run an external command and process
+stdout](#ex-parse-subprocess-output), while an external [`Command`] is
+running, process its standard output. A new pipe is created by
+[`Stdio::piped`] and it is read by a [`BufReader`]. Note, a buffered
+reader uses a buffer to reduce the number of I/O requests.
```
budziq: I'm not sure that we need to explain what `BufReader` does

src/basics.md
```diff
@@ -502,6 +503,57 @@ fn run() -> Result<()> {
...
 
+[ex-continuous-process-output]: #ex-continuous-process-output
+<a name="ex-continuous-process-output"></a>
+## Continuously process child process' outputs
+
+[![std-badge]][std] [![cat-os-badge]][cat-os]
+
+Contrary to [Run an external command and process
+stdout](#ex-parse-subprocess-output), while an external [`Command`] is
+running, process its standard output. A new pipe is created by
```
budziq: the explanation is slightly unclear. I'd emphasize that the we do not wait for the child process to finish and process the output continuously.

src/basics.md
ludwigpacifici: True, it's slightly off topic. While I was implementing the example, I ask myself the question `std::io::BufReader` vs `std::io::Reader`. I will remove it. Plus the [`BufReader`](https://doc.rust-lang.org/std/io/struct.BufReader.html) link to the doc explains it:

> It can be excessively inefficient to work directly with a Read instance. For example, every call to read on TcpStream results in a system call. A BufReader performs large, infrequent reads on the underlying Read and maintains an in-memory buffer of the results.

src/basics.md
ludwigpacifici: Yep. Plus it will be homogeneous with similar recipes like [Run an external command and process stdout](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#ex-parse-subprocess-output).

src/basics.md
ludwigpacifici: hmm, I remember rustfmt was doing this for me before. But not now. Anyway, let's group them.

PR #374, opened 2017-11-28
Issue285enh semver

### LINKED ISSUE #285: add "Find the latest version satisfying given range" example

anna-liao:
fixes #285 with additional pre-release example.

r? @budziq @ludwigpacifici 
This is not compiling.  I am stuck on the separate function.  Do you see what might be the issue?  When I run the code in the block comment, it works properly.

### DIFF
--- a/src/app.md
+++ b/src/app.md
@@ -17,6 +17,7 @@
 | [Parse and increment a version string][ex-semver-increment] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Parse a complex version string][ex-semver-complex] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check if given version is pre-release][ex-semver-prerelease] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
+| [Find the latest version satisfying given range][ex-semver-latest] | [![semver-badge]][semver] | [![cat-config-badge]][cat-config] |
 | [Check external command version for compatibility][ex-semver-command] | [![semver-badge]][semver] | [![cat-text-processing-badge]][cat-text-processing] [![cat-os-badge]][cat-os]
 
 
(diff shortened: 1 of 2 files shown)


ludwigpacifici:
Hello @anna-liao 

I am a bit confused. Can you tell me what's the relation between this PR and  PR #371? I was expecting to do everything in #371. Let me know if I missed something.

src/app.md
```diff
@@ -725,6 +726,119 @@ fn run() -> Result<()> {
...
+
+use semver::{Version, VersionReq, AlphaNumeric, Numeric};
+#
+# error_chain! {
+#     foreign_links {
+#         SemVer(semver::ReqParseError);
+#     }
+# }
+
+fn version_latest<'a>(
```
budziq: I'd probably try to give a more descriptive name here like `find_max_matching_version`

src/app.md
```diff
@@ -725,6 +726,119 @@ fn run() -> Result<()> {
...
+    version_req_str: &str,
+    version_str_vec: &'a Vec<&str>,
```
budziq: I'd suggest taking `IntoIter` as a parameter instead of `Vec`

src/app.md
```diff
@@ -725,6 +726,119 @@ fn run() -> Result<()> {
...
+) -> Result<Version, &'a str> {
+    let vreq = VersionReq::parse(version_req_str)?;
+
+    version_str_vec
+        .iter()
+        .filter_map(|s| Version::parse(s).ok())
+        .filter(|s| vreq.matches(s))
+        .max()
+        .ok_or_else(|| "Unable to find version satisfying given range")
```
budziq: Then you'd remove the `ok_or_else` part and wrap the returned value in `Ok()`

src/app.md
```diff
@@ -725,6 +726,119 @@ fn run() -> Result<()> {
...
+            pre: vec![],
+            build: vec![],
+        })
+    );
+
+    assert_eq!(
+        version_latest(
+            ">1.2.3-alpha.3",
+            &vec![
+                "1.2.3-alpha.3",
```
budziq: lets add some diversity here with some different prerelease tags

src/app.md
```diff
@@ -725,6 +726,119 @@ fn run() -> Result<()> {
...
+            patch: 3,
+            pre: vec![AlphaNumeric(String::from("alpha")), Numeric(10)],
+            build: vec![],
+        })
+    );
+*/
+
+    assert_eq!(
+        version_latest("<= 1.0.0", &vec!["0.9.0", "1.0.0", "1.0.1"]),
+        Ok(Version {
```
budziq: error-chain generated `Result`'s tend not to have `PartialEq` implemented so such assertion will not compile but one in which you'd use `?` after the function call will.

anna-liao:
@budziq wow, thanks for the tips for fixing my code!  There's no way I'd figure that out on my own.  Ok, I'm closing this PR and moving back to the original PR.

PR #385, opened 2018-01-15
fix C++ linking example

jrakow:
This is a trivial fix.

### Things to check before submitting a PR

- [x] the tests are passing locally with `cargo test`
- [x] cookbook renders correctly in `mdbook serve -o`
- [x] commits are squashed into one and rebased to latest master
- [x] non rendered items are in sorted order (links, reference, identifiers, Cargo.toml)
- [x] links to docs.rs have wildcard version `https://docs.rs/tar/*/tar/struct.Entry.html`
- [x] example has standard [error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)
- [x] code identifiers in description are in hyperlinked backticks 
```markdown
[`Entry::unpack`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack
```

### Things to do after submitting PR
- [x] check if CI is happy with your PR
- [x] drop a comment on https://github.com/rust-lang-nursery/rust-cookbook/issues/209 if you consent to repository being relicensed with [CC0 license](https://creativecommons.org/choose/zero/) :+1:  

Thank you for reading, you may now delete this text! Thank you! :smile:

PR #389, opened 2018-01-25
Added function input and return types to C++ library example so it will

### LINKED ISSUE #388: Compile and link statically do a bundled C++ library does not compile
[This example](https://rust-lang-nursery.github.io/rust-cookbook/build_tools.html#ex-cc-static-bundled-cpp) doesn't compile currently.

The function inputs in [`foo.cpp`](https://rust-lang-nursery.github.io/rust-cookbook/build_tools.html#srcfoocpp) aren't specified, and the return type needs to be added to the `extern` in [`main.rs`](https://rust-lang-nursery.github.io/rust-cookbook/build_tools.html#srcmainrs-1).

Right now the cpp code an error saying `x` and `y` were not declared.
```
cargo:warning=src/foo.cpp: In function ‘int multiply()’:
cargo:warning=src/foo.cpp:7:12: error: ‘x’ was not declared in this scope
cargo:warning=     return x * y;
cargo:warning=            ^
cargo:warning=src/foo.cpp:7:16: error: ‘y’ was not declared in this scope
cargo:warning=     return x * y;
cargo:warning=                ^
exit code: 1
```
The rust side fails with
```
error[E0277]: the trait bound `(): std::fmt::Display` is not satisfied
 --> src/main.rs:7:24
  |
7 |         println!("{}", multiply(5, 7));
  |                        ^^^^^^^^^^^^^^ `()` cannot be formatted with the default formatter; try using `:?` instead if you are using a format string
  |
  = help: the trait `std::fmt::Display` is not implemented for `()`
  = note: required by `std::fmt::Display::fmt`
```
as the return type is currently `()`

Fierthraix:
Explicitly added return types and function inputs in example.

fixes #388

### DIFF
--- a/src/build_tools.md
+++ b/src/build_tools.md
@@ -152,7 +152,7 @@ extern "C" {
     int multiply(int x, int y);
 }
 
-int multiply() {
+int multiply(int x, int y) {
     return x*y;
 }
 ```
@@ -161,7 +161,7 @@ int multiply() {
 
 ```rust,ignore
 extern {
-    fn multiply(x : i32, y : i32);
+    fn multiply(x : i32, y : i32) -> i32;
 }
 
 fn main(){


jrakow:
Duplicate of #385 
😃

Fierthraix:
Sorry! I just checked the issues first without looking at the pull requests!

Fierthraix:
Our code is identical, but yours didn't pass the tests for some reason.
Is it possible to rerun the integration tests?

AndyGauge:
Closing as duplicate

PR #404, opened 2018-04-26
Restructure Cookbook

### LINKED ISSUE #139: Remove passive voice
I'm not satisfied with the passive voice convention that has developed for example descriptions. I find it backwards and harder to read than I would like.

Without meaning to pick on any individual instance, let me use this one as an example:

> HTTP POST request to gists API v3 is made with reqwest::Client in order to create a gist. A request body is created with serde_json::json! macro and set set with RequestBuilder::json. Request is prepared with Client::post, authenticated with RequestBuilder::basic_auth and synchronously executed with RequestBuilder::send.
>
> Gist is subsequently deleted with HTTP DELETE request prepared with Client::delete and executed as before.

I don't have a strong preference between "make" and "makes" but either way I think this would be more readable without passive voice "is made." The rustdoc convention is "makes" tense so here is what that would look like:

> Makes a POST request to GitHub's gist API to create a new gist. Then makes a DELETE request to delete the gist with the id that was just created.
>
> The reqwest::Client is responsible for setting up details of both requests including the endpoint URL, the authentication, and the request body. Here the POST body comes from serde_json's `json!` macro which provides a way to pass an arbitrary JSON body, but elsewhere a struct may be more convenient (link to Serialize example). The DELETE request requires no body.
>
> The call to RequestBuilder::send synchronously executes the HTTP request and returns the GitHub server's response.

Any opinion @budziq? I know you were an advocate of the passive voice.

### LINKED ISSUE #280: How do we keep tables in sync
Currently the tables have to be kept in sync:
- intro.md with rest of the chapters
- the chapter tables themselves with the actual examples.

This is error prone and tedious. Most likely the tables should be generated and included into the files. This might be a job for a dedicated function in mdBook or a pre-build script.

### LINKED ISSUE #375: Reorganizing the cookbook
@aturon and I have been discussing the cookbook over the past few weeks. One of the main things we've talked about is organization and focus, to make it easy for users to find the recipes they're looking for.

An idea that I believe was @aturon's originally was to organize it along the same lines as [crates.io categories](https://crates.io/categories/), so that they can cross reference each other. The goal would be to grow toward a 1:1 correspondence between categories and sections in the cookbook (some categories may not be appropriate for sections, we'll see in the long term).

Additionally, we had some concern about the *specificity* of some recipes we have right now. We think recipe headers should be written in a way that emphasizes the main learning for the recipe, so that users can find the recipe that solves their problem. The recipe might, in fact, demonstrate how to do some concrete thing, but the header would not

For example, the recipe would be called "spawn a thread." Ideally, it would do a concretely useful thing with the thread it spawned, but the header wouldn't be "do X on a new thread," just "spawn a thread."

Steps to this reorganization (which can be partially parallelized):

- We need to finish the skeleton of the new organization for the cookbook.
- We need to mold the existing content into the new skeleton.
- We need to fill out the missing sections.

Probably the next step if people are enthusiastic about this direction is to create tracking issues for each category where the skeleton for that section can live.

### LINKED ISSUE #408: broken links
There are currently at ~32 broken links in the cookbook. Tested with ruby `html-proofer` script.

This means that either:
- the descriptions have wrong links
- are outdated leading to outdated API's
- some crates docs have been broken (most likely `rand` in this case)

```bash
gem install html-proofer && htmlproofer --empty-alt-ignore ./book/ 
```
```log
- ./book/basics.html
  *  External link https://doc.rust-lang.org/rand/rand/distributions/index.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/distributions/normal/struct.Normal.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/distributions/range/struct.Range.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/distributions/trait.IndependentSample.html#tymethod.ind_sample failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/fn.thread_rng.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/trait.Rand.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/trait.Rng.html failed: 404 No error
  *  External link https://doc.rust-lang.org/rand/rand/trait.Rng.html#method.gen_range failed: 404 No error
  *  External link https://doc.rust-lang.org/regex/regex/struct.Regex.html failed: 404 No error
  *  External link https://doc.rust-lang.org/regex/regex/struct.Regex.html#method.captures_iter failed: 404 No error
  *  External link https://doc.rust-lang.org/regex/regex/struct.RegexSet.html failed: 404 No error
  *  External link https://doc.rust-lang.org/regex/regex/struct.RegexSetBuilder.html failed: 404 No error
  *  External link https://docs.rs/rand/ failed: 404 No error
  *  External link https://docs.rs/rand/*/rand/trait.Rng.html#method.choose failed: 404 No error
  *  External link https://docs.rs/rand/*/rand/trait.Rng.html#method.gen_ascii_chars failed: 404 No error
- ./book/encoding.html
  *  External link https://docs.rs/url/*/url/percent_encoding/fn.percent_decode.html failed: 404 No error
  *  External link https://docs.rs/url/*/url/percent_encoding/fn.utf8_percent_encode.html failed: 404 No error
- ./book/logging.html
  *  External link https://doc.rust-lang.org/log/env_logger/#enabling-logging failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/enum.Target.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/fn.init.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/struct.Builder.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/struct.Builder.html#method.target failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/struct.LogBuilder.html#method.format failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/struct.LogBuilder.html#method.init failed: 404 No error
  *  External link https://doc.rust-lang.org/log/env_logger/struct.LogBuilder.html#method.parse failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/enum.LogLevel.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/enum.LogLevelFilter.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/fn.set_logger.html failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/struct.LogRecord.html#method.args failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/struct.LogRecord.html#method.level failed: 404 No error
  *  External link https://doc.rust-lang.org/log/log/trait.Log.html failed: 404 No error
- ./book/print.html
  *  External link https://docs.rs/reqwest/*/reqwest/header/struct.ContentRange.htm failed: 404 No error
htmlproofer 3.8.0 | Error:  HTML-Proofer found 32 failures!
```

### LINKED ISSUE #421: Update Contributing and PR templates
I'd like to land #410 then update the Contributing an PR templates.

AndyGauge:
See it [here](http://www.yetanother.site/rust-cookbook/)

Fixes #375 
Fixes #280 
Fixes #139 
Fixes #408 
Fixes #421

```link-checker ./book/```
Summary: seen:5735 failed:0 unique:877

### CODE REVIEW

**Comment by aturon** [comment]
btw, if you ever need reviewing, feel free to ping me on gitter!

**Comment by budziq** [comment]
@AndyGauge this starts to look awesome.

Just few thoughts.

- rendering on http://www.yetanother.site/rust-cookbook/ is a little wonky (missing icons etc. what version of mdbook did you use?)
- there are some broken links and tables here and there. this leads me back to an idea to incorporate more lints to travis (link checker, markdown lint and aspell) so we do not need to spend any more time on rendering and link veriffication after each PR (especually major ones like this one)
- do you have any thoughts where do we put tables (do webkeep them at all, split to smaller and smaller chunks)? The oroginal ones are not intended for this level of nesting and keeping them in sync will be even bigger chore than before. Maybe lets generate them on the fly?
- I'm not sure if using {include} this much by hand will not make the maintenance and contribution more error prone (maybe we could just add each example to the SUMMARY.md directly or propose some changes to mdBook itself).

I'm really interested on your thoughts!

**Comment by AndyGauge** [comment]
@budziq Thank you for the kind words.

I'm using https://github.com/rust-lang-nursery/mdBook/pull/682 as mdBook.  It is a work in progress.  The lastest version was before I incorporated relative path checking.  

I haven't addressed the brokeness inside of the old categories.  I"m trying to address dead links as I go, but this PR is far from finished.  I expect to do a final manual pass once it is complete to address the broken tables and links.  I am all for link verification post processing, but not blocking progress forward.

I'm hoping to generate tables on the fly eventually.  If we keep mdBook I'm happy to write a TOC generator.

I'm hoping this much include syntax will be replaced eventually with a generator, but the idea is to break it up first.  I envision a layout file, either TOML or YAML that generates everything.

Again, I do appreciate the thoughts.  Last few days have been hacking away at mdBook and not the cookbook.

**Comment by budziq** [comment]
I tottally agree.

I'll try to find some time to add linkcheckers and linters to CI next week.  It has been bugging me for some time already. 

The layout file sounds interesting please feel free to make a braindump with the details somewhere (possibly a separate issue in the repo might be a good place for discussion and getting help from contributors)

**Comment by AndyGauge** [comment]
@budziq  btw: https://github.com/Michael-F-Bryan/mdbook-linkcheck

**Comment by budziq** [comment]
Very cool! Thanks!

**Comment by ludwigpacifici** [comment]
FYI, #356 is open for checking links

**Comment by budziq** [comment]
@ludwigpacifici Yep, I've opened it a long while ago and it has been nagging me since then ;) I hope to find some time to address it mid next week.

**Comment by AndyGauge** [comment]
I'm thinking about redoing the includes so I can release this on mdbook 0.1.7

File: src/web/clients/api/rate-limited.md [hunk 1]
```diff
@@ -69,5 +69,5 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
-[`hyper::header!`]: https://docs.rs/hyper/*/hyper/macro.header.html
+[`hyper::header!`]: https://doc.servo.org/hyper/header/index.html#defining-custom-headers
```
AndyGauge: This basically just references the 0.11 branch without breaking the wildcard rule.  Not ideal, but #419 should solve it.

aturon:
@AndyGauge This is a fantastic step forward! Just a few notes going through:

- Concurrency: the "threads" vs "parallel tasks" is slightly confusing -- perhaps "explicit threads" vs "data parallelism" or something would be better?

- Data structures
  - I'm not sure that lazy_static should go here -- at least for me, it's not super-connected to data structures. It's a really tricky one. 
  - I didn't quite get the "custom" subcategory. Maybe just "bit manipulation"? (Are these subcategories also trying to map onto crates.io?)

- Date/time
  - The "Duration" category seems to include more than just duration-related stuff, e.g. the "examine the date and time" example is not about durations.
  - The "convert a local time ..." example isn't about parsing/displaying, I think -- more like timezone calculation?

- Development tools
  - I don't think "error handling" should go here; development tools are more about the development process itself (like debugging), whereas error handling is part of the actual application logic.

- Encoding
  - The "strings" category feels a little too generic -- I'd have a hard time guessing what was in it. Similarly with "complex data types"
  - The relationship with "text processing" is a bit tricky, too

- Web programming
  - I don't think "scraping web pages" should be a category -- better to break down into finer-grained pieces
  - Should use "URL" rather than "Uniform Resource Location" IMO; it's a ubiquitous acronym
  - For "Clients", the "Using APIs" category name is a bit unclear

AndyGauge:
@aturon I had a tough time fitting all the recipes into categories--I may just need to create memory management and move lazy static there.  Some of the series that I created were meant to divide the recipes in half to make them easier to digest.  Sometimes the categories didn't fit the recipes.  In Time/date, it probably makes sense to make a third series.  

Can we petition crates.io to make an error handling category?  Since we can't stuff it anywhere else in the cookbook I'll create a new root category for it.

AndyGauge:
@aturon I've worked through each of the comments, though some of them don't have great solutions.  Did you want to review the updates?

sandeep-datta:
Please consider the following changes ...

1. Development tools does not seem to be the right place for Versioning. I'd rather see a separate category called change management (perhaps under development tools).

2. Rename Encoding to Encoding / Decoding. Though technically correct placement of CSV processing and Structured data seems to be a little awkward. Perhaps create a separate category called Structured Data?

AndyGauge:
@sandeep-datta Thank you for your review!  

I actually do think Development tools is the right place for Versioning.  I feel that change management might refer to state machines and not the development process.  I'm wrong a lot though.

I certainly could see Encoding being renamed to Encoding / Decoding.  I'll probably do that.  We are trying to align the cookbook with the crates.io categories: https://crates.io/categories
I'd like to avoid creating extra categories.

sandeep-datta:
I have never heard of change management being used in that sense but I think I should have said [Configuration Management](https://en.wikipedia.org/wiki/Configuration_management) in the first place.

>  I'd like to avoid creating extra categories.

Well this is unfortunate but understandable. You may be already doing this but if I were you I'd make sure to look at the crates already using the chosen categories. If you were to look at the crates with "version" in their name you'd notice that most of them could have used a more appropriate category. I, for one, don't believe the list of categories on crates.io is complete by any means. But you be the judge. 

Thanks again for all the hard work that you have put into this PR regardless of what you decide.

PR #406, opened 2018-04-29
Add example of MIME from string

### LINKED ISSUE #398: Recipe for Mime::FromStrError
Show an invalid mime string, and how to handle processing the error.

math-foo:
fixes #398


src/encoding.md
```diff
@@ -688,6 +689,53 @@ fn run() -> Result<()> {
...
+[ex-mime-from-string]: #ex-mime-from-string
+<a name="ex-mime-from-string"></a>
+## Get MIME type from string
+
+[![mime-badge]][mime] [![cat-encoding-badge]][cat-encoding]
+
+The following example shows how to parse a [`MIME`] type from a string using the [mime] crate. You can handle a possible [`FromStrError`] by providing a default [`MIME`] type in an `unwrap_or` clause.
+
+```rust
+# #[macro_use]
```
AndyGauge: In this example there is no error handling (you are demonstrating how to explicitly handle the error in fact) so there is no need for the error_chain crate in this example.  You can remove lines 701-703,705-711, 733,736.

src/encoding.md
```diff
@@ -688,6 +689,53 @@ fn run() -> Result<()> {
...
+extern crate mime;
+use mime::{Mime, APPLICATION_OCTET_STREAM};
+#
+# error_chain! {
+#     foreign_links {
+#         IOError(std::io::Error);
+#    }
+# }
+
+fn run() -> Result<()> {
```
AndyGauge: with error chain gone, this becomes main()

math-foo:
Thank you for the feedback. This is much simpler now!

PR #407, opened 2018-05-02
Add recipe "Parse the MIME type of a HTTP response"

### LINKED ISSUE #397: Make HTTP request and parse mime type
In this example, make a request to a static url and parse the mime type response.  Potentially show 2 mime types and an request without a mime header.

NiekGr:
This recipe shows how the media type of a response, received with
reqwest, can be obtained from it's headers. The media type is then
matched with predefined constants from the mime crate.

fixes #397


NiekGr:
Thank you @AndyGauge.

PR #410, opened 2018-05-09
VIP support for additional linters (html-proofer link checker)

### LINKED ISSUE #356: Have a link checker in the CI
We seriously need a link checker with following properties:
- [ ] check if internal links are rendered correctly (it has to [understand markdown](https://crates.io/crates/pulldown-cmark) )
- [ ] understand the mdbook's `{{include links.md}}` syntax
- [ ] **optionally** makes requests to the actual links testing for errors (this would not be used in CI as all the environments of that type forbid networked requests)
- [ ] standalone cli tool receiving list of files or directory as an input or be a part of the mdbook itself https://github.com/azerupi/mdBook/issues/356
- [ ] ideally it would be incorporated to our CI (this is not a must for the first prototype)

### LINKED ISSUE #358: Have a spellchecker in CI
Yep we really need it

Possibly steal the ideas from TRPL ( [script](https://github.com/rust-lang/book/blob/master/second-edition/spellcheck.sh) + [dict](https://github.com/rust-lang/book/blob/master/second-edition/dictionary.txt) )

Unfortunately they are APACHE licensed while we plan to be CC0 making the direct port a little tricky

Otherwise we might consider using https://crates.io/crates/ispell

budziq:
This is very much work in progress so please do not merge for now.

- added additional test matrix items just for linters
- install and build mdbook only when needed to speedup CI
- reorganized the travis scripts
- deployment is run from linter build
- linkcheck build that is likely to have spurious failures - is currently set as "allow_failure" to not block builds
- added spellchecker to the ci
- moved link checking to https://github.com/budziq/link-checker.
    - local link problems are a hard error
    - external link arrors are just printed for informational purposes (there are false positives on docs.rs sites chrono and same_file which have javascript redirects)

TODO:
- [ ] <s>update contributing and PR template with information about the linters workflow</s>
- [x] find a link checker that works with nested mdBook directory output (or fix mdBook)

fixes #356 
fixes #358


ci/dictionary.txt
```diff
@@ -0,0 +1,327 @@
+personal_ws-1.1 en 0 utf-8
```
AndyGauge: This file looks like it will become a challenge to maintain.  What is the plan to keep this up to date?

budziq: Yeah I know that this is not ideal.
This is actually how rust-book manages spellchecking in their CI. 
A contributor would have to add missing words as lines to somewgere in the dictionary.txt . The spellchecker already uses aspell-en dictionary so most English vocabulary is already present and some technical lingo and variable names still need to be added.

budziq:
> Doesn't Travis prevent you from making external calls?

Surprisingly it does not! This approach to CI linkchecking seams to be fairly common with jeckyl blogs. The only problem is that the communication is not fully deterministic and large number of concurrent requests also causes spurious failures (html-proofer does not seam to have an easy way to throttle the connection count).

budziq:
I'm relatively happy with the setup now.
- Spellchecking is mandatory via separate Travis CI build target. It should not be that hard to maintain. Although we would need update the contributing and PR template to mention it.
- I'm not totally satisfied with Link checking yet. It is made optional (its spurious failures will not cause whole CI build to fail). But it is easily accessible by going into given Travis build. 

Do you have any thoughts @AndyGauge?

AndyGauge:
the spell checker works pretty good locally.  I would suggest adding a message explaining how to add entries to the dictionary that displays in CI if there's a failure.

htmlproofer found 2556 dead links in my reorganize branch, and I'm trying to figure out what is going wrong.  It has to do with relative paths, so I'd like to figure that out.

budziq:
> EDIT The only workaround I could find for the spurious errors was --external-only

Hmm it looks like the html-proofer does not handle the internal sidebar links of html files in subdirectories (only these exhibit problems mentioned by you). The sidebar links are relative to the index.html location while the html-proofer seams to assume that these would be relative to the origin html location
so links in rust-cookbook/web/mime.html pointing to rust-cookbook/about.html become links to rust-cookbook/web/about.html and links to rust-cookbook/web/mime.html from the web dir turn to rust- cookbook/web/web/mime.html and so on.

I do not experience the problem due to flat structure of the original cookbook.

On the other side using `--external-only` option causes us to miss broken internal links like the one to
complex.html#serialize-and-deserialize-unstructured-json in index.html, intro.html and encoding.html in our build.

We probably need to choose another linkchecker tool.
--------
By the way I love hov the reorganization work starts to look like :heart: ! @AndyGauge

budziq:
It looks like the mdBook uses   ```<base href="../">``` which link checkers do not follow. I'm experimenting with writing my own link checker script based on https://github.com/mike42/mdcheckr

AndyGauge:
That does look like the problem, but I'm seeing this https://github.com/gjtorikian/html-proofer/pull/358

Let me do some experimentation to see if html proofer just needs a PR to fix the issue.

budziq:
I've written a small link checker in python https://github.com/budziq/link-checker that seams to work ok on restructured cookbook (supports ```<base>``` and internal anchors which does not seam to be the case for most other linkcheckers)

budziq:
@AndyGauge the linkchecker I've written works fairly well now and should have nonfalse positives for local links https://github.com/budziq/link-checker

I'll try to find some time next week to workout the remaining kinks and integrate it into this PR. I guess I'll split the linkchecking tontwo phases 
- local links only - which should be reliable and can fail CI
- remote links - that cannot be made fully reliable and would be only printed for informational purposes.

budziq:
@AndyGauge if youbhave any outputbfrom instalation process then please post it as an issuebto the link-checker repo. It might be error with packaging. But most likely you have installed itbas a user and nit system wide and the location of user installed pip executables is nit addedbto your PATH

AndyGauge:
I was able to get it to install with a few hoops.  It was my first pip install so some of the workflow wasn't obvious.  I was able to get all the dead links resolved in `reorganize`.  I like the link-checker that you made, it even catches bad named anchors and redirection.  

Do you want me to do the updates to the template/Contributing to support this?

budziq:
> I like the link-checker that you made, it even catches bad named anchors and redirection.

:slightly_smiling_face: 

> Do you want me to do the updates to the template/Contributing to support this?

That would be awesome. I'm really low on time these days :/

PR #411, opened 2018-05-13
Fix broken links

dominikWin:
cc #408


budziq:
Nicely done @dominikWin!

PR #412, opened 2018-05-14
bumped crate versions

budziq:
`csv`, `hyper`, `regex`, `rayon`, `image`, `log`, `log4rs`, `env_logger`

also fixed broken links along the way


AndyGauge:
I'm glad you took this on.  I started overzealously upgrading everything when regex 1.0 came out.  Ran into a ton of test failures, and never got the scope right.  I think this is a great scope of work.  I'm especially happy csv is past beta.

budziq:
> I'm glad you took this on. 

Thanks, I try to do this from time to time. It's not that painfull once you employ `sed`.  

>  I'm especially happy csv is past beta.

Yep it was bugging me to no end too :)

PR #417, opened 2018-05-18
Fix broken rand links

dominikWin:
cc #408


budziq:
@dominikWin thanks! 

I see that these still point to original domain  docs.rust-lang.org instead of new docs.rs. Could you correct that as well for consistency sake ?

budziq:
Thanks!

PR #418, opened 2018-06-01
fix rand crate example and broken links

yogesum:
Since `rand` v0.5.0 some APIs are deprecated and some [examples](https://rust-lang-nursery.github.io/rust-cookbook/basics.html#generate-random-values-of-a-custom-type) are broken.

Updated failing rust programs and one inside [concurrency](https://rust-lang-nursery.github.io/rust-cookbook/concurrency.html#sort-a-vector-in-parallel) section (using `thread_rng`, don't know any other way to replace `weak_rng`)


budziq:
Thanks @yogesum It looks very good!

PR #422, opened 2018-07-01
Add an example to read an environment variable

### LINKED ISSUE #401: Open Config file from given environment variable location
|  |  |
| Crates  | std  |
| Section  | 15.1 External Command  |

Read environment variable to find a config file, open the config file and demonstrate reading contents.  No need to parse the configuration in this example, but highlight the environment variable and how to pass it to the process.  Looking to highlight `std::env`.  Optionally demonstrate failing back to default value.

Dineshs91:
fixes #401

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -38,7 +38,8 @@
 | [Parse string into DateTime struct][ex-parse-datetime] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Perform checked date and time calculations][ex-datetime-arithmetic] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
 | [Examine the date and time][ex-examine-date-and-time] | [![chrono-badge]][chrono] | [![cat-date-and-time-badge]][cat-date-and-time] |
-| [File names that have been modified in the last 24 hours for the working directory][ex-file-24-hours-modified] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] [![cat-os-badge]][cat-os] |
+| [File names that have been modified in the last 24 hours for the working directory][ex-file-24-hours-modified] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] [![cat-os-badge]][cat-os] | 
+| [Read environment variable][ex-read-environment-variable] | [![std-badge]][std] | [![cat-filesystem-badge]][cat-filesystem] |
 
 [ex-std-read-lines]: #ex-std-read-lines
 <a name="ex-std-read-lines"></a>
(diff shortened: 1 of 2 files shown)


AndyGauge:
Thanks for the contribution.  The AppVeyor might be issues with the new Rust version on Windows.  Let me dig into that for you to make sure master is working as the errors seem to be unrelated to the code change.

src/basics.md
```diff
@@ -1770,6 +1771,69 @@ fn run() -> Result<()> {
...
+
+fn create_sample_config(path: &PathBuf) -> std::io::Result<()> {
+    let mut output = File::create(path)?;
+    write!(output, "Sample content")?;
+
+    Ok(())
+}
+
+fn set_env_variable() {
+  env::set_var("CONFIG_FILE", "/tmp/config.toml");
```
AndyGauge: This is not a cross platform solution.  I'm looking into the way the cookbook wants to recommend implementing file system abstractions.  If you have any ideas, I'm interested.  Right now I'm reading https://github.com/rust-lang-nursery/cli-wg/issues/10

AndyGauge:
OK, I dug into it and there is a symptom of new regression due to rust version update, but that is not causing the error in AppVeyor.

Dineshs91:
@AndyGauge Thanks for the quick response. I will update the code to work on windows as well.

Dineshs91:
I've updated the code and removed all the file operations. I will squash the commits once everything is fine with this PR.

AndyGauge:
I forgot to mention putting in a few lines about the code, but you took care of that already!  Thanks.

Dineshs91:
@AndyGauge I've squashed the commits.

src/basics.md
```diff
@@ -1770,6 +1771,48 @@ fn run() -> Result<()> {
 # quick_main!(run);
 ```
 
+[ex-read-environment-variable]: #ex-read-environment-variable
+<a name="ex-read-environment-variable"></a>
+## Read environment variable
+[![std-badge]][std] [![cat-filesystem-badge]][cat-filesystem]
+
+Set an environment variable and read it back. If the variable is not found, then fallback
```
budziq: I'd suggest to link to std documentation for `std::env::set_var` and `std::env::var`

src/basics.md
```diff
@@ -1770,6 +1771,48 @@ fn run() -> Result<()> {
...
+use std::path::PathBuf;
+
+# error_chain! {
+#     foreign_links {
+#         Utf8(std::str::Utf8Error);
+#         AddrParse(std::net::AddrParseError);
+#     }
+# }
+
+fn set_env_variable() {
```
budziq: I'm not really sold on using separate function to wrap one liner here.
Also the duplication of the "config.toml" string in two places might be slightly confusing how about we use different values?

src/basics.md
```diff
@@ -1770,6 +1771,48 @@ fn run() -> Result<()> {
...
+
+```rust
+# #[macro_use]
+# extern crate error_chain;
+use std::env;
+use std::path::PathBuf;
+
+# error_chain! {
+#     foreign_links {
+#         Utf8(std::str::Utf8Error);
```
budziq: Neither of these are needed for this example

src/basics.md
```diff
@@ -1770,6 +1771,48 @@ fn run() -> Result<()> {
...
+fn set_env_variable() {
+  env::set_var("CONFIG_FILE", "config.toml");
+}
+
+fn run() -> Result<()> {
+    set_env_variable();
+
+    let default_config_file = "config.toml";
+
+    let config_file_path = match env::var("CONFIG_FILE") {
```
budziq: This might a matter of taste but how about rewriting this simple transformation from match to `Result` combinators with something like: 
```rust
    let config_file_path = env::var("CONFIG_FILE")
        .map(PathBuf::from)
        .unwrap_or_else(|_| default_config_file.into());
```

src/basics.md
```diff
@@ -1770,6 +1771,48 @@ fn run() -> Result<()> {
...
+#         AddrParse(std::net::AddrParseError);
+#     }
+# }
+
+fn set_env_variable() {
+  env::set_var("CONFIG_FILE", "config.toml");
+}
+
+fn run() -> Result<()> {
+    set_env_variable();
```
budziq: setting and reading the same env variable does not make a control flow that is close to being realistic. How about we set one variable and read another?

AndyGauge:
@Dineshs91  would be happy to help you get the changes made to complete this.  Please let me know if you are in need of support.

PR #424, opened 2018-07-06
Adding examples for deriving Debug and Display.

### LINKED ISSUE #402: Recipe: Implement Display, Debug for custom type
Given a simple custom type, implement debug and display directly.  Do not lean on derive.  Similar to this: https://doc.rust-lang.org/1.6.0/std/fmt/trait.Display.html#examples 

It is OK for the cookbook to have similar examples to other sources, as this may be the first place someone looks to solve this problem.  Do not use an x,y point as the example though.  Mixed data types would be ideal.

plauche:
fixes #402

### DIFF
--- a/src/basics.md
+++ b/src/basics.md
@@ -10,6 +10,8 @@
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Create random passwords from a set of alphanumeric characters][ex-rand-passwd] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
 | [Create random passwords from a set of user-defined characters][ex-rand-choose] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
+| [Implement Debug for a custom type][ex-impl-debug]                       | [![std-badge]][std]         |  [![cat-os-badge]][cat]                                                       |
+| [Implement Display for a custom type][ex-impl-display]      | [![std-badge]][std]         |  [![cat-os-badge]][cat-os]                                                    |
 | [Run an external command and process stdout][ex-parse-subprocess-output] | [![regex-badge]][regex] | [![cat-os-badge]][cat-os] [![cat-text-processing-badge]][cat-text-processing] |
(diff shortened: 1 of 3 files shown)


AndyGauge:
I really like the custom enum option.  A recent addition to the Rust Cookbook is a spell checker, which can be a pain to use with all the variable names, types, etc.  The CI failures need words added to the dictionary to pass.  Unfortunately, I haven't merged the changes to Contributing file that supports this.  Here's that file: https://github.com/AndyGauge/rust-cookbook/blob/reorganize/CONTRIBUTING.md#linters

budziq:
Thanks for contributing @plauche !

I'm not sure that we really need two separate recipes for procedures that are essentially identical (implementation of `Debug` and `Display` traits). how about we just limit the example to one of the traits with mention of the other in the textual description? I'd also suggest to mention that this is an alternative to just using `#[derive(Debug)]`which gives implementer control of the output.

What are your thoughts @AndyGauge ?

AndyGauge:
I agree with what @budziq said about there being a single recipe.  What I had originally thought of was a `#[derive(Debug)]` and a Display implementation in a single recipe, with textual description that explains the difference.

AndyGauge:
I've made this harder for you, if you need help let me know.

The organization has changed dramatically, so merging master will really mess with the file layout.  I'd love for this example to live in Text Processing.

AndyGauge:
@plauche Do you want me to make the changes requested?

src/basics.md
```diff
@@ -1770,6 +1772,110 @@ fn run() -> Result<()> {
...
+            FileType::Image {
+                format,
+                width,
+                height,
+            } => write!(
+                f,
+                "Image {{ format: {}, width: {}, height: {} }}",
+                format, width, height
+            ),
+            FileType::Text { format } => write!(f, "Text {{ format: {} }}", format),
```
KodrAus: We actually don't need to manually write out the `Text { format: value }` struct-like format here. There are methods on the `fmt::Formatter` like [`debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) that will do this for us and keep it consistent with what you get from `#[derive(Debug)]`.

Issue #443, opened 2018-08-20
1.30 Serde Test failues


AndyGauge:
```
error[E0658]: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
  --> /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/log4rs-0.8.0/src/append/file.rs:25:1
   |
25 | #[serde(deny_unknown_fields)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add #![feature(custom_attribute)] to the crate attributes to enable

error[E0658]: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
  --> /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/log4rs-0.8.0/src/append/console.rs:24:1
   |
24 | #[serde(deny_unknown_fields)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add #![feature(custom_attribute)] to the crate attributes to enable

error[E0658]: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
  --> /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/log4rs-0.8.0/src/append/rolling_file/policy/compound/roll/delete.rs:16:1
   |
16 | #[serde(deny_unknown_fields)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add #![feature(custom_attribute)] to the crate attributes to enable

error[E0658]: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
  --> /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/log4rs-0.8.0/src/append/rolling_file/policy/compound/roll/fixed_window.rs:17:1
   |
17 | #[serde(deny_unknown_fields)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add #![feature(custom_attribute)] to the crate attributes to enable

error[E0658]: The attribute `serde` is currently unknown to the compiler and may have meaning added to it in the future (see issue #29642)
  --> /home/andy/.cargo/registry/src/github.com-1ecc6299db9ec823/log4rs-0.8.0/src/append/rolling_file/policy/compound/trigger/size.rs:21:1
[...]

PR #444, opened 2018-08-20
Recipe Series: Sort Vector #437

### LINKED ISSUE #437: Recipe Series: Sort Vector
|  |  |
| Crates  | std  |
| Section  | 1.1 Sorting  |

This recipe series should demonstrate how sorting a vector is accomplished.  The initial recipe is a simple `std` implementation of sorting integers, and descending.  Next show how to sort `f64`. Implement a custom struct and demonstrate sorting a vector of these types.

Januson:
* Add simple sorting vector of integers
* Add sorting of vector of floats
* Add sorting of vector of structs by custom natural order
* Add sorting of vector of structs by custom comparator

fixes #437

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -4,6 +4,7 @@
 [About](about.md)
 - [Algorithms](algorithms.md)
   - [Generate Random Values](algorithms/randomness.md)
+  - [Sort a Vector](algorithms/sorting.md)
 - [Command Line](cli.md)
   - [Argument Parsing](cli/arguments.md)
 - [Compression](compression.md)
--- a/src/algorithms.md
+++ b/src/algorithms.md
@@ -8,6 +8,9 @@
 | [Generate random values of a custom type][ex-rand-custom] | [![rand-badge]][rand] | [![cat-science-badge]][cat-science] |
 | [Create random passwords from a set of alphanumeric characters][ex-rand-passwd] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
 | [Create random passwords from a set of user-defined characters][ex-rand-choose] | [![rand-badge]][rand] | [![cat-os-badge]][cat-os] |
+| [Sort a Vector of Integers][ex-sort-integers] | [![std-badge]][std] | [![cat-science-badge]][cat-science] |
+| [Sort a Vector of Floats][ex-sort-floats] | [![std-badge]][std] | [![cat-science-badge]][cat-science] |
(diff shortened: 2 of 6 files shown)


AndyGauge:
BTW: https://github.com/rust-lang-nursery/rust-cookbook/issues/443 The test failures are from a rust-lang bug

src/algorithms/sorting/sort_struct_natural.md
```diff
@@ -0,0 +1,70 @@
+## Sort a Vector of Structs by natural ordering
+
+[![std-badge]][std] [![cat-science-badge]][cat-science]
+
+Sorts a Vector of Person structs with a single property `name` by its natural
+order (By name). In order to make Person sortable you need four traits [`Eq`],
+[`PartialEq`], [`Ord`] and [`PartialOrd`]. Eq can be simply derived.
```
budziq: I feel that the proposed examples are quite verbose for relatively trivial functionality that is shown.

I'd at least suggest to combine the last two examples  src/algorithms/sorting/sort_struct_custom.md and  src/algorithms/sorting/sort_struct_natural.md

Also these last two examples are more focused in actually making the values sortable as opposed to actual sorting. This is not bad in itself but I fell that title does not does the example justice as the actual sorting seams to be the least important aspect of the example 

I'd also mention that `Eq`, `PartialEq`, `Ord` and `PartialOrd` can be just `#[derive]`'d. Moreover I'd only do the `impl`'s by hand if these do something interesting justifying the human work otherwise `#[derive]` is always a better / more idiomatic choice (which we should prefer), So to summarize I'd either use auto derives or make the struct more interesting to justify the manual impls.

src/algorithms/sorting/sort_struct_custom.md
```diff
@@ -0,0 +1,55 @@
...
+
+impl fmt::Debug for Person {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.name)
+    }
+}
+
+fn main() {
+    let expected = vec![
+        Person::new("Al".to_string(), 60),
```
budziq: I'd propose to move the whole body of `expected` to the right side of `assert_eq` as in the previous examples

Januson: I inlined the expected variables.

src/algorithms/sorting/sort_struct_natural.md
Januson: Hello, thanks for your suggestions. I merged the struct sorting examples and derived the structs.

AndyGauge:
I'm pretty happy with the way this turned out.

Issue #446, opened 2018-09-05
Recipe: Distance between points


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | std, num |
| Section  | 16.1 Mathematics  |

Given a geo-coordinate pair (lat/long) of 2 points, find the distance between them.

ludwigpacifici:
Hello @AndyGauge 

I'd be happy to write that recipe.

Munchotaur:
There is a minor error in this recipe. Paris is to the east of the prime meridian, so its coordinates should be assigned with a positive longitude, not negative:
 
`    let (paris_latitude_degrees, paris_longitude_degrees) = (48.85341_f64, 2.34880_f64);`

PR #452, opened 2018-09-29
Recipe: Distance between points

ludwigpacifici:
* Added recipe: Distance between two points on the Earth
* Update std badge version

fixes: #446

### DIFF
--- a/src/links.md
+++ b/src/links.md
@@ -114,7 +114,7 @@ Keep lines sorted.
 [serde-json-badge]: https://badge-cache.kominick.com/crates/v/serde_json.svg?label=serde_json
 [serde-json]: https://docs.rs/serde_json/*/serde_json/
 [serde]: https://docs.rs/serde/
-[std-badge]: https://badge-cache.kominick.com/badge/std-1.25.0-blue.svg
+[std-badge]: https://badge-cache.kominick.com/badge/std-1.29.1-blue.svg
 [std]: https://doc.rust-lang.org/std
 [syslog-badge]: https://badge-cache.kominick.com/crates/v/syslog.svg?label=syslog
 [syslog]: https://docs.rs/syslog/
--- a/src/science.md
+++ b/src/science.md
@@ -5,9 +5,11 @@
 | [Creating complex numbers][ex-creating-complex-numbers] | [![num-badge]][num] | [![cat-science-badge]][cat-science] |
 | [Adding complex numbers][ex-adding-complex-numbers] | [![num-badge]][num] | [![cat-science-badge]][cat-science] |
 | [Mathematical functions on complex numbers][ex-mathematical-functions] | [![num-badge]][num] | [![cat-science-badge]][cat-science] |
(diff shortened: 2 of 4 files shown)


src/SUMMARY.md
```diff
@@ -43,6 +43,8 @@
   - [Server](net/server.md)
 - [Operating System](os.md)
   - [External Command](os/external.md)
+- [Mathematics](mathematics.md)
```
AndyGauge: So I'd like this to the be **Science** category to match the Crates.io category "Science"  I'd also rather group all the **Mathematics** into a sub-section of the science category rather than breaking them into calculus/geometry/etc.

src/mathematics/calculus/latitude-longitude.md
```diff
@@ -0,0 +1,46 @@
+## Distance between two points on the Earth
+
+[![std-badge]][std]
+
+By default, Rust provides mathematical [float methods] such as
+trigonometric functions, square root, conversion functions between
+radians and degrees, and so forth.
+
+The following example will use [`to_radians`], [`sin`], [`cos`],
```
AndyGauge: There's a lot of passive voice in there.  Would you mind restructuring the sentances to be Subject noun verb rather than indirect object, to-verb noun?  Like THhis.  ['to_radians'] computes the distance in kilometers between two points ...  Your next sentance is fine, because the active voice of defining a struct is hard.  I can help you if active voice is a challenge.

src/SUMMARY.md
ludwigpacifici: To make sure I understand what you mean, change the titles from:

```
16. Mathematics
    16.1. Calculus
```

to:

```
16. Science
    16.1. Mathematics
```
?

src/mathematics/calculus/latitude-longitude.md
ludwigpacifici: Here is a version more with more active voice:

```
The following example computes the distance in kilometers between two
points on the Earth with the [Haversine formula]. Points are expressed
as pairs of latitude and longitude in degrees. Then, [`to_radians`]
converts them in radian. [`sin`], [`cos`], [`powi`] and [`sqrt`]
compute the central angle. Finally, it's possible to calculate the
distance.
```

If you think it can be better, help is welcomed :-) English is not my mother tong.

src/SUMMARY.md
AndyGauge: yes

src/mathematics/calculus/latitude-longitude.md
AndyGauge: Yeah, that will work.  I appreciate the effort

ludwigpacifici:
It's updated. Let me know if the hierarchy of flies is fine, i.e. the recipe is under Mathematics even though there is recently added section on complex numbers.

AndyGauge:
I'll let the dust settle, and do a little tidying at the end.  It's good!

Issue #455, opened 2018-10-01
Recipe: Inserting and Querying Data


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | rusqlite|
| Section  | 7.1 SQLite  |

Once the database has structure, insert and query data

robertDurst:
I'd like to take a shot at this if it hasn't been claimed yet!

AndyGauge:
I'm struggling with this, as this relies on #454 
You want them both?

robertDurst:
I can, but also want to leave it open for anyone else doing Hacktoberfest :)

AndyGauge:
That's why I made them separate, to help out with the hackathon!

robertDurst:
Cool, can I maybe claim #454 instead then?

lnicola:
Should we also have a recipe for transactions? Either here or separate.

AndyGauge:
#455 claimed by @robertDurst pending #454 structure.

robertDurst:
Cool @lnicola ! I'll keep my eyes peeled.

Thanks @AndyGauge !

robertDurst:
Is this good to begin? @lnicola 

A bit swamped in work today but will do have this done by Thursday Night/Friday.

lnicola:
@robertDurst You should probably base your code on #464. I'm not really planning to make any changes (unless I get some feedback otherwise), but it wasn't merged yet because of a Windows CI issue.

lnicola:
Yeah, you'll probably want to use `rust,no_run` in your code sections to avoid that issue.

robertDurst:
@AndyGauge and @lnicola I will have to "unclaim" this. Got very busy and will not quote have time :(

vishalsodani:
@AndyGauge Can I claim this ?

AndyGauge:
Yes.  Please perform Inserts and Querying in a recipe to show off Joining the two tables.  I'll create an Issue for Transactions.

Issue #454, opened 2018-10-01
Recipe: Creating Tables in Database


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | rusqlite|
| Section  | 7.1 SQLite  |

Build off of creating a database by creating a database structure

lnicola:
Claimed by myself as discussed in #453.

lnicola:
This was closed by mistake, can you please reopen it?

AndyGauge:
I'm available to help if you need anything

lnicola:
It was implemented in #464, right?

AndyGauge:
I didn't read the recipe, I just read the title, but it looks like it was.  I'll close this and update the Recipe Title.

Thanks!

PR #460, opened 2018-10-02
Upgrade to reqwest 0.9

### LINKED ISSUE #453: Recipe: Creating Database in File System
|  |  |
| Crates  | rusqlite|
| Section  | 7.1 SQLite  |

This recipe should demonstrate using rusqlite by saving the database to the file system.

### LINKED ISSUE #459: Bump reqwest to 0.9
`reqwest 0.9` was recently released, using `hyper 0.12`. A couple of examples need to be updated because of the header changes. It should also help with `openssl` compatibility -- I can't build the examples because of a version mismatch.

I'll try to work on this.

lnicola:
fixes #453, #454

Based on #460. I'll rebase afterwards, please ignore the first two commits.

CC @robertDurst as per #455.

--------------------------
### Things to check before submitting a PR

- [x] the tests are passing locally with `cargo test`
- [x] cookbook renders correctly in `mdbook serve -o`
- [x] PR contains correct "fixes #ISSUE_ID" clause to autoclose the issue on PR merge
- [x] spell check runs without errors `./ci/spellchecker.sh`
- [x] link check runs without errors `link-checker ./book`
- [x] non rendered items are in sorted order (links, reference, identifiers, Cargo.toml)
- [x] links to docs.rs have wildcard version `https://docs.rs/tar/*/tar/struct.Entry.html`
- [x] example has standard [error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)
- [x] code identifiers in description are in hyperlinked backticks

### Things to do after submitting PR
- [ ] check if CI is happy with your PR


src/web/clients/download/partial.md
```diff
@@ -34,32 +36,29 @@ use reqwest::StatusCode;
...
+#     type Item = HeaderValue;
 #
 #     fn next(&mut self) -> Option<Self::Item> {
 #         if self.start > self.end {
 #             None
 #         } else {
 #             let prev_start = self.start;
 #             self.start += std::cmp::min(self.buffer_size as u64, self.end - self.start + 1);
-#             Some(Range::bytes(prev_start, self.start - 1))
+#             Some(HeaderValue::from_str(&format!("bytes={}-{}", prev_start, self.start - 1)).unwrap())
```
AndyGauge: Can we avoid `unwrap()`?  We are really trying to teach how to implement these functions in production code, to handle runtime errors gracefully.

src/web/clients/download/partial.md
```diff
@@ -2,28 +2,30 @@
...
 [![reqwest-badge]][reqwest] [![cat-net-badge]][cat-net]
 
-Uses [`reqwest::Client::head`] to get the content-length and validates the
-server setting the header [`reqwest::header::ContentRange`].
+Uses [`reqwest::Client::head`] to get the [Content-Length] of the response.
 
-If supported, downloads the content using [`reqwest::get`] by setting the
-[`reqwest::header::Range`] to do partial downloads printing basic progress
-messages in chunks of 10240 bytes.
+If supported, downloads the content using [`reqwest::Client::get`] and
```
AndyGauge: It took a few re-reads for me to get what was 'setting the Range header'.  Can we go with: 
If supported, downloads the content using reqwest::Client::get with the Range header set to do partial ...

src/web/clients/download/partial.md
lnicola: I don't think so. The error case is for things like non-ASCII characters, I suppose. That can be proven statically (`prev_start` and `self.start - 1` are integers, and the format string is fine), but we can't avoid the `unwrap()` until it gets hidden away in the `headers` crate.

src/web/mime/request.md
```diff
@@ -3,38 +3,44 @@
 [![reqwest-badge]][reqwest] [![mime-badge]][mime] [![cat-net-badge]][cat-net] [![cat-encoding-badge]][cat-encoding]
 
 When receiving a HTTP reponse from *reqwest* the [MIME type] or media type may be
```
AndyGauge: I made a mistake when I merged the crate names Italicized.  Can you replace *reqwest* with `reqwest` and *mime* with `mime` in this PR?

src/web/clients/download/partial.md
lnicola: To be fair, the old phrasing wasn't much better :-).

Does this sound all right?

> The code then uses [`reqwest::Client::get`] to download the content in
chunks of 10240 bytes, while printing progress messages. The requested
ranges are specified using with the [Range] header.

lnicola:
Added a new commit with some phrasing changes. I'll squash after you review it.

src/web/clients/download/partial.md
lnicola: Added a comment to explain why the `unwrap` never fails.

lnicola:
Rebased and squashed.

PR #463, opened 2018-10-02
Add ndarray badge and link

JamesHinshelwood:
Seeing as we are adding a few `ndarray` recipes in #450 it makes sense to add the badge and link.

Issue #450, opened 2018-09-05
Recipe: Matricies Series


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | ndarray |
| Section  | 16.2 Matricies  |

Investigate and create the following recipes:

- [x] Add two vectors
- [x] Norm of two vectors
- [ ] A vector time a matrix
- [ ] Add, multiply two matrix
- [x] invert a Matrix
- [x] A scalar time a vector time a matrix
- [ ] Serialize/Deserialize a big matrix

j-haj:
Section 16 is Text Processing - is the intent for `ndarray` work to land in the text processing section?

AndyGauge:
Section 16 is Science.  It doesn't exist yet.  I think #452 Creates the Science section.

mfajer:
I'll work on this one.

AndyGauge:
@mfajer do you want all of them or just a few?  Feel free to do individual PRs per recipe.

j-haj:
@mfajer FYI I have started working on the vector sum item

j-haj:
Per #452, I think this work belongs under `linear-algebra`, so the directory layout will be `src/science/mathematics/linear-algebra/`.

mfajer:
@AndyGauge I will do individual PRs per recipe so that @j-haj and others can contribute.

JamesHinshelwood:
@j-haj @mfajer I would like to help out with some of these recipes, if they aren't already being worked on. Which are in progress? Would I be safe to start with some of the matrix stuff?

j-haj:
@JamesHinshelwood yea! I have a PR for the vector sum example and will have another PR for the vector norm example shortly, so those two are good. I'm not sure what @mfajer is working on but as far as I am aware everything else is up for grabs.

JamesHinshelwood:
`ndarray` does not provide a method to calculate the inverse of a matrix. This is provided by the [`ndarray_linalg`](https://docs.rs/ndarray-linalg/0.10.0/ndarray_linalg/index.html) crate - should we use this?

Or perhaps we could use `nalgebra`? (which I have a personal preference for :wink:)

j-haj:
[...]

PR #466, opened 2018-10-02
Adds vector norm example.

j-haj:
Part of tracking issue #450.

Once #452 is merged I will rebase and add appropriate section links.
Similarly, once #463 is merged I will update badge links.


j-haj:
Rebased. Adds function that normalizes vector per suggestion by @mfajer

Issue #433, opened 2018-07-12
Replace quick_main with main returning Result


AndyGauge:
Rust 1.26 allows return types on main.  This should be taught in the cookbook, rather than relying on quick_main.  

TODO: 

- [ ] Replace functions in recipes that leverage a function returning a Result and quick-main macros
- [ ] Document in a note about error handling

https://www.reddit.com/r/rust/comments/8xubdr/rust_cookbook_is_now_reorganized/e28efj7/

sepiropht:
I will Be happy to take this one too

budziq:
+1

budziq:
Yep, optimaly all occurences should be unneccessary now including the non example text sections. Just grep the source for `quick_main`. If you'll have any questions just drop a comment here.

AndyGauge:
cc #525

PR #468, opened 2018-10-03
Removes quick run in favor of main for compression recipes.

j-haj:
Cleanup PR. See issue #433

### DIFF
--- a/src/compression/tar/tar-strip-prefix.md
+++ b/src/compression/tar/tar-strip-prefix.md
@@ -11,20 +11,20 @@ via [`Entry::unpack`].
 # extern crate error_chain;
 extern crate flate2;
 extern crate tar;
-#
-# error_chain! {
-#     foreign_links {
-#         Io(std::io::Error);
-#         StripPrefixError(::std::path::StripPrefixError);
-#     }
-# }
 
 use std::fs::File;
 use std::path::PathBuf;
 use flate2::read::GzDecoder;
 use tar::Archive;
+# 
+# error_chain! {
+#   foreign_links {
+#     Io(std::io::Error);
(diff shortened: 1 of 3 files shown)


src/compression/tar/tar-strip-prefix.md
```diff
@@ -7,24 +7,33 @@ the specified path prefix (`bundle/logs`).  Finally, extract the [`tar::Entry`]
...
-#     }
-# }
 
 use std::fs::File;
 use std::path::PathBuf;
 use flate2::read::GzDecoder;
 use tar::Archive;
 
-fn run() -> Result<()> {
+#[derive(Debug)]
```
AndyGauge: I would hide these new lines by default.  If you put a hash tag in front of them and a space, it would make understanding the core of the recipe easier.

j-haj:
Sure thing. Do you want the entire `StripError` stuff hidden or just the impls?

AndyGauge:
Let's go with all of it.  It is really easy to find it, and it makes reasoning the recipe easier as an entry point.

JamesHinshelwood:
In my code I've usually just used `Box<std::error::Error>`, for functions with multiple possible error types. I'm wondering which method is considered more 'idiomatic'?

j-haj:
@JamesHinshelwood I saw this approach mentioned but wasn't sure which way was more idiomatic as well. This seemed most explicit but also a bit clunky. Could be a good idea for a recipe..

AndyGauge:
I've been doing a poor job of reviewing.  We use error chain when there are multiple error variants.

j-haj:
I will update the PR this morning

PR #470, opened 2018-10-03
Add Statistics Section With Examples

### LINKED ISSUE #447: Recipe: Statistics
|  |  |
| Crates  | std |
| Section  | 16.1 Mathematics  |

Given a collection of integers, calculate the mean, median, mode, standard deviation, maybe Z-score of the highest outlier.

Lacaranian:
Add a statistics section to the cookbook!

Add an example showing how to calculate mean, median, and mode using the
Rust standard library.

Add a second example showing how to compute the standard deviation and zscore
of a set of data.

fixes #447


src/intro.md
```diff
@@ -44,8 +44,10 @@ community. It needs and welcomes help. For details see
 
 {{#include os.md}}
 
+{{#include science.md}}
```
AndyGauge: Good Catch

src/mathematics/statistics.md
```diff
@@ -0,0 +1 @@
+# Statistics
```
AndyGauge: We don't want this file

src/science/mathematics/statistics/central-tendency.md
```diff
@@ -0,0 +1,88 @@
+### Measures of central tendency
+
+[![std-badge]][std]
+
+Measures of central tendency summarize a set of data as an "average" or "expected" value that is likely to be representative of a typical member of the data set in some way. The most common measures of central tendency are the
```
AndyGauge: The description should be shorter than this.  We don't need to know the implementation details, instead it is better to know from a high level what is going on, and linking API documentation.  Shoot for no more than 5 sentences / 2 paragraphs.  For such a simple task (calculating the mean) there is excessive description.

I would move the description of Option from Standard Deviation to mean's description.

src/science/mathematics/statistics/central-tendency.md
```diff
@@ -0,0 +1,88 @@
...
+    let mean = match count {
+       positive if positive > 0 => Some(sum  / count as f32),
+       _ => None
+    };
+
+    println!("Mean of the data is {:?}", mean);
+}
+```
+
+Calculating the median [`sort`]s the data in-place, requiring the input slice to be mutable. If the number of elements is even, calculate the mean of the middle two measurements, which can be passed to the previous mean function as a sub[`slice`] of the original slice. For odd-sized data sets, simply pick the measurement from the middle of the sorted set.
```
AndyGauge: Shouldn't the median not affect the data set?  Is there a way to calculate the median without mutating the data?

Lacaranian:
Can do, I'll get these updated shortly

src/science/mathematics/statistics/central-tendency.md
Lacaranian: There are a few ways, though a straight `clone` and `sort` isn't very performant. I might be able to conjure a good implementation of the quickselect algorithm, though that may be slightly complex for one of these examples (while it could show off some of Rust's finer points). Bears some thinking about...

src/science/mathematics/statistics/central-tendency.md
Lacaranian: :+1:

src/mathematics/statistics.md
Lacaranian: :+1:

Lacaranian:
This went off the rails somehow and lost its reference to my fork, so I'll close this out in favor of the working PR.

PR #461, opened 2018-10-02
Adds vector sum example.

j-haj:
This is part of tracking issue #450. Also, it does not appear there is an `ndarray` badge.

Note: There are a few additional items to add to this commit. I can rebase off of #452 or wait until that PR is merged and rebase off master. Either way, once that is done I can add in the proper links.

PR #471, opened 2018-10-03
Add various simple matrix recipes

JamesHinshelwood:
Part of #450 

I'm unsure as to how we want to organise these. #461 seems to use another `linear-algebra` crate, but I'm not sure this is necessary. Could we just put all the recipes in `science/mathematics/matrices.md`?


JamesHinshelwood:
This will need to be rebased after/if #473 lands.

AndyGauge:
Will you pull master and update the links?

JamesHinshelwood:
@AndyGauge Done

PR #473, opened 2018-10-04
Reorganise science and fix broken links

JamesHinshelwood:
fixes #472 

We should formalise how to lay things out to avoid this in the future and conventions. e.g. `_`s and `-`s, where to put tables, etc.

This format does involve a fair bit of (error prone) duplication from writers - maybe in the future this could be automaed?

Here is a rough diagram I drew up
```
section.md - Contains {{#include section/category.md}} for large categories and tables for small_subcategories
section
├──category.md - Contains tables for each subcategory
├──category - Contains multiple subcategories
│   ├──subcategory.md - Contains {{#include subcategory/recipe.md}}s
│   └──subcategory
│       └──recipe.md - Contains actual recipe
│
├──small_subcategory.md - Contains {{#include small_subcategory/recipe.md}}
└──small_subcategory - Should be preferred when an extra level of nesting is unnecessary
    └──recipe.md - Contains actual recipe
```

PR #478, opened 2018-10-06
Fix broken links in Science section

pduzinki:
Links for science section were broken (error 404)

### DIFF
--- a/src/science.md
+++ b/src/science.md
@@ -11,13 +11,13 @@
 | [Vector Sum][ex-ndarray-vec-sum] | [![ndarray-badge]][ndarray] | [![cat-science-badge]][cat-science] |
 | [Vector norm calculation][ex-ndarray-vec-norm] | [![ndarray-badge]][ndarray] | [![cat-science-badge]][cat-science] |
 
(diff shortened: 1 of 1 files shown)


src/science/mathematics/big-integers.md
```diff
@@ -0,0 +1,30 @@
+# Miscellaneous
+
+## Big integers
+
+[![num-badge]][num] [![cat-science-badge]][cat-science]
+
+If we need to make calculations on integers exceeding size of standard integer type, we can use BigInt.
```
JamesHinshelwood: I think we should link to [BigInt](https://docs.rs/num-bigint/0.2.0/num_bigint/struct.BigInt.html) here.

src/science/mathematics/big-integers.md
```diff
@@ -0,0 +1,30 @@
...
+    for i in 1..(x+1) {
+        factorial = factorial * i;
+    }
+
+    factorial
+}
+
+fn main() {
+    let x = 100;
+    println!("{}! equals {}", x, factorial(100));
```
JamesHinshelwood: Small point, but maybe `println!("{}! equals {}", x, factorial(x)` would look nicer?

AndyGauge:
fixed in #473

PR #482, opened 2018-10-09
Bump mdbook version

JamesHinshelwood:
Is there any good reason we aren't using the latest version of mdbook? All that was required was to move the `theme` directory into `src`.


CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ cd rust-cookbook
 Cookbook is built with [mdBook], so install that first with Cargo:
 
 ```
-cargo install --version 0.1.8 mdbook
+cargo install mdbook
```
AndyGauge: lets be specific.  If you want to go to 0.2, specify it in CONTRIBUTING

README.md
```diff
@@ -21,7 +21,7 @@ If you'd like to read it locally:
 ```bash
 $ git clone https://github.com/rust-lang-nursery/rust-cookbook
 $ cd rust-cookbook
-$ cargo install mdbook --vers "0.1.8"
```
AndyGauge: And in README

JamesHinshelwood:
@AndyGauge Fixed

AndyGauge:
This one makes me nervous.

JamesHinshelwood:
@AndyGauge You were right to be nervous, this has broken how we handle links.

[0.2.0](https://github.com/rust-lang-nursery/mdBook/releases/tag/v0.2.0)
> This release changes how links are handled in mdBook. Previously, relative links were interpreted relative to the book's root. In 0.2.0+ links are relative to the page they are in. This has several advantages, such as making links work in other markdown viewers like GitHub. You will likely have to change links in your book to accommodate this change.

This makes sense, but we include the table from a subpage in the main page (e.g. `science.md` includes the table for `mathematics.md`), so I see no way of getting this to work with the way we handle things. Currently, either the `science` or `mathematics` tables will link to the wrong thing.

JamesHinshelwood:
Apologies for not noticing this when upgrading. In future, we can pass the build through something like [mdbook-linkchecker](https://github.com/Michael-F-Bryan/mdbook-linkcheck), which I have verified will catch this error. It also lets us check external links, meaning we might be able to use it instead of `link-checker` as well.

AndyGauge:
Its fine, we just need to update links to be absolute.

JamesHinshelwood:
Production doesn't work for me right now. Clicking `Vector Sum` from the `science` table leads to the [right page](https://rust-lang-nursery.github.io/rust-cookbook/science/mathematics/linear_algebra.html#vector-sum), whereas clicking `Vector Sum` from the `mathematics` table leads to a [missing page](https://rust-lang-nursery.github.io/rust-cookbook/science/science/mathematics/linear_algebra.html#vector-sum) (Note the `science/science` url).

PR #486, opened 2018-10-10
Improve vector norm recipe

jturner314:
This removes an unnecessary allocation caused by `.mapv()` in the original version, replaces `&Array1<f64>` argument types with `ArrayView1<f64>` (see the new text for reasoning), and takes advantage of the `array!` macro to make the array creation more concise.

### DIFF
--- a/src/science/mathematics/linear_algebra/vector-norm.md
+++ b/src/science/mathematics/linear_algebra/vector-norm.md
@@ -1,48 +1,58 @@
 ## Vector Norm
 [![ndarray-badge]][ndarray]
 
-This recipe demonstrates use of the [`Array1`] type, [`dot`], [`map`], and
-[`scalar_sum`] in computing the [l1] and [l2] norms of a given vector. The l2
-norm calculation is the simpler of the two, as it is the square root of the dot
-product of a vector with itself, shown in the function `l2_norm`. The l1 norm,
-shown in the function `l1_norm` is computed by mapping the elements of the input
-vector to their absolute value, and then calling `scalar_sum`, which reduces the
-elements via summation.
+This recipe demonstrates use of the [`Array1`] type, [`ArrayView1`] type,
+[`fold`] method, and [`dot`] method in computing the [l1] and [l2] norms of a
+given vector. The l2 norm calculation is the simpler of the two, as it is the
+square root of the dot product of a vector with itself, shown in the function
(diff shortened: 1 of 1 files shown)


src/science/mathematics/linear_algebra/vector-norm.md
```diff
@@ -1,48 +1,58 @@
...
+square root of the dot product of a vector with itself, shown in the function
+`l2_norm`. The l1 norm, shown in the function `l1_norm`, is computed by a `fold`
+operation that sums the absolute values of the elements. (This could also be
+performed with `x.mapv(f64::abs).scalar_sum()`, but that would allocate a new
+array for the result of the `mapv`.)
+
+Note that both `l1_norm` and `l2_norm` take the [`ArrayView1`] type. This recipe
 considers vector norms, so the norm functions only need to accept one
-dimensional arrays (hence [`Array1`]).
+dimensional views (hence [`ArrayView1`]). While the functions could take a
```
AndyGauge: This level of detail is helpful.  Typically we shoot for smaller descriptions, but I'm going to keep it as is.

src/science/mathematics/linear_algebra/vector-norm.md
```diff
@@ -1,48 +1,58 @@
...
+or `&view` instead of `x.view()`. If the function is part of your public API,
+that may be a better choice for the benefit of your users, but for internal
+functions, the more concise `ArrayView1<f64>` may be preferable.
 
 ```rust
+#[macro_use(array)]
 extern crate ndarray;
 
-use ndarray::{Array, Array1};
+use ndarray::prelude::*;
```
AndyGauge: Although prelude is often the library's often preferred method of including types, the cookbook prefers to be explicit about it.

AndyGauge:
Oh, and thanks for this effort!

src/science/mathematics/linear_algebra/vector-norm.md
jturner314: Okay, I've made it explicit again.

src/science/mathematics/linear_algebra/vector-norm.md
jturner314: Yeah, there might be a better place for this explanation, or maybe we should add it to the `ndarray` docs instead and leave it out of the cookbook. I thought it would be useful for newcomers, though, since one of the first questions when writing functions with `ndarray` types is what parameter type is best. I'll follow your guidance on whether to keep/remove this.

PR #496, opened 2018-11-09
Update crossbeam-spawn.md

stjepang:
In the latest version od Crossbeam ([docs](https://docs.rs/crossbeam/0.5.0/crossbeam/)) we have two changes:

1. Spawned closures take an argument of type `&Scope` which can be used for nested spawns.
2. The `scope` function returns a `Result` indicating whether spawned threads have been joined with success or not.

Updating the example accordingly.


src/concurrency/thread/crossbeam-spawn.md
```diff
@@ -28,13 +28,13 @@ fn find_max(arr: &[i32], start: usize, end: usize) -> i32 {
...
-        let left = scope.spawn(|| find_max(arr, start, mid));
-        let right = scope.spawn(|| find_max(arr, mid, end));
+        let left = scope.spawn(|_| find_max(arr, start, mid));
+        let right = scope.spawn(|_| find_max(arr, mid, end));
 
         // NOTE(unwrap): `join` will return an error if the thread panicked.
         // This way, panics will be propagated up to the `scope` call
         cmp::max(left.join().unwrap(), right.join().unwrap())
-    })
+    }).unwrap()
```
AndyGauge: Can you catch the thread panics rather than unwrapping them?

stjepang: What are you suggesting exactly?

We could call `.expect("a child thread panicked")` to display a more informative message. Or maybe print the panic errors and continue with the program. But it's better to just `.unwrap()` since we do the same thing one line above - we just propagate errors up.

In the end, this `.unwrap()` has no significance because we don't join any threads at the end of the scope - note that all spawned threads are already joined inside the scope.

AndyGauge: In the cookbook we try to reinforce good habits.  Even if the use case isn't as practical as another, we like to treach how to propogate errors up, like you say.  Here, the return type of the function is `i32` and if you instead return `Result<i32>`.

stjepang: I've updated the code so that `find_max()` returns an `Option<T>` (returns `None` when the array is empty). Does it look better now?

PR #509, opened 2018-12-27
Implement the FromStr trait for a custom struct

### LINKED ISSUE #400: Recipe: custom type implements FromStr
|  |  |
| Crates  | std  |
| Section  | 16.2 String Parsing  |

Create a custom type that implements FromStr.  An example would be IPv4Address.  This could be an email address, splitting on the '@', or separating first,middle (optional), and last names (this is really hard internationally, but a contrived example might be helpful).

KappaDistributive:
fixes #400

### DIFF
--- a/ci/dictionary.txt
+++ b/ci/dictionary.txt
@@ -103,6 +103,7 @@ fname
 Formatter
 foundational
 fout
+fromstr
 FromStr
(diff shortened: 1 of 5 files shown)


AndyGauge:
Thanks!

PR #511, opened 2019-01-10
Updates for rand 0.6.4

### LINKED ISSUE #497: Update the random number generation page to use rand 0.6
The `rand` crate just released its 0.6 release, so the cookbook should be updated accordingly, especially since the API got a big more ergonomic to use.

Here's an update guide: https://rust-random.github.io/book/update-0.6.html

Here's an overview of how code with `rand` looks like: https://rust-random.github.io/book/overview.html

The biggest benefit is the `random()` function being able to generate values of any type with just importing the prelude, which is really useful. One note I'd like to say is that generating random booleans should be in the cookbook because that functionality is extremely common and most people use `if rng.gen::<f64>() < 0.5` which works but doesn't look as good as the recent `if random()`

brettlyons:
fixes #497


AndyGauge:
I think you need to update the Cargo.toml file to match the version for the tests to pass.

PR #512, opened 2019-01-12
Fix SQLite example so it compiles

k2l8m11n2:
Previous version gave me
```
error[E0282]: type annotations needed
  --> src/main.rs:13:9
   |
13 |         &[],
   |         ^^^ cannot infer type

```
on stable and nightly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ rayon = "1.0"
 regex = "1.0"
 reqwest = "0.9"
 ring = "0.13.0-alpha"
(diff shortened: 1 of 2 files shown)


k2l8m11n2:
Seems that `NO_PARAMS` was only added in rusqlite 0.16.0, I'll bump the version.

gwenn:
When executing [DDL](https://en.wikipedia.org/wiki/Data_definition_language), [`execute_batch`](https://docs.rs/rusqlite/*/rusqlite/struct.Connection.html#method.execute_batch) should be preferred.

PR #517, opened 2019-03-17
add unicode_segmentation example

galvez:
Hey @branderson, would you mind taking a look at this? `cargo test` seems fine but I can't get the `mdbook` test to run locally (only the one test I'm adding).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ toml = "0.4"
 url = "1.6"
 walkdir = "2.0"
 ansi_term = "0.11.0"
+unicode-segmentation = "1.2.1"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 syslog = "4.0"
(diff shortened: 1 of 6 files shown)


AndyGauge:
This is what I see:
error[E0277]: can't compare `&str` with `str`
  --> C:\Users\appveyor\AppData\Local\Temp\1\rust-skeptic.NZor5XZnVwqW\test.rs:10:5
   |
10 |     assert_eq!(&graphemes[3], "é");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `&str == str`
   |
   = help: the trait `std::cmp::PartialEq<str>` is not implemented for `&str`
   = note: required because of the requirements on the impl of `std::cmp::PartialEq<&str>` for `&&str`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

I would try  `assert_eq!(graphemes[3], "é");`

galvez:
@AndyGauge thank you! fixed it :)

galvez:
@AndyGauge seems all is good now -- moved it to text/string_parsing too.

Issue #456, opened 2018-10-01
Recipe Series: Postgres


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | postgres |
| Section  | 7.2 Postgres Database  |

Looking for 3 Recipes that demonstrate using Postgres to:

 - [x] Create Tables and database structure
 - [x] Perform Simple insert and query operation
 - [x] Perform aggregate operations

The goal of the cookbook is to provide enough insight about the crates that anyone could get their feet wet, and learn the best practices.

AndyGauge:
There should be a new section 'Database' that comes after data structure.  SQLlite comes first then Postgres.

AndyGauge:
Do you want the aggregate recipe, or should I un-claim the issue?

vishalsodani:
I would like to do the aggregate recipe. What would you like to see in this recipe ? Count(*) , max ?

AndyGauge:
How about sort authors by number of books?

vishalsodani:
> How about sort authors by number of books?

ok. The insert recipe did not show inserting data into the Book table. So, should I start by showing inserting data into the Book table and the do the sort query?

KappaDistributive:
Are there any other recipes that we would like to add to the Postgres section?

AndyGauge:
I'm closing this issue, if new postgres ideas come up we can open new issues.  Thanks!

PR #520, opened 2019-04-06
added data aggregation example for postgres

KappaDistributive:
addresses issue #456


AndyGauge:
Thanks, I like the subject matter and simplicity of the example.

PR #442, opened 2018-08-16
add new gzip section in compression chapter

sepiropht:
--------------------------
### Things to check before submitting a PR

- [x] the tests are passing locally with `cargo test`
- [x] cookbook renders correctly in `mdbook serve -o`
- [x] commits are squashed into one and rebased to latest master
- [ ] PR contains correct "fixes #ISSUE_ID" clause to autoclose the issue on PR merge
    -  if issue does not exist consider creating it or remove the clause
- [ ] spell check runs without errors `./ci/spellchecker.sh`
- [ ] link check runs without errors `link-checker ./book`
- [ ] non rendered items are in sorted order (links, reference, identifiers, Cargo.toml)
- [ ] links to docs.rs have wildcard version `https://docs.rs/tar/*/tar/struct.Entry.html`
- [ ] example has standard [error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)
- [ ] code identifiers in description are in hyperlinked backticks
```markdown
[`Entry::unpack`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack
```

### Things to do after submitting PR
- [ ] check if CI is happy with your PR

Thank you for reading, you may now delete this text! Thank you! :smile:

PR #524, opened 2019-04-12
439 download gzip content

### LINKED ISSUE #439: Recipe: Download GZip content
|  |  |
| Crates  | reqwest, flate2  |
| Section  | 3.2 Download GZip content  |

Download and decompress gzipped content from an Http site.

KappaDistributive:
fixes #439 

Note: This doesn't use the [recommended way of error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling) because `std::error::Error` doesn't implement the trait `std::convert::From<reqwest::Error>` and adding that implementation, in my opinion, would have cluttered the recipe too much.

### DIFF
--- a/ci/dictionary.txt
+++ b/ci/dictionary.txt
@@ -116,6 +116,7 @@ graphemes
 Graphemes
 GzDecoder
 GzEncoder
+gzip
 Hackerman
 Hardcoded
 hardcoded
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -10,6 +10,7 @@
   - [ANSI Terminal](cli/ansi_terminal.md)
 - [Compression](compression.md)
   - [Working with Tarballs](compression/tar.md)
+  - [Working with gzip files](compression/gzip.md)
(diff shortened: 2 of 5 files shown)


AndyGauge:
I apologize for not making this clearer, but what we are attempting to showcase is what #442 attempted.  Resources on the internet may be encoded by gzip.   In that PR you can see https://httpbin.org/gzip is used as a source, which happens to be json encoded with gzip, compressing the stream served over http.  The example is very basic, as reqwest automatically decodes gzip content (a fact I did not know when I recommended the recipe).

KappaDistributive:
Okay, no problem. I'll adapt my PR (either later today or tomorrow).

KappaDistributive:
Now that I've looked at it. The way reqwest automatically handles the gzip compression really doesn't leave anything for the user to do. 

I'd recommend we close this issue.

AndyGauge:
Thanks for the effort, I think I agree with you that there's nothing to show.

PR #540, opened 2019-08-21
Fix typo

Stranger6667:
Hej :) It is a small typo that breaks syntax highlighting


AndyGauge:
Thanks, but this is a duplicate of #535

PR #547, opened 2019-09-18
Update inclusive range syntax, fix script pathname

### LINKED ISSUE #541: Replace ... with ..=
`warning: `...` range patterns are deprecated`

workingjubilee:
This PR fixes pathnames referencing `spellchecker.sh` as opposed to `spellcheck.sh`, and adds clarifying remarks on its behavior, since for half a second I was honestly mildly uncertain on whether it had executed correctly.
It also fixes #541 by updating instances of inclusive range syntax to use the `..=` style.
It also updates the url crate dependency and explicitly depends on the percent-encoding crate that the percent encoding example was depending on (url no longer re-exports this, and it was causing errors on one of the builds anyways).
This commit compiles and tests correctly on rustc v1.37.0; link-checker found 452 failures before and after this commit, so no new errors were added by this commit.

### Resolved Before PR Opened:
- [X] the tests are passing locally with `cargo test`
- [X] cookbook renders correctly in `mdbook serve -o`
- [X] PR contains correct "fixes #ISSUE_ID" clause to autoclose the issue on PR merge
- [X] spell check runs without errors `./ci/spellcheck.sh` <--- this was incorrect, this PR fixes that
- [X] link check runs without errors `link-checker ./book` ... **ish**
- [X] Remainder were N/A

### Resolving After PR Opens:
- [x] check if CI is happy with your PR


workingjubilee:
...After a few go-arounds, I managed to fix the percent encoding example to where the final result tests fine on Xenial + Nightly, which was an error before, but now the Mac + Nightly build errors because it can't find AsciiSet or and CONTROLS in the dependency but everyone else can (I assume it's pulling in the wrong version of the dependency, since this installs both). I am going to write that off as "good enough".

AndyGauge:
The fact this didn't fail in nightly linux has me more perplexed.  I appreciate the effort, I think today I'll try to make the build green.

workingjubilee:
Oh hey... now it builds...???

Issue #440, opened 2018-07-29
Recipe Series: Sharing data between threads


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | std,  crossbeam_channel |
| Section  | 4.3 Sharing data between threads  |

    - shared ownership with `Arc`
    - shared stack with `crossbeam`
    - mutable data with `Mutex`
    - passing messages with `mspc` (crossbeam_channel)

MagnumOpus21:
Can I take a shot at this? Any pointers are appreciated.

AndyGauge:
You don't have to tackle the whole thing, if you want to attempt some of it, we can split it up.

AndyGauge:
Also, I am not an expert in threading.  In fact, I don't know how to recognize idiomatic threaded code.  I'm sure we can pull in some other resources for this, but it might take some time.

MagnumOpus21:
It makes sense to split this up.

sn99:
I could work this one but it might take a 4-5 days

j-haj:
Just as a heads up for anyone interested in this issue, I am working on a message passing example with `crossbeam`

KodrAus:
@j-haj Sounds good!

@MagnumOpus21 A good place to get started is [this section of the Rust book](https://doc.rust-lang.org/stable/book/second-edition/ch16-01-threads.html). It has a lot of examples of using the [`std::thread`](https://doc.rust-lang.org/std/thread/index.html) and [`std::sync`](https://doc.rust-lang.org/std/sync/index.html) modules for concurrency. If you'd like to take a shot at one of the examples, like sharing data between threads with `Arc`, then we could provide some more support on the PR itself. It doesn't have to be complete when you open it!

`crossbeam` is a very subtle set of tools, so we'll probably want to keep any examples using `crossbeam::epoch` directly very simple, or just avoid them altogether in the cookbook, and rely on the higher-level pieces that have already been built over the top.

j-haj:
I have two examples for this: a simple single produce, single consumer example and a more complex example using the so-called ventilator pattern of a single produce to N consumers followed by N producers to a single consumer. I will have the PRs ready in a day or so.

papac25:
how can u determine if you have been hacked

papac25:
#- - - cschneid

kkibria:
Hi all:
[...]

PR #551, opened 2019-10-01
Adds simple single producer, single consumer crossbeam channel example.

j-haj:
Part of tracking issue #440. This adds a simple message passing example with two threads.


j-haj:
@AndyGauge do we know why the `percent_encoding` example is failing? I noticed this locally but it seemed somewhat flaky.

AndyGauge:
It's nightly only problem and it has to do with the way skeptic reads rlibs which is way above my capabilities.  That's diving into the internals, consuming a non-public API.  I was looking at rolling back the percent-encoding example but that attempt failed too.  If you can figure out a way to get tests to pass I bet I'd approve the PR

PR #553, opened 2019-10-02
Add matrix-eigenvalues example to /science/mathematics/linear_algebra

mxxo:
More basic linear algebra for the matrix series: #450


mxxo:
Thinking on this, it's better if this example doesn't require `nalgebra-lapack` and its associated dependencies. 

There is a pure-Rust general eigendecomposition [module](https://github.com/rustsim/nalgebra/blob/dev/src/linalg/eigen.rs) in `nalgebra`, but it's undocumented and isn't ready for prime-time: 
https://github.com/rustsim/nalgebra/blob/48f8333e00f31f5211716f18fce625187cdb2fd8/src/linalg/mod.rs#L22

I still think this pull request is worth including as a basic linear algebra example, but I'm going to close this for now until the relevant `nalgebra` module is ready.

PR #556, opened 2019-10-14
Adds cargo clean to CI test script.

j-haj:
This adds `cargo clean` to the test script run by TravisCI, which was motivated by [this](https://github.com/budziq/rust-skeptic/issues/111#issuecomment-539662466) comment.

I guess it doesn't hurt adding this, but it doesn't seem to solve the issue for me locally.

### DIFF
--- a/ci/test_script.sh
+++ b/ci/test_script.sh
@@ -46,6 +46,7 @@ if [[ "${CONTENT_TESTS:-}" == 1 ]]; then
     fi
 else
     echo "Testing code snippets"
+    cargo clean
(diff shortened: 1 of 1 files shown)


AndyGauge:
The nice thing about travis and docker is you get a clean environment to run the tests every test run.  It will fetch the latest rust version onto a completely new container, just like installing rust for the first time onto a new machine.  Because of this `cargo clean` doesn't resolve these types of issues in travis.  I really appreciate the effort to try to resolve this bug that's hurting our cookbook!

PR #561, opened 2019-12-14
Exclude percent encoding from skeptic

### LINKED ISSUE #557: percent-encode.md has some typos
- The url badge links to Wikipedia instead of docs.rs
- The badge is confusing because it refers to "url" while percent-encoding now seems to be its [own crate] (though it still comes from https://github.com/servo/rust-url

[own crate]: https://docs.rs/percent-encoding/

AndyGauge:
fixes #557

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target/
 book/
 *.swp
 lines.txt
+.idea/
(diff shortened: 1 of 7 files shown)


AndyGauge:
fixes tests but kills syntax highlighting :face_with_head_bandage:

Issue #502, opened 2018-12-04
on error handling


BurntSushi:
I realize opinions might run a bit strong here, but I'd like to make a case against the use of `error-chain` (or `failure`) for error handling in the cookbook examples. I think the error handling mechanisms provided by the standard library should be used instead.

Some background: I've been using Rust for a long time (since before 1.0), so the cookbook isn't actually something I've ever read in close detail before. I've also never used `error-chain` before (which I think probably reveals some bias in my argument here, FWIW).

I have an experience report from a colleague of mine that is in the process of learning Rust. They were using this cookbook as a way of reading example code. One of the things they did in their own code was copy the use of `Result<()>`. But their code wouldn't compile and they couldn't figure out why. When they asked me, I said, "oh, you're probably missing a `type Result<T> = ...;` somewhere in your code." But then they pointed me to [the example in which they saw `Result<()>`](https://rust-lang-nursery.github.io/rust-cookbook/file/read-write.html#read-lines-of-strings-from-a-file), which does not have any such type alias.

It took me at least a few minutes of perplexed digging before I was actually able to figure out how that code was compiling. It turns out that the `error_chain!` macro is defining a `Result` type alias for you. Personally, I think this makes learning from these examples much harder than necessary. Some might even argue that the `Result` type alias [itself is problematic](https://github.com/rust-lang-nursery/failure/issues/95#issuecomment-361435699), but even aside from that argument, I think having it implicitly defined and brought into scope is really quite surprising, especially for beginners that might be trying to learn Rust through this cookbook.

Of course, this is just one experience report. In any case, what are the alternatives? Here are some that I can think of:

* Keep `error-chain`, but don't use the `Result` type alias. (What does one use instead that doesn't suffer from the same implicitness fault?)
* Switch to using `failure`.
* Switch to using only the facilities provided by the standard library.
* Remove proper error handling altogether and use `unwrap`/`panic`.

I don't think we should remove proper error handling. I think a cookbook that shows how to do error propagation is much more valuable than a cookbook that doesn't. And I think such examples are incredibly useful for beginners.

[...]

Issue #370, opened 2017-11-23
Consider examples for "failure" crate?


ludwigpacifici:
Recently, [Failure](https://crates.io/crates/failure) was announced (discussions [here](https://www.reddit.com/r/rust/comments/7dg95u/announcing_failure/), [also here](https://www.reddit.com/r/rust/comments/7b88qp/failure_a_new_error_management_story/)). Is it too early to think about recipes (equivalent to  [error-chain](https://crates.io/crates/error-chain)) for it?

budziq:
@ludwigpacifici Thanks for pointing this out!
Imho we should wait a little before Failure gets more traction and possibly matures a little.
The error-chain crate had a lot of time to mature and enjoyed wide adoption before it got included in cookbook. I'd not want to promote a crate before both community and author are confident enough in the crate API stability (which I'm guessing is quite early for Failure)

@withoutboats might have some thoughts on the subject as far as I know they are working with @aturon on proposed crate list for the cookbook.

AndyGauge:
So `failure` has taken the de facto error handling position in the ecosystem.  I think a rip and replace would be reasonable.  Here's something kinda official https://github.com/rust-lang-nursery/failure/issues/181

budziq:
While `failure` has gained some popularity I'm not entirely sure it has much development traction as its repo is not super active especially in regard to 1.0. Post the initial release it has been barely updated in crates.io with no releases since Nov 30, 2017 and it looks [barely more active](https://github.com/rust-lang-nursery/failure/graphs/code-frequency) than [error_chain](https://github.com/rust-lang-nursery/error-chain/graphs/code-frequency).

With `error_chain` still having 2x daily downloads than `failure` I don't think that we should just decide by ourselves that the idiomatic and production ready error handling in rust is now all about `failure`.
I'd like to invite wg-ecosystem to the discussion 
ping @KodrAus

budziq:
Thanks @Dylan-DPC!

I can see two issues here:
- `failure`'s design is in flux and I feel inclined towards waiting untill things take shape (the same reason why there is no tokio/async chapter yet)
- the issue raised in https://github.com/rust-lang-nursery/ecosystem-wg/issues/22. how do we choose the best/most idiomatic crate when there are several equivalent contenders?

mulkieran:
failure had a new release fairly recently: https://github.com/rust-lang-nursery/failure/releases/tag/0.1.3.

[...]

PR #565, opened 2019-12-19
Update to edition 2018

### LINKED ISSUE #530: Update examples to 2018 edition

### LINKED ISSUE #569: type annotations needed
error[E0282]: type annotations needed for `rusqlite::MappedRows<'_, [closure@/tmp/rust-skeptic.W3uKvyKglSl5/test.rs:41:42: 46:6]>`

  --> /tmp/rust-skeptic.W3uKvyKglSl5/test.rs:42:9

   |

41 |     let cats = stmt.query_map(NO_PARAMS, |row| {

   |         ---- consider giving `cats` the explicit type `rusqlite::MappedRows<'_, [closure@/tmp/rust-skeptic.W3uKvyKglSl5/test.rs:41:42: 46:6]>`, where the type parameter `E` is specified

42 |         Ok(Cat {

   |         ^^ cannot infer type for `E`

For more information about this error, try `rustc --explain E0282`.

test insert_select_sect_insert_and_select_data_line_8 ... FAILED

pickfire:
Update mdbook to 0.3.5

Command:
`sed -i '/extern crate/ {N;s/\n$//}' src/**.md; sed -i 's/extern crate error_chain;/use error_chain::error_chain;/; s/extern crate lazy_static;/use lazy_static::lazy_static;/; s/extern crate bitflags;/use bitflags::bitflags;/; s/extern crate serde_json;/ use serde_json::json;/; s/extern crate serde_derive;/use serde::{Serialize, Deserialize};/; /macro_use/d; /extern crate/ d; s/```rust/```rust,edition2018/; s/bail!/error_chain::&/; s/\(debug\|info\|warn\|error\)!/log::&/;' src/**.md`

Fixes #530
Fixes #569


pickfire:
@AndyGauge Interested to review?

AndyGauge:
Yeah, I was reading it and I'm impressed.  One thing I'm concerned about is if there are any `extern crate` definitions that the text doesn't include a reference.  One thing nice about having those extern crate statements is it makes it easy to set up the recipe's Cargo.toml.  I'll build it locally and peruse.

AndyGauge:
duh, there's crate badges right above the recipes.  That should be instructive enough.  I've peeked at it and from 30000 ft it looks good.

pickfire:
@AndyGauge I think the notable changes are `index.hbs`, the google analytics were removed I guess. Not sure if we need it?

pickfire:
@AndyGauge What should I do next for this?

AndyGauge:
I'm going to try and get it compiled and uploaded to review it in book format next.

AndyGauge:
OK, I've published it here http://www.yetanother.site/rust-cookbook/ and am reviewing

AndyGauge:
OK, so I'm now seeing new errors... it looks like error-chain does not like its result type being ignored.  

error[E0107]: wrong number of type arguments: expected 2, found 1
  --> /tmp/rust-skeptic.HOf7zmrK9Y0r/test.rs:12:14
   |
12 | fn main() -> Result<()> {
   |              ^^^^^^^^^^ expected 2 type arguments
error: aborting due to 2 previous errors
For more information about this error, try `rustc --explain E0107`.
test about_sect_a_note_about_error_handling_line_150 ... FAILED
error: cannot find macro `error_chain_processing` in this scope
  --> /tmp/rust-skeptic.qfSON6UfVh6g/test.rs:6:1
   |
6  | / error_chain! {
7  | |     foreign_links {
8  | |         Utf8(std::str::Utf8Error);
9  | |         AddrParse(std::net::AddrParseError);
10 | |     }
11 | | }
   | |_^

AndyGauge:
looks like we need to do this https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html
use error-chain::Result

AndyGauge:
In order to make Travis useful, I had to fix #569 to figure out what errors are caused by this change.

AndyGauge:
OK, here's the problem.  `error-chain` never was updated for Rust 2018 🤕 .  Now I can do that, but it looks like here https://github.com/rust-lang/rust/issues/57013 the way to fix error-chain for now is to `use` all of the internal macros.  

```
use error_chain::{error_chain, error_chain_processing, impl_error_chain_kind, impl_error_chain_processed, impl_extract_backtrace, ChainedError};
```

AndyGauge:
It is really hard to come up with consensus on error handling.  I'm biased, because I maintain error-chain (picked up both projects after Brian left Mozilla).  I think I would use `thiserror` in my own library at this point though.  Feel free to address error-handling in #502 or #370 
The primary reason we don't switch crates is there is no ecosystem stability right now.  The cookbook is all about curated, stable libraries that the library team originally approved.  Adding crates to the list should be a hard thing.  In fact if you want to showcase `anyhow` as a recipe I'd probably approve it, and throw a `thiserror` in there too, but for blanket switching all recipes I think that would give me an aneurysm.

pickfire:
@AndyGauge What do we need to do for this pull request?

AndyGauge:
We've discovered a major pain point using `error-chain` with 2018.  Either we need to use the internal macros, fix error-chain, or change the error handling library.  I believe I wrote those in order of complexity.

I doubt I will be working on this error-chain change immediately, but I think that would be a good solution.  If you would rather rewrite the error handling story of the cookbook let me know if you have that capacity.

AndyGauge:
Interestingly I discovered that error-chain 0.12 included the 2018 fix.  cargo downloads both error-chain 0.11 and 0.12, and incorrectly was using the 0.11 version.  

I feel like this is a skeptic problem.

I just recompiled this branch with rustc 1.45-nightly and had every single test fail.

AndyGauge:
OK, let's party like its 2018!

pickfire:
> I doubt I will be working on this error-chain change immediately, but I think that would be a good solution. If you would rather rewrite the error handling story of the cookbook let me know if you have that capacity.

Yes, I could rewrite the rust error handling story to use anyhow or thiserror.

Also, I would like change the global data story to use once_cell.

But I may not have time recently but should have some time to work this out in few weeks.

AndyGauge:
Turns out error chain was updated and the problem was with the dependency
graph. At this time we shouldnt rewrite the error handling. Thanks!

pickfire:
How about lazy_static? once_cell looks like a better alternative without using macro. https://docs.rs/once_cell/1.4.0/once_cell/#faq

AndyGauge:
Funny story.  @matklad was helping me with a pre-rfc when once_cell was born.  https://internals.rust-lang.org/t/pre-rfc-lazy-static-move-to-std/7993/36 <https://internals.rust-lang.org/t/pre-rfc-lazy-static-move-to-std/7993/36>

I’m fine with a recipe on it.  How about highlighting the regex use case?  I know @burtsushi describes that as a big part of using either of these, and creating a regex is a good introductory problem.  Like https://docs.rs/once_cell/1.4.0/once_cell/#building-block <https://docs.rs/once_cell/1.4.0/once_cell/#building-block> without creating a macro?

Andrew

pickfire:
> I’m fine with a recipe on it.  How about highlighting the regex use case?  I know @burtsushi describes that as a big part of using either of these, and creating a regex is a good introductory problem.  Like https://docs.rs/once_cell/1.4.0/once_cell/#building-block <https://docs.rs/once_cell/1.4.0/once_cell/#building-block> without creating a macro?

https://docs.rs/once_cell/1.4.0/once_cell/#building-block is creating a macro itself and regex already uses `lazy_static`, we would have to replace it with `once_cell` and can also explain the benefits of using `once_cell` such that one could do lazy initialization without macro.

PR #567, opened 2019-12-31
fix partial download code example

bernhardfritz:
code example did not work right away, had to add blocking:: to make it work.


AndyGauge:
I see how the updated crate is now using async ( #571 ).  We should probably highlight how to do things the async way rather than showcasing the synchronous, blocking method.  Its a lot more work though.

AndyGauge:
OK, so I looked at implementing async in a cookbook fashion and I really think we should decouple demonstrating reqwest and async.  I think the blocking style of this PR is the right way to introduce `reqwest` in the cookbook

AndyGauge:
We've updated the examples.  This example has been updated.  Thank you for your contribution!

Issue #573, opened 2020-02-24
Does not compile with current ring crate v0.16.11 - Sect. 5.1 "Sign and verify a message with HMAC digest"


quaesaevum:
Section 5.1 does not compile, apparently due to name changes in ring ^0.16.1.

Updates appear to be only the names in the ring crate.

quaesaevum:
Locally this works:
```
use ring::{hmac, rand};
use ring::rand::SecureRandom;
use ring::error::Unspecified;

fn main() -> Result<(), Unspecified> {
    let mut key_value = [0u8; 48];
    let rng = rand::SystemRandom::new();
    rng.fill(&mut key_value)?;
    let key = hmac::Key::new(hmac::HMAC_SHA256, &key_value);

    let message = "Legitimate and important message.";
    let signature = hmac::sign(&key, message.as_bytes());
    hmac::verify(&key, message.as_bytes(), signature.as_ref())?;

    Ok(())
}
```

AndyGauge:
care to submit a PR?

quaesaevum:
Opened PR.

First time contributor - please advise, admonish, etc. as needed.

PR #574, opened 2020-02-25
Fix issue #573 - hmac example

quaesaevum:
fixes issue #573 

Questions/concerns:
Section does not pass test under cargo test, but runs fine locally, on Rust playground, and under 'mdbook serve' on localhost.  It appears testing may refer to an older version of the ring crate.  Which version of ring does 'cargo test' call?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ rand = "0.6"
 rayon = "1.0"
 regex = "1.0"
 reqwest = "0.9"
-ring = "0.13.0-alpha"
+ring = "0.16.11"
 rusqlite = { version = "0.16", features = ["chrono"] }
 same-file = "1.0"
 select = "0.4"
--- a/src/concurrency/thread/threadpool-walk.md
+++ b/src/concurrency/thread/threadpool-walk.md
@@ -1,12 +1,12 @@
-## Calculate SHA1 sum of iso files concurrently
+## Calculate SHA256 sum of iso files concurrently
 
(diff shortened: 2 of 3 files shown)


AndyGauge:
This is great, but you want to also update the Cargo.toml file https://github.com/rust-lang-nursery/rust-cookbook/blob/master/Cargo.toml#L43
`cargo test` uses the cookbooks Cargo.toml file.

quaesaevum:
Updated Cargo.toml as well.

AndyGauge:
So, it looks like this breaks https://rust-lang-nursery.github.io/rust-cookbook/concurrency/threads.html#calculate-sha1-sum-of-iso-files-concurrently due to:
error[E0432]: unresolved import `ring::digest::SHA1`
  --> /tmp/rust-skeptic.WtyY3w319zV3/test.rs:13:37
   |
13 | use ring::digest::{Context, Digest, SHA1};
   |                                     ^^^^ no `SHA1` in `digest`
For more information about this error, try `rustc --explain E0432`.
test threadpool_walk_sect_calculate_sha1_sum_of_iso_files_concurrently_line_11 ... FAILED

Seems that SHA1 has been deprecated so we ought to highlight SHA-256.

Let me know if you are comfortable with that or you need me to help.

quaesaevum:
Gave it a shot.

I'm still seeing one error, but it seems unrelated to any of my changes.

AndyGauge:
Yeah, I have that fixed in the 2018 edition update.  I should probably separate it now that 2018 looks like a chore.

AndyGauge:
Thank you
!

PR #575, opened 2020-02-26
async reqwest with tokio

### LINKED ISSUE #571: reqwest examples are outdated
Since reqwest 0.10.1, it uses `async` by default, making most examples in the cookbook outdated.

I'll try and make a PR, but I do not think that have sufficient experience with the crate. I will also try to contact the project maintainers.

AndyGauge:
Fixes #571


src/web/clients/requests/get.md
```diff
@@ -32,9 +28,33 @@ fn main() -> Result<()> {
...
+
+A similar approach can be used by including the [`tokio`] executor
+to make the main function asynchronous, retrieving the same information.
+
+In this example, [`tokio::main`] handles all the heavy executor setup
+and allows sequential code implemented without blocking.
+
+```rust,no_run,edition2018
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error >> {
```
AndyGauge: So Box<dyn std::error::Error> return type seems like cheating.  Is there a more specific error type that can be used here?

JOE1994: `Tokio` team seems to be using the return type `Box<dyn std::error::Error>` as a norm;
All [Example programs from the Tokio repo](https://github.com/tokio-rs/tokio/tree/master/examples) are using the same return type
`Box<dyn std::error::Error>`. I think we can leave this unchanged for now. 😃

AndyGauge:
figured out proper error handling.  Not much changed luckily!

PR #604, opened 2020-06-13
read environment variable

### LINKED ISSUE #401: Open Config file from given environment variable location

KappaDistributive:
fixes #401

### DIFF
--- a/src/os/external/read-env-variable.md
+++ b/src/os/external/read-env-variable.md
@@ -1,0 +1,25 @@
+## Read Environment Variable 
+
+[![std-badge]][std] [![cat-os-badge]][cat-os]
+
+Reads an environment variable via [std::env::var].
+
+```rust,edition2018,no_run
+use std::env;
+use std::fs;
+use std::io::Error;
+
+fn main() -> Result<(), Error> {
+    // read `config_path` from the environment variable `CONFIG`.
+    // If `CONFIG` isn't set, fall back to a default config path.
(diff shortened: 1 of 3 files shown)


src/os/external/read-env-variable.md
```diff
@@ -0,0 +1,25 @@
...
+```rust,edition2018,no_run
+use std::env;
+use std::fs;
+use std::io::Error;
+
+fn main() -> Result<(), Error> {
+    // read `config_path` from the environment variable `CONFIG`.
+    // If `CONFIG` isn't set, fall back to a default config path.
+    let config_path = env::var("CONFIG")
+        .unwrap_or("<default config path>".to_string());
```
AndyGauge, line 16: love the unwrap_or usage here, but lets give it a static path, like /etc/config so at a glance it is obvious how to repurpose the recipe.

AndyGauge:
Looks good.  Glad to see some new recipes.  Can you just make one quick change for me and demonstrate using an actual path?

src/os/external/read-env-variable.md
KappaDistributive: Sure thing :wink:

Issue #436, opened 2018-07-18
Recipe Series: Web Authentication


AndyGauge:
|  |  |
| ------------- | ------------- |
| Crates  | reqwest  |
| Section  | 17.4.3 Web Authentication  |

A new series for Web Authentication should be created.  This series needs 3 recipes for the following types of authentication:

- [ ] [Basic Authentication](https://docs.rs/reqwest/*/reqwest/struct.RequestBuilder.html#method.basic_auth)
- [ ] Digest Authentication (On Hold until https://github.com/hyperium/headers imported)
- [ ] Authentication for HTTPS request

A good option for demonstrating authentication is [httpbin.org](http://httpbin.org/#/Auth)

[Thank you @fawix](https://github.com/rust-lang-nursery/rust-cookbook/issues/84#issuecomment-354306864)

Immortalin:
Also server-side auth: user registration, login/logout, confirmation emails, the various different token based authentication methods and protocols, multifactor (e.g. U2F), email-only auth, securing APIs etc.

PR #605, opened 2020-06-14
demonstrate basic http authentication

KappaDistributive:
Part of #436 

Perform a basic HTTP authentication with `reqwest`'s blocking client.

If you'd prefer to use an asynchronous client instead, just say the word. (Personally, I don't have a strong preference for the sake of this example in either direction. But ultimately opted to go with the synchronous client to minimize the complexity of this example and purely focus it on the authentication aspect.)


AndyGauge:
thx

Issue #563, opened 2019-12-17
Remove hidden code


AndyGauge:
Let's expand all the recipes.  Hidden code is a stumbling block for far too many people.

PR #606, opened 2020-06-15
596 headers

### LINKED ISSUE #596: reqwest::headers
We ignored and filtered the headers recipe, but now `reqwest` headers work.  

https://github.com/rust-lang-nursery/rust-cookbook/blob/master/src/web/clients/requests/header.md

https://stackoverflow.com/a/47912021/5586783
https://docs.rs/reqwest/0.10.6/reqwest/header/index.html

KappaDistributive:
I've made a few changed to the original recipe:

- made the usage of `error_chain` visible to account for #563 
- switch to the synchronous reqwest client to simplify the recipe
- removed the authorization header since this is now handled via [reqwest::RequestBuilder::bearer_auth](https://docs.rs/reqwest/0.10.6/reqwest/blocking/struct.RequestBuilder.html#method.bearer_auth)
- got rid of the `hyper::header!` macro
- gave the `X-Powered-By` header a more generic/sterile value
- added missing links

fixes #596

### DIFF
--- a/src/web.md
+++ b/src/web.md
@@ -31,6 +31,7 @@
 | Recipe | Crates | Categories |
 |--------|--------|------------|
 | [Make a HTTP GET request][ex-url-basic] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
+| [Set custom headers and URL parameters for a REST request][ex-url-header] | [![reqwest-badge]][reqwest] | [![cat-net-badge]][cat-net] |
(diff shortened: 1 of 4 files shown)


KappaDistributive:
Please let me now if you'd like to see any changes.

The decision I'm least confident about is the ad-hoc creating of the custom header.

stappersg (commented):
LGTM

AndyGauge:
Thanks, this looks good.

PR #619, opened 2020-11-19
Fix rand_distr version in algorithms examples

ekzhang:
Hi, I noticed when reading the Cookbook that rand_distr was labeled as v0.7.3, whereas its actual version is v0.3.0. Just fixing this by adjusting a badge link.


ekzhang:
@AndyGauge Just wanted to bump this, in case you missed my PR. The visual bug is still here and it's a one-line change to fix.

image

(should say rand_distr v0.4.0)

ekzhang:
@AndyGauge It's been over half a year, so I'm bumping this one more time. Typo is still present on the front page of rust-cookbook.

image

This is my last bump, so if you don't reply to this, I'm unsubscribing to the PR and won't mention you again.

PR #623, opened 2020-12-23
fix random range generator gen_range call

rahulunair:
fixes issue with `gen_range` call using `rand` crate 0.8.0, with API changes, `gen_range` argument should be a range.


AndyGauge:
seems this is resolved with another PR, failed to get them in order.

PR #636, opened 2021-02-03
Use edition 2018 everywhere

MaxSem:


AndyGauge:
Hmm..  Not sure.  I like skeptic.  How does this help?

AndyGauge (approved):
I built locally with it and it checks out.  Seems skeptic is running in 2021

PR #639, opened 2021-02-24
Reduce calling of `to_string()`

zzy:
- [x] the tests are passing locally with `cargo test`
- [x] cookbook renders correctly in `mdbook serve -o`


AndyGauge:
This is way easier to read.

PR #640, opened 2021-03-15
fix incorrect badge link

Jaxelr:
The badge of the ansi_term package is pointing incorrectly to the base64 crate, this is causing that the ansi_term package shows a non existent version

image


djc (approved):

stappersg (approved):
Thanks for catching this.

PR #641, opened 2021-03-17
Convert u8 to char (necessary since rand 0.8)

djc:
This fixes one breaking skeptic test due to a change in the type of `Alphanumeric` between rand 0.7 and rand 0.8.


stappersg (commented):
LGTM

v-i-s-h:
I think this PR needs to be merged. I could not understand why the checks failed. The trace seems to be from something unrelated.

AndyGauge:
Thanks @djc

PR #654, opened 2022-01-15
Add "Encode and decode base32" example

ia0:
The 2 following tests are failing at master with `cargo test` and are the only failing at this PR:
- `rand_choose_sect_create_random_passwords_from_a_set_of_user_defined_characters_line_8`
- `rand_range_sect_generate_random_numbers_within_a_range_line_7`

The link checker fails at master and still fails at this PR with the same error (I made sure to have the same version of mdbook and link-checker as described in CONTRIBUTING.md):
```
IsADirectoryError: [Errno 21] Is a directory: '[...]/rust-cookbook/book/'
```

I decided to not create an issue for this PR since it's mostly a copy/paste from the "Encode and decode hex" example. Both use the same crate which is a generic library over hex, base32, base64, etc.

All other items in the PR list are taken into account without issue.


src/encoding.md
```diff
@@ -5,6 +5,7 @@
 | [Percent-encode a string][ex-percent-encode] | [![percent-encoding-badge]][url] | [![cat-encoding-badge]][cat-encoding] |
 | [Encode a string as application/x-www-form-urlencoded][ex-urlencoded] | [![url-badge]][url] | [![cat-encoding-badge]][cat-encoding] |
 | [Encode and decode hex][ex-hex-encode-decode] | [![data-encoding-badge]][data-encoding] | [![cat-encoding-badge]][cat-encoding] |
+| [Encode and decode base32][ex-base32] | [![data-encoding-badge]][data-encoding] | [![cat-encoding-badge]][cat-encoding] |
```
AndyGauge: Can the recipe title be something like Generate OTP with a Base32 set?

src/encoding/string/base32.md
```diff
@@ -0,0 +1,73 @@
...
+
+    // Reconstitute the random bytes (correcting the visual errors).
+    let final_entropy = read_entropy(&entered_password)?;
+    assert_eq!(final_entropy, initial_entropy);
+
+    Ok(())
+}
+
+fn make_password(entropy: &[u8]) -> String {
+    BASE32_NOPAD.encode(&entropy)
```
AndyGauge:
```suggestion
    BASE32_NOPAD.encode(entropy)
```
would trigger clippy::needless_borrow

AndyGauge:
Appreciate the changes, just 2 minor nits

src/encoding/string/base32.md
ia0: oops, that's a refactoring left-over

PR #655, opened 2022-01-18
Added ImageError handler to error_chain! macro.

Jorge-Gonzalez:
It is not compiling because of:
`?` couldn't convert the error to `Error`
   --> src/main.rs:90:35
    |
67  | fn main() -> Result<()> {
    |              ---------- expected `Error` because of this
...
90  |     let _ = img.save("output.png")?;
    |                                   ^ the trait `From<ImageError>` is not implemented for `Error`


AndyGauge (approved):
error_chain turns out to be picky

PR #663, opened 2022-04-01
querying-the-web-api examples don't work due to missing user-agent-headers

### LINKED ISSUE #662: querying-the-web-api examples don't work due to missing user-agent-headers
Github API requires `user-agent-header` and without it, it returns a `403` reponse, ultimately resulting in vague decoding errors that could be difficult to debug for trying out the examples. 

For instance, the code snippet in https://rust-lang-nursery.github.io/rust-cookbook/web/clients/apis.html#query-the-github-api results in following error.

### 403 due to missing user-agent-header
```
https://api.github.com/repos/rust-lang-nursery/rust-cookbook/stargazers
[web-api-client/src/bin/web-api-client.rs:19] &response = Response {
    url: Url {
        scheme: "https",
        cannot_be_a_base: false,
        username: "",
        password: None,
        host: Some(
            Domain(
                "api.github.com",
            ),
        ),
        port: None,
        path: "/repos/rust-lang-nursery/rust-cookbook/stargazers",
        query: None,
        fragment: None,
    },
    status: 403,
    headers: {
        "cache-control": "no-cache",
        "content-type": "text/html; charset=utf-8",
        "strict-transport-security": "max-age=31536000",
        "x-content-type-options": "nosniff",
        "x-frame-options": "deny",
        "x-xss-protection": "0",
        "content-security-policy": "default-src 'none'; style-src 'unsafe-inline'",
        "connection": "close",
    },
}
Error: reqwest::Error { kind: Decode, source: Error("expected value", line: 2, column: 1) }

```

### Solution:

One of the fixes for above is to use `reqwest client` and pass in `user-agent-header` along with the request such as below.

```rust
 // ref: https://stackoverflow.com/questions/47911513/how-do-i-set-the-request-headers-using-reqwest
    use reqwest::header::USER_AGENT;

    let client = reqwest::Client::new();
    let response = client
        .get(request_url)
        .header(USER_AGENT, "rust web-api-client demo")
        .send()
        .await?;
```

Related: https://github.com/seanmonstar/reqwest/issues/1516

suren-m:
Fixed runtime decoding error occurring due to 403 responses from `gh api` because of missing user-agent header

fixes #662 

Feel free to close if you still prefer to use `reqwest::get` for the example but with a different api that doesn't require `user-agent-header`. I'm currently unaware of how to pass headers along with `reqwest::get` and thus used a `reqwest client` in the fix

### DIFF
--- a/src/web/clients/api/rest-get.md
+++ b/src/web/clients/api/rest-get.md
@@ -12,6 +12,7 @@ processing the response into User instances.
 ```rust,edition2018,no_run
 use serde::Deserialize;
 use reqwest::Error;
+use reqwest::header::USER_AGENT;
 
 #[derive(Deserialize, Debug)]
 struct User {
@@ -25,7 +26,13 @@ async fn main() -> Result<(), Error> {
                               owner = "rust-lang-nursery",
                               repo = "rust-cookbook");
     println!("{}", request_url);
-    let response = reqwest::get(&request_url).await?;
+    
+    let client = reqwest::Client::new();
+    let response = client
+        .get(request_url)
+        .header(USER_AGENT, "rust-web-api-client") // gh api requires a user-agent header
+        .send()
+        .await?;
 
     let users: Vec<User> = response.json().await?;
     println!("{:?}", users);


toanju:
Ran into the same issue. Thanks for writing this up. Worked like a charm. I wish it was already merged ;)

toanju (approved):

tompscanlan:
Also hit the same thing and this is better work than I came up with :)

AndyGauge (approved):
demonstrating headers is pretty foundational with web get anyway.  Love the improvement.

Issue #634, opened 2021-01-21
Broken Mathematics Links on the Intro Page


davehadley:
Links on the front page <https://rust-lang-nursery.github.io/rust-cookbook/intro.html> in the mathematics section link to pages with a 404 error.

For example the vector norm link is:
<https://rust-lang-nursery.github.io/rust-cookbook/mathematics/linear_algebra.html#vector-norm>

Links on the mathematics page itself (<https://rust-lang-nursery.github.io/rust-cookbook/science/mathematics.html>) seem to work correctly.
For example here the vector norm link here is:
<https://rust-lang-nursery.github.io/rust-cookbook/science/mathematics/linear_algebra.html#vector-norm>
which works for me.

Is mathematics a science? 🤔
It seems the rust-cookbook can't make up its mind! 😄

PR #677, opened 2022-11-22
fixes #671 - Development_tools/debugging relative links

### LINKED ISSUE #671: Some documents are unavailable
for example
https://rust-lang-nursery.github.io/rust-cookbook/development_tools/development_tools/debugging/log.html#log-a-debug-message-to-the-console
gives 404

Fax:
fixes #671
Same issue as in #634, links in md files included in md files from other folders will have wrong relative path in one of the two locations. in this case the root file development_tools.md was working and the debugging.md was broken.

### DIFF
--- a/src/development_tools.md
+++ b/src/development_tools.md
@@ -1,31 +1,52 @@
 # Development Tools
 
-{{#include development_tools/debugging.md}}
+## Debugging
+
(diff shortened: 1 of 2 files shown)


AndyGauge (approved):
thanks for the report

PR #680, opened 2022-12-06
chore: update `rand_distr` badge url

alex-pakalniskis:
Previous badge url display version numbers associated with `rand` crate which do not map to `rand_distr` versions. 

Current: https://badge-cache.kominick.com/crates/v/rand.svg?label=rand_distr
Proposed: https://badge-cache.kominick.com/crates/v/rand_distr.svg?label=rand_distr


AndyGauge (approved):
thank you

PR #690, opened 2024-01-04
net/server/listen-unused: simpler and with IPv6 support.

kmkaplan:
Uses ToSocketAddrs' trait to have any of IPv4 or IPv6 as supported by the OS, or even DNS names.


AndyGauge (approved):
Thanks for fixing my code.

PR #694, opened 2024-04-08
updated and repaired clap-basic.md

### LINKED ISSUE #693: clap-basic.md is stale, wrong
`clap-basic.md` contains very stale example code, and also handles its filename incorrectly. The several turbofish in there are also an issue.

BartMassey:
* Brought example code to clap version 4, noted in doc.
* Removed turbofishes from example code.
* Changed to use PathBuf for filenames; documented this.

fixes #693


src/cli/arguments/clap-basic.md
```diff
@@ -6,39 +6,55 @@ This application describes the structure of its command-line interface using
 `clap`'s builder style. The [documentation] gives two other possible ways to
 instantiate an application.
 
-In the builder style, `with_name` is the unique identifier that `value_of` will
-use to retrieve the value passed. The `short` and `long` options control the
+The code here requires `clap` version 4. Use `clap = "4"` in
```
AndyGauge, line 9: the cargo.toml is included as a reference.  Rather than this recommendation, update the Cargo.toml file so we can ensure the build passes.

BartMassey: Does the below work? Most of the `skeptic` tests seem to fail due to some kind of misconfiguration so I don't know what I'm looking for here.

AndyGauge (approved):
Thanks!  `skeptic` issues you say.  That's not the first 😅

PR #698, opened 2024-06-06
Update tar-compress docs

schneems:
- Call finish to finalize the archive [recommended by tar](https://docs.rs/tar/0.4.41/tar/struct.Builder.html#method.append_dir_all).
- Show an example of adding an archive **without** renaming it.
- Callout differences between `tar::Builder` defaults and `tar(1)` defaults

## Context

On seeing these docs I was unsure if there was another, better way, to add all the contents of one directory into an archive. After researching, I see people use this function by call it with an empty string. I feel this is a common operation and would like to see it spelled out in the documentation.

In addition, I hit an edge case where `tar(1)` produced significantly smaller files than the `tar` crate, reproduction: https://github.com/schneems/tar_comparison/blob/bfd420a012b46e80435cf4e7c67ca1661357fde3/README.md. It turns out that the problem was that the directory contained symlinks to other files in the same directory. The `follow_symlinks(true)` behavior is on by default and is the opposite of the `tar(1)` default. It caused the program to duplicate the same file multiple times which significantly increased the archive size. I believe someone looking for "How to  Compress a directory into tarball" is likely looking to replicate `tar czf` behavior and would like to know the main differences.

### DIFF
--- a/src/compression/tar/tar-compress.md
+++ b/src/compression/tar/tar-compress.md
@@ -6,7 +6,7 @@ Compress `/var/log` directory into `archive.tar.gz`.
 
 Creates a [`File`] wrapped in [`GzEncoder`]
 and [`tar::Builder`]. </br>Adds contents of `/var/log` directory recursively into the archive
-under `backup/logs`path with [`Builder::append_dir_all`].
+under `backup/logs` path with [`Builder::append_dir_all`].
 [`GzEncoder`] is responsible for transparently compressing the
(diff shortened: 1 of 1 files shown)


AndyGauge (approved):
I like the easy recipe addition, it actually is more useful

schneems:
Thanks!

AndyGauge:
oh hey I remember meeting you outside railsconf picking up my badge.  Appreciate the help.

PR #700, opened 2024-11-25
Corrected impossible example in side-length.md

TymonM:
A triangle with a right angle (1.57 radians) and an angle of 2 radians cannot exist and is physically impossible.


AndyGauge (approved):
Yeah, that makes sense.  It would be better to include _about 57 degrees_ to the description so it clicks why 1 radian is a better approach.

PR #703, opened 2024-11-29
change edition 2024

AndyGauge:
```
rustup default nightly

mdbook serve --open
```

Playground detects 2015 version, doesn't even support 2021 version.


AndyGauge:
Migrate recipes one by one to the edition rather than in bulk.  We need to write and verify crates/features, and make it easier to use

PR #737, opened 2025-10-02
Move algorithms/randomness to crates

### LINKED ISSUE #711: Move examples to rust projects (remove skeptic)
The current approach to writing examples is to write the code directly in the markdown files an then combine and use skeptic for testing.

The downside of this is that it's more difficult to edit the code using the standard rust tooling (rust-analyzer, rustfmt, clippy, etc.)

A better approach (@jamesgraves did this in https://github.com/jamesgraves/rust-cookbook/) is to move all the code into actual projects and then import from files into mdbook (see https://rust-lang.github.io/mdBook/format/mdbook.html#including-files)

Doing this means that in general you can write code, lint it and format it, and then view the code snippet in the book directly. It also means that each snippet can be supported by the relevant cargo manifest necessary just for that snippet rather than a manifest for the entire cookbook, and it makes it possible to run each example directly when you clone the repo.

## Implementation

We use this approach successfully in the [Ratatui](https://ratatui.rs/) website. This started as an mdbook site and then we moved to using [Starlight](https://starlight.astro.build/) (a doc site template for [Astro](https://astro.build/)). To do this we added the same idea of importing code from files to the stock code block element, and then went a little further to reduce the amount of `// ANCHOR` comments needed for importing rust functions using tree sitter to parse out just functions of interest. This might be possible to backport into mdbook.

## Minor side note (code organization)

I'm not 100% in agreement with the arrangement of where the rust projects end up in the tree in @jamesgraves version. They feel a bit jumbled together which is difficult to navigate. In Ratatui, we had a second root `code` folder with just the crates / code. This observation is a minor one however, as I think it's much better to have this functionality than to bikeshed on exactly how it's organized. Perhaps there's some hybrid where each crate has a `src` folder with a `docs` folder alongside it. That would make it feel less jumbled (I'm not whether it's possible with mdbook to have the file location and url not fully aligned like this)

joshka:
This takes a slightly different approach to the way that the web crate works by moving all rust code to proper rust files and checking if the code compiles in CI.

Related to but does not completely fix #711

--------------------------
### Things to check before submitting a PR

- [ ] the tests are passing locally with `cargo xtask test all`
  - existing tests are broken
- [x] PR contains correct "fixes #ISSUE_ID" clause to autoclose the issue on PR merge
    -  if issue does not exist consider creating it or remove the clause
- [x] non rendered items are in sorted order (links, reference, identifiers, Cargo.toml)
- [x] links to docs.rs have wildcard version `https://docs.rs/tar/*/tar/struct.Entry.html`
  - latest instead
- [ ] example has standard [error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling)
- [x] code identifiers in description are in hyperlinked backticks
```markdown
[`Entry::unpack`]: https://docs.rs/tar/*/tar/struct.Entry.html#method.unpack
```

### Things to do after submitting PR
- [ ] check if CI is happy with your PR


.github/workflows/mdbook-test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Set up Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          cache: 'true'
+          cache: "true"
           toolchain: stable
```
joshka, line 20: This config is actually unnecessary as these are the default values

crates/algorithms/randomness/src/bin/choose.rs
```diff
@@ -0,0 +1,19 @@
...
+                        abcdefghijklmnopqrstuvwxyz\
+                        0123456789)(*&^%$#@!~";
+const PASSWORD_LEN: usize = 30;
+
+fn main() {
+    let mut rng = rand::rng();
+
+    let password: String = (0..PASSWORD_LEN)
+        .map(|_| {
+            let idx = rng.random_range(0..CHARSET.len());
```
joshka, line 13: Fixed for rand 0.9 (`gen_range` -> `random_range`).

Moving away from skeptic here highlights deprecated code like this easily with standard tooling

crates/algorithms/randomness/src/bin/choose.rs
joshka, line 1: Note these files are in `src/bin/`. An alternative would be to put these in `tests/` and have them actually run when running tests. This would require adding `#[test]` to each main function. I'm ambivalent as to which option is better with a slight preference for src/bin, but if running each example to ensure that it doesn't error when run has a higher precedence than I've assumed then `tests/` with `#[test]` might be more convenient.

crates/algorithms/randomness/src/bin/passwd.rs
```diff
@@ -0,0 +1,13 @@
+use rand::Rng;
+use rand_distr::Alphanumeric;
+
+fn main() {
+    const PASSWORD_LEN: usize = 30;
+    let mut rng = rand::rng();
+
+    let password: String = (0..PASSWORD_LEN)
+        .map(|_| rng.sample(Alphanumeric) as char)
+        .collect();
```
joshka, line 10: The example from where this was copied was not using Alphanumeric, but the text said it was. Fixed this in this change.

src/algorithms/randomness/rand-choose.md
```diff
@@ -2,29 +2,11 @@
 
 [![rand-badge]][rand] [![cat-os-badge]][cat-os]
 
-Randomly generates a string of given length ASCII characters with custom
-user-defined bytestring, with [`gen_range`].
+Randomly generates a string of given length ASCII characters with custom user-defined bytestring,
+with [`random_range`].
```
joshka, line 6: In general, I've rewrapped markdown at 100 chars to make it easier to read when editing. I'd recommend choosing a value (likely 100) and adding a markdownlint config file for this.

src/algorithms/randomness/rand-passwd.md
```diff
@@ -2,32 +2,11 @@
 
 [![rand-badge]][rand] [![cat-os-badge]][cat-os]
 
-Randomly generates a string of given length ASCII characters in the range `A-Z,
-a-z, 0-9`, with [`Alphanumeric`] sample.
+Randomly generates a string of given length ASCII characters in the range `A-Z, a-z, 0-9`, with
+[`Alphanumeric`] sample.
```
joshka, line 6: The example below was incorrect according to the description here.

build.rs
```diff
@@ -5,27 +5,17 @@ const REMOVED_TESTS: &[&str] = &[
...
-            "./src/algorithms/randomness/rand-range.md",
-            "./src/algorithms/randomness/rand-dist.md",
-            "./src/algorithms/randomness/rand-custom.md",
-            "./src/algorithms/randomness/rand-passwd.md",
-            "./src/algorithms/randomness/rand-choose.md",
-        ];
-        skeptic::generate_doc_tests(&rand_paths[..]);
-        return;
-    }
+const REMOVED_PREFIXES: &[&str] = &["./src/algorithms/randomness/"];
```
joshka, line 8: Went with a prefix approach here as it makes it easy to crank down the folders, e.g once algorithms/sorting is also done, this can just include `./src/algorithms/sorting`.

Cargo.toml
```diff
@@ -1,18 +1,35 @@
+[workspace]
+members = ["crates/algorithms/*", "crates/web"]
+
+[workspace.package]
+edition = "2024"
```
joshka, line 5: Going with the latest edition here for the new code. Thoughts / problems?

Cargo.toml
```diff
@@ -1,18 +1,35 @@
...
+license = "MIT OR Apache-2.0"
+publish = false
+
 [package]
 name = "rust-cookbook"
 version = "1.1.0"
 authors = ["Brian Anderson <banderson@mozilla.com>", "Andrew Gauger <andygauge@gmail.com>"]
 edition = "2018"
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
```
joshka, line 16: This format was incorrect

Cargo.toml
```diff
@@ -1,18 +1,35 @@
...
 publish = false
 build = "build.rs"
 
 [features]
 default = []
-test-rand = []
 
+[workspace.dependencies]
+rand = "0.9"
+rand_distr = "0.5"
```
joshka, line 25: These won't yet affect the playground, but could reasonably easily by modifying the following code to include workspace deps:

https://github.com/rust-lang/rust-playground/blob/f8d7de52a3c139a0df4fe116bbbff19c30668c99/top-crates/src/lib.rs#L134-L161

Cargo.toml
```diff
@@ -1,18 +1,35 @@
...
+
+# NOTE: These dependencies add dependencies to the rust playground (play.rust-lang.org).
+# Be wary removing or changing these without considering this fact.
+# See https://github.com/rust-lang/rust-playground/blob/f8d7de52a3c139a0df4fe116bbbff19c30668c99/top-crates/src/lib.rs#L134-L161
```
joshka, line 29: Added this as it's tempting to remove the deps from here as they are not needed (rand / rand-distr are no longer included in the generated skeptic tests).

Cargo.toml
```diff
@@ -1,18 +1,35 @@
...
 [dependencies]
```
joshka, line 30: Re-sorted this section correctly

joshka (commented):
Self review notes.

crates/algorithms/randomness/src/bin/choose.rs
AndyGauge: **Praise:** This makes the recipes and tests synchronized and there's no extra boiler plate

crates/algorithms/randomness/src/bin/choose.rs
AndyGauge: **Praise:** isolating our dependencies makes them much more deterministic.

src/algorithms/randomness/rand-choose.md
AndyGauge: I'd accept a lint for it, I know the feeling about wanting to wrap lines.

AndyGauge (approved):
I will see about not making ci required so we can make changes.

Cargo.toml
AndyGauge: **Praise:** Thank you for fixing that

crates/algorithms/randomness/src/bin/choose.rs
joshka: This is probably the main comment that I wanted feedback on. Do you want to run the code as part of the build, or is confirming that the code compiles enough?

The old rust adage of if it compiles it's correct is probably good enough for most things that would appear in the cookbook, but it could be reasonable to instead just write normal unit tests that show both the code and the expected output of things like this. I.e. instead of `println!("...")` in a `main()` function, we'd have have a test with an assertion that shows what's expected. This change would be useful for those that understand testing well, but it's a fairly significant difference to the way things are currently.

Issue #760, opened 2026-04-18
Expand Command Line section with clap derive, subcommands, and env vars


AndyGauge:
## Why

Data-driven gap analysis against `crates.io` recent 90-day downloads (April 2026):

- `command-line-interface` category: ~1.1B recent downloads
- Current cookbook coverage: `cli/arguments.md` demonstrates the pre-4.0 clap builder pattern only

`clap` is the de facto CLI crate for Rust and used by nearly every application. The current recipe doesn't reflect how anyone writes new CLIs today (derive API) or the common patterns that follow (subcommands, env vars, validation).

## Proposed recipes

Expand the `Command Line` section:

- `clap` derive API — `#[derive(Parser)]` for a typed args struct
- Subcommands — `#[derive(Subcommand)]` with per-command args
- Read arguments from environment variables — `env` attribute
- Argument validation — `value_parser`, ranges, enum args
- Global flags shared between subcommands

## Contributor guidance

- New dependency work, or richer examples than the root `Cargo.toml` exposes, should go in a **standalone crate** under `crates/` — see `CLAUDE.md` and `crates/web_leptos/` for the pattern. Derive-style CLI examples are especially awkward inline, so a standalone crate is the idiomatic home.
- No `unwrap` — propagate with `?`. See [a note about error handling](https://rust-lang-nursery.github.io/rust-cookbook/about.html#a-note-about-error-handling).
- `docs.rs` links use wildcards: `https://docs.rs/clap/*/clap/...`.
- Hyperlinked backticks for code identifiers.
- Full checklist in `.github/PULL_REQUEST_TEMPLATE.md`.

PR #799, opened 2026-06-02
Add basic cli argument parsing + env variables

muchuiwilliam:
Addresses:
"Read arguments from environment variables — env attribute" from #760

### DIFF
--- a/src/cli/arguments/basic.md
+++ b/src/cli/arguments/basic.md
@@ -188,7 +188,7 @@ cargo run --release -- -l -p src/main.rs
 [`args`]: https://doc.rust-lang.org/std/env/fn.args.html
 [`count()`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count
 [`env`]: https://doc.rust-lang.org/std/env/index.html
-[`lines()`]: https://doc.rust-lang.org/std/io/trait.BufRead.html#method.lines
+[`lines()`]: https://doc.rust-lang.org/std/primitive.str.html#method.lines
 [`read_to_string()`]: https://doc.rust-lang.org/std/fs/fn.read_to_string.html
 [`skip(1)`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip
 [`split_whitespace()`]: https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace
--- a/src/cli/arguments/basic.md
+++ b/src/cli/arguments/basic.md
@@ -1,0 +1,195 @@
+## Basic Argument Parsing
+This section covers a basic implementation of commandline argument parsing.
+Use [`args`] to get the arguments this process was started with:
+
+```rust,edition2024
+use std::env;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    assert!(!args.is_empty());
+
+    // Skip the program name, it's usually the first argument
+    let args: Vec<String> = env::args().skip(1).collect();
+}
+```
+
+Some args, e.g. file paths, may not be valid UTF-8. Use [`args_os`] to handle those safely:
+
+```rust,edition2024
+use std::{env, ffi::OsString};
+
+fn main() {
+    let os_args: Vec<OsString> = env::args_os().collect();
+    assert!(!os_args.is_empty());
+}
+```
+
+## Calculate File Stats
+In this section we parse CLI args and compute per-file stats (line/word counts). A mini [`wc`]-like
+utility
+
+Flags:
+- --lines, -l Count lines per file
+- --words, -w Count words per file
+- --paths, -p Remaining args treated as file paths
+
+> `-p` acts as a terminator. Everything after it is consumed as file paths and parsing stops
+> immediately.
+
+`CliArgs` holds the parsed state of the process args. It uses [`args_os`] over [`args`] to
+safely handle non-UTF-8 paths, which can occur on Linux. [`skip(1)`] drops the bin name since it is
+always the first arg. Once `-p` flag is is encountered, all remaining args are collected into paths.
+
+`Stats<'p>` represents the computed result for a single file. It borrows filepath directly from
+`CliArgs` rather than cloning it, with the lifetime 'p tying each Stats instance to its source
+`CliArgs`. Both `lines` and `words` are `Option<usize>`, where `None` means the flag was absent and
+the stat was never computed, while `Some(n)` means the flag was set and the count is `n`. This
+distinction matters because it avoids confusing "not requested" with a legitimate count of zero.
+
+`stat_files` is the core logic. Iterates over `cli_args.paths` and builds a `Vec<Stats<'_>>`.
+Non-files and IO errors are skipped silently via `continue` for simplicity.  Each file is read
+to a string buffer with fs::[`read_to_string()`] for processing. Line count uses
+content.[`lines()`].[`count()`]. Word count splits each line on whitespace via
+[`split_whitespace()`], counts tokens per line, then sums across all lines, which correctly handles
+tabs and multiple consecutive spaces. Both stats are computed only if their respective flag was set,
+so there is no wasted work.
+
+The result is output as one line per file. Stats omitted by the user are excluded from the output
+entirely rather than printed as zero or a placeholder.
+
+```rust,edition2024,no_run
+use std::env;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+// Represents the arguments passed to this process during initialization
+#[derive(Debug)]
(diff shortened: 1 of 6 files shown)


src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
+## Basic Argument Parsing
+This section covers a basic implementation of commandline argument parsing.
+Use [`args`] to get the arguments this process was started with:
+
+```rust,edition=2024
```
AndyGauge:
```suggestion
```rust,edition2024
```
The equal sign is not supported in the skeptic dialect

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
...
+
+        let words = match cli_args.words {
+            true => Some(
+                content
+                    .lines()
+                    .map(|line| line.unwrap_or_default().split_whitespace().count())
+                    .sum(),
+            ),
+            false => None,
+        };
```
AndyGauge:
```suggestion
let mut content = String::new();
  if file.read_to_string(&mut content).is_err() {
      continue;
  }

  let lines = match cli_args.lines {
      true => Some(content.lines().count()),
      false => None,
  };

  let words = match cli_args.words {
      true => Some(
          content
              .lines()
              .map(|line| line.split_whitespace().count())
              .sum(),
      ),
      false => None,
  };
```
I believe this is easier to follow, loading the file into a `String` rather than messing with the `&[u8]` and `BufRead`.  The import would only need `io::Read` and the `lines()` method can be updated to https://doc.rust-lang.org/std/primitive.str.html#method.lines

I'd like to know your thoughts about this refactor.

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
...
+The result is output as one line per file. Stats omitted by the user are excluded from the output
+entirely rather than printed as zero or a placeholder.
+
+```rust,edition=2024,no_run
+use std::{
+    env,
+    ffi::OsString,
+    fs::File,
+    io::{BufRead, Read},
+    os::unix::ffi::OsStrExt,
```
AndyGauge: I wouldn't categorize ffi as basic.  Ideally a basic implementation would be cross-platform and avoid the non-UTF-8 path sematics.

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
...
+            }
+        }
+
+        CliArgs {
+            lines,
+            words,
+            paths,
+        }
+    }
+}
```
AndyGauge, line 100:
```suggestion
fn parse() -> Self {
      let mut lines = false;
      let mut words = false;
      let mut paths = Vec::new();

      let mut args = env::args_os().skip(1);
      while let Some(arg) = args.next() {
          match arg.to_str().unwrap_or_default() {
              "-l" | "--lines" => lines = true,
              "-w" | "--words" => words = true,
              "-p" | "--paths" => {
                  paths.extend(args.by_ref().map(PathBuf::from));
                  break;
              }
              _ => {}
          }
      }

      CliArgs { lines, words, paths }
  }
```
If we drop the bytes it becomes pure string matching which is going to be easier for someone to extend.  Plus we can drop the collect() to avoid overallocating.  Explaining by_ref() to make the borrow checker happy and pass by reference may help.

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
+# Environment Variables
+The [`env`] module lets you inspect and manipulate the process environment, including env vars, CLI
+args, and working directories.
+
+## Get Environment Variables
+[`var`] fetches an env var from the current process. It returns `Result<String, VarError>`, so you
+decide how to handle missing values:
+
+```rust,edition=2024,no_run
```
AndyGauge:
```suggestion
```rust,edition2024,no_run
```

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
...
+
+// Represents the statistics of a given filepath
+struct Stats<'p> {
+    filepath: &'p Path,
+    lines: Option<usize>,
+    words: Option<usize>,
+}
+
+fn stat_files<'p>(cli_args: &'p CliArgs) -> Vec<Stats<'p>> {
```
AndyGauge:
```suggestion
fn stat_files(cli_args: &CliArgs) -> Vec<Stats>
```
The elision shouldn't be required.

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
...
+        env::set_var("MODE", "debug");
+    }
+}
+```
+
+## Remove Environment Variables
+
+Same threading rules apply as [`set_var`]. [`See safety note first`].
+
+```rust,edition=2024,no_run
```
AndyGauge:
```suggestion
```rust,edition2024,no_run
```

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Get Environment Variables
+[`var`] fetches an env var from the current process. It returns `Result<String, VarError>`, so you
+decide how to handle missing values:
+
+```rust,edition=2024,no_run
+use std::env;
+
+fn main() {
+    let mode = env::var("MODE").expect("Env variable MODE not set");
```
AndyGauge: The `expect` here is a panic, which the cookbook intentionally avoids.  It would be helpful to show how to handle this gracefully

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+## Set Environment Variables
+
+[`set_var`] is safe in single-threaded programs, and always safe on Windows. Avoid it in
+multi-threaded programs on other OSes. [`See safety note`].
+
+```rust,edition=2024,no_run
```
AndyGauge:
```suggestion
```rust,edition2024,no_run
```

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
...
+    Error,
+    Debug,
+    Trace,
+}
+
+// Handles conversion of string env variables to LogLevel enum.
+impl TryFrom<String> for LogLevel {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, <LogLevel as TryFrom<String>>::Error> {
```
AndyGauge:
```suggestion
    fn try_from(value: String) -> Result<Self, Self::Error> {
```
I think we should be able to simplify this bit

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,203 @@
...
+    let mut stat_collection: Vec<Stats> = Vec::new();
+    for filepath in &cli_args.paths {
+        // Skip anything that ain't a file
+        if !filepath.is_file() {
+            continue;
+        }
+
+        // Open file for reading
+        let Ok(mut file) = File::options().read(true).open(filepath) else {
```
AndyGauge:
```suggestion
        let Ok(mut file) = File::open(filepath) else {
```
should be enough

src/cli/env.md
```diff
@@ -0,0 +1,141 @@
...
+            server_url: DEFAULT_SERVER_URL.into(),
+            log_level: DEFAULT_LOG_LEVEL,
+            server_port: DEFAULT_SERVER_PORT,
+        }
+    }
+}
+
+impl Config {
+    fn load() -> Self {
+        let server_url = env::var("BACKEND_SERVER_URL").unwrap_or(DEFAULT_SERVER_URL.into());
```
AndyGauge: There's a lazy version of `unwrap_or` worth mentioning.  `unwrap_or_else(|_|)` only allocates during the evaluation of the else branch.

AndyGauge (commented):
Let me know your thoughts about these ideas.  I appreciate your contributions!

src/cli/arguments/basic.md
```diff
@@ -0,0 +1,195 @@
...
+
+cargo run --release -- -l -p src/main.rs
+# Path: "src/main.rs" Lines: 84
+```
+
+[`args_os`]: https://doc.rust-lang.org/std/env/fn.args_os.html
+[`args`]: https://doc.rust-lang.org/std/env/fn.args.html
+[`count()`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count
+[`env`]: https://doc.rust-lang.org/std/env/index.html
+[`lines()`]: https://doc.rust-lang.org/std/io/trait.BufRead.html#method.lines
```
AndyGauge:
```suggestion
[`lines()`]: https://doc.rust-lang.org/std/primitive.str.html#method.lines
```

AndyGauge (approved):
Just one update to make which shifted based on the read-to-string approach!

AndyGauge (approved):

